PR #36, opened 2019-02-14
Prevent viewing deleted or banned posts

### LINKED ISSUE #13: Prevent viewing posts marked as deleted
Users should not be able to view md5 of deleted posts. They should also not be able to generate or view file urls on posts marked as deleted.

There appears to be a whole bunch of loopholes in this, and even blocked posts in the Danbooru codebase.

kiranoot:
This fixes an exploit where you can view deleted and banned posts
through the atom service regardless of user level.

Closes #13

### DIFF
--- a/config/danbooru_default_config.rb
+++ b/config/danbooru_default_config.rb
@@ -462,7 +462,8 @@ module Danbooru
     end
 
     def can_user_see_post?(user, post)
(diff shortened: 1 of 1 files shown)


kiranoot:
TODO: sitemap is suspect as it manually dumps file urls as well
Investigate removal of the sitemap as it doesn't seem to provide useful posts to a search engine.

PR #55, opened 2019-03-14
Tickets

kiranoot:


app/views/static/error.html.erb
```diff
@@ -1,8 +1,9 @@
-<% if CurrentUser.user.try(:is_builder?) && @exception.present? %>
```
kiranoot: Undo this change before merging. Was left in by mistake.

kiranoot:
Future notes, look into making these into rails polymorphic relations to clean up the models and many other attributes of them.

PR #62, opened 2019-04-09
Vote cleanup

kiranoot:


kiranoot:
@liamwhite for review if possible, updated based on comments

PR #94, opened 2019-06-23
Fix new tags starting with the wrong tag count

kiranoot:


kiranoot:
This turned out to be caused by a double commit of post objects since I was updating the locked_tags object incorrectly to assign it to nil. This fixes the incorrect assignment and prevents two commits of the post object in various states of incompleteness.

The tag count update code would run too early and then the tags and tags_was variables would be wrong for the rest of the update. Resulting in tags being created after their counts were applied.

PR #180, opened 2020-01-04
Added own_vote to post serializer

SafoSnep:
Adds an `own_vote` key to the API response to show whether the user has voted on a post or not, and which way. Possible values are `1` for an upvoted post, `-1` for downvoted, or `0` for post not yet voted for.

Ends up being always `0` for logged out users since the CurrentUser.user is checked in [models/post.rb](https://github.com/zwagoth/e621ng/blob/7a1f10b977434981b92fbb55d7f60bf98231a768/app/models/post.rb#L1220).

### DIFF
--- a/app/serializers/post_serializer.rb
+++ b/app/serializers/post_serializer.rb
@@ -92,7 +92,11 @@ class PostSerializer < ActiveModel::Serializer
     object.is_favorited?
   end
 
+  def own_vote
+    object.own_vote&.score || 0
+  end
+
   attributes :id, :created_at, :updated_at, :file, :preview, :sample, :score, :tags, :locked_tags, :change_seq, :flags,
              :rating, :fav_count, :sources, :pools, :relationships, :approver_id, :uploader_id, :description,
-             :comment_count, :is_favorited
+             :comment_count, :is_favorited, :own_vote
 end


kiranoot:
Let me ensure that this doesn't cause an explosion in query volume for lists of posts. The vote value isn't one of the items that is pulled in automatically because rails doesn't express good ways to eager load contextual values like this. At least as far as I know there isn't a good way to preload a singular item like this automatically because it isn't a relationship, and can't be expressed as one.

The other values listed are all set up so that there is minimal additional fetching required to display them, and I'd like to keep it that way. The favorited value is all a big hack, but it does prevent a lookup.

I also need to check and see if the current system of having children_ids is feasible or not. I didn't test it as much as I probably should have, I just made it match the existing danbooru field list and moved them around.

SafoSnep:
Yep I just took a closer look at is_favorited, that's very interesting.

If this PR ends up being unfeasible due to the extra DB query, no worries and feel free to close this. The count of queries would go up by one per post if the user is logged in, so possibly 320 per request if the old limits apply.

kiranoot:
I will look at making this more feasible when I start moving things over to a dedicated API endpoint instead of having them all mixed in with the normal endpoints. That will let me cater the lookups and responses for API use much more effectively. Right now all of the existing endpoints as designed around how the site itself uses them and there is no specialization. This leads to poor runtimes and less than ideal output for API use where more information is available and needed in bulk over how the site typically uses and fetches the information.

Sindrake:
This was eventually resolved in #2059.

PR #230, opened 2020-08-09
Post replacement system

kiranoot:


app/models/user.rb
```diff
@@ -590,10 +592,12 @@ def upload_limit
 
     def upload_limit_pieces
       deleted_count = Post.deleted.for_user(id).count
+      rejected_replacement_count = PostReplacement.rejected.for_user(id).count
```
Earlopain, line 595: Did you perhaps forget to add this to deleted? It's just standing there

kiranoot: It's up in the air on if this will count towards deleted or not. Same with approved replacements against posts uploaded by a user. There are still discussions around potentially modifying the way the upload limit works a little bit in order to limit or discourage bad behaviors.

PR #233, opened 2020-10-04
Make file formatting comply with POSIX standards and remove trailing space characters

aaronfranke:
This PR is split into 3 commits:

* Convert line endings from CRLF to LF (remove carriage return characters). LF takes up less disk space and is more compatible with Linux/Unix tools. Legacy Windows applications usually prefer CRLF, but most modern Windows applications support LF. If someone needs CRLF on their system, Git has a setting called `core.autocrlf` to automatically convert line endings, so you can store LF in the repo and have CRLF when the files are checked out.

* Ensure files end in newlines for POSIX compliance. By POSIX definition, a file that does not end in a newline character is not a valid text file. Many editors add newlines automatically, and some tools (especially Linux/Unix ones) don't work properly unless the files contain a terminating newline character. Most files in this repo already comply, but many don't.

* Remove trailing space characters. This one is pretty straight forward, trailing space characters generally do nothing except take up disk space, so they should be removed.

If desired, I can set up a script to fix these things and set it up on CI to automatically check for these things.

If there are conflicting changes, I can rebase this, it's easy to replicate these changes.


kiranoot:
As this overlaps with some in progress projects that I have locally, I will do this later on myself. Thank you for bringing it to my attention.

aaronfranke:
Any update? I'm not trying to be impatient, I just want to help out and follow up on this. If you want me to do it, just let me know and open the PR and I will force-push a remake of this PR.

Also, I forgot to link you the script I used to make this PR, so [here you go](https://pastebin.com/3u7EtFJu). Comment/uncomment as desired.

PR #244, opened 2021-03-20
Modaction fixes

Earlopain:
This MR does a few things:

- Fixes a bunch of modactions like post_unapprove which weren't present in the decorator/dropdown menu
- Fix for logging empty user flag changes and display of negative user ban durations
- Renames a few modactions in the decorator which had their name changed for logging purposes but not in the decorator itself
- Adds logging to forum post hide/unhide and forum topic sticky/unsticky which were already in the decorator but not logged
- Seperate presentation for wiki page creation and wiki page renaming (old_title is null on creation)

This is the first time I'm doing anything ruby/rails, hopefully the code is good enough or something.
If the extra logging of the post/topic isn't wanted I can simply remove it. I just thought that it wouldn't hurt to add.
I tested the changes on my local copy of e6 and eveything seemed to work fine.

### DIFF
--- a/app/decorators/mod_action_decorator.rb
+++ b/app/decorators/mod_action_decorator.rb
@@ -46,11 +46,7 @@ class ModActionDecorator < ApplicationDecorator
     when "artist_unban"
       "Marked artist ##{vals['artist_id']} as no longer DNP"
     when "artist_page_rename"
-      if vals['old_name'].nil?
-        "Created artist page \"#{vals['new_name']}\":/artists/show_or_new?name=#{vals['new_name']}"
-      else
-        "Renamed artist page (\"#{vals['old_name']}\":/artists/show_or_new?name=#{vals['old_name']} -> \"#{vals['new_name']}\":/artists/show_or_new?name=#{vals['new_name']})"
-      end
+      "Renamed artist page (\"#{vals['old_name']}\":/artists/show_or_new?name=#{vals['old_name']} -> \"#{vals['new_name']}\":/artists/show_or_new?name=#{vals['new_name']})"
     when "artist_page_lock"
       "Locked artist page artist ##{vals['artist_page']}"
     when "artist_page_unlock"
@@ -307,11 +303,7 @@ class ModActionDecorator < ApplicationDecorator
     when "wiki_page_unlock"
       "Unlocked wiki page [[#{vals['wiki_page']}]]"
     when "wiki_page_rename"
-      if vals['old_title'].nil?
-        "Created wiki page [[#{vals['new_title']}]]"
-      else
-        "Renamed wiki page ([[#{vals['old_title']}]] → [[#{vals['new_title']}]])"
-      end
+      "Renamed wiki page ([[#{vals['old_title']}]] → [[#{vals['new_title']}]])"
 
     when "bulk_revert"
       "Processed bulk revert for #{vals['constraints']} by #{user}"
--- a/app/decorators/mod_action_decorator.rb
+++ b/app/decorators/mod_action_decorator.rb
@@ -295,7 +295,11 @@ class ModActionDecorator < ApplicationDecorator
     when "wiki_page_unlock"
       "Unlocked wiki page [[#{vals['wiki_page']}]]"
     when "wiki_page_rename"
-      "Renamed wiki page ([[#{vals['old_title']}]] → [[#{vals['new_title']}]])"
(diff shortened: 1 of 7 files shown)


app/logical/user_promotion.rb
```diff
@@ -63,7 +63,7 @@ def flag_check(added, removed, flag, friendly_name)
     flag_check(added, removed, "no_flagging", "flag ban")
     flag_check(added, removed, "no_feedback", "feedback_ban")
 
-    if added || removed
+    if !added.empty? || !removed.empty?
```
kiranoot: In ruby it is ideal to use `unless` for inverted conditions. Making this statement `unless added.empty? && removed.empty?`

app/models/forum_post.rb
```diff
@@ -26,6 +26,12 @@ class ForumPost < ApplicationRecord
   after_update(:if => ->(rec) {rec.updater_id != rec.creator_id}) do |rec|
     ModAction.log(:forum_post_update, {forum_post_id: rec.id, forum_topic_id: rec.topic_id, user_id: rec.creator_id})
   end
+  after_update(:if => ->(rec) {rec.is_hidden? && rec.saved_change_to_is_hidden?}) do |rec|
```
kiranoot: It would be cleaner to combine these two callbacks into a single callback and change the value passed as the first value to ModAction.log based on the current value of `is_hidden?`

app/models/forum_topic.rb
```diff
@@ -27,6 +27,12 @@ class ForumTopic < ApplicationRecord
   after_save(:if => ->(rec) {!rec.is_locked? && rec.saved_change_to_is_locked?}) do |rec|
     ModAction.log(:forum_topic_unlock, {forum_topic_id: rec.id, forum_topic_title: rec.title, user_id: rec.creator_id})
   end
+  after_save(:if => ->(rec) {rec.is_sticky? && rec.saved_change_to_is_sticky?}) do |rec|
```
kiranoot: Combine these as well please.

app/decorators/mod_action_decorator.rb
```diff
@@ -295,7 +295,11 @@ def format_description
     when "wiki_page_unlock"
       "Unlocked wiki page [[#{vals['wiki_page']}]]"
     when "wiki_page_rename"
-      "Renamed wiki page ([[#{vals['old_title']}]] → [[#{vals['new_title']}]])"
+      if vals['old_title'].nil?
```
kiranoot: Wiki page creation logging appears to be a bug instead of a feature. Wikis have a history so their creator is known, so it isn't as important to log them to the action log. As only moderators, and in the case of artists, janitors, can destroy these entities, there is sufficient logging to investigate the source of them. Updating the logging action to fire only on update instead of on creation would fix this.

You can use `new_record?` to test if a record has been saved or not in AR callbacks.

app/decorators/mod_action_decorator.rb
```diff
@@ -45,6 +45,16 @@ def format_description
       "Marked artist ##{vals['artist_id']} as DNP"
     when "artist_unban"
       "Marked artist ##{vals['artist_id']} as no longer DNP"
+    when "artist_page_rename"
+      if vals['old_name'].nil?
```
kiranoot: This should have the same fate as wiki page creation.

kiranoot (changes requested):
Looks good. Thank you for taking the time to fix these up.

I have two small changes I'd hope you could make. I reviewed how we're logging creations and it seems that it doesn't provide a useful purpose, so removing those log entries would be ideal to help reduce the clutter in the mod action log.

app/logical/user_promotion.rb
Earlopain: Fixed, didn't know about that

app/models/forum_post.rb
Earlopain: Yeah, you're right. Did it to this one and the other one as well.

app/decorators/mod_action_decorator.rb
Earlopain: Makes sense, the action is named rename after all. I removed the decorator code with special handling when it was created, the old logactions will still say created but I think thats ok

kiranoot:
I have some other tasks to play with today, but I will test this and merge it as soon as I can. Thank you again.

kiranoot (approved):

PR #245, opened 2021-03-20
[Seeding] Fix system user not being promoted

Earlopain:
It's me again, this time with a fix for the initial setup of the databse.

Creating users with the level property set has no effect, it will be overwritten in the function `promote_to_admin_if_first_user` (at least thats how I understand it). This prevents some system user actions which require higher privileges.

I also changed the system user name and level to the values of `auto_moderator` on the live site because I think thats the system user there. Correct me if I'm wrong.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -58,6 +58,8 @@ unless Rails.env.test?
 
     md5 = Digest::MD5.hexdigest(data)
     service = UploadService.new({
+                                    uploader_id: CurrentUser.id,
+                                    uploader_ip_addr: CurrentUser.ip_addr,
                                     file: file,
                                     tag_string: post["tags"].values.flatten.join(" "),
                                     source: post["sources"].join("\n"),
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -31,7 +31,7 @@ User.find_or_create_by!(name: Danbooru.config.system_user) do |user|
   user.password_hash = ""
   user.email = "system@e621.net"
   user.can_upload_free = true
-  user.level = User::Levels::ADMIN
+  user.level = User::Levels::JANITOR
 end
 
 unless Rails.env.test?
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -22,6 +22,7 @@ admin = User.find_or_create_by!(name: "admin") do |user|
   user.password_hash = ""
   user.email = "admin@e621.net"
   user.can_upload_free = true
+  user.can_approve_posts = true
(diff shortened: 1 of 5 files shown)


kiranoot:
Thinking about this. It seems like this is a kind of anti-pattern. We are already seeding the database, more or less as a requirement of the software operating. So the `promote_to_admin_if_first_user` behavior is not needed, or even desirable. It also created unexpected behavior within the system as it forces setting levels in circuitous ways, and breaks testing.

Upon review the default user level in the database isn't properly configured to be `20` the way that would be most useful, but instead defaults to `0` which is an anonymous user.

Creating a migration and using `change_column_default` to update this to be 20 is an option. Doing some kind of check in code and setting it to 20 if it is 0 upon creation is another, however that may interfere with tests(?).

I'll have to sit and think about what the best avenue for this one is. It's an invasive change that I'm suggesting, while yours doesn't touch anything, but is a work around to a deeper problem.

My initial hesitation is having to deal with a merge related to structure.sql and migrations, and the replacement system branch.

Earlopain:
I'll leave it up to you then. This is not something thats very important anyways, you just have to set it up once and can then forget about it. I just noticed it while setting things up.

Settings a default on the db level sound reasonable to me. This isn't urgent, I'll just close it for now

Earlopain:
On second though I'll just leave this here. I removed the promote function and added an migration for the default in the db.
You can just take up when the post replacement stuff is done

Earlopain:
Redid the migration, which fixed some merge conflict

Earlopain:
Rebased once again.

Commit a28466bdaf2b6c3826c9d0914387c9305983397d broke seeding altogether because uploader attributes where not set by default anymore. If I should do that in a seperate PR please do tell

db/seeds.rb
```diff
@@ -31,7 +31,7 @@
   user.password_hash = ""
   user.email = "system@e621.net"
   user.can_upload_free = true
-  user.level = User::Levels::ADMIN
+  user.level = User::Levels::JANITOR
```
kiranoot, line 36: System user, and probably the admin user, should be able to approve posts, otherwise it is likely to run into problems during execution. For the admin user, it would just be nice to have, since otherwise you have to set it manually.

kiranoot:
Good catch. Thank you. Also this PR is fine. I should be able to get to testing and merging this tomorrow.

db/seeds.rb
Earlopain: I thought it was implied by the user level but I guess not. Added

PR #247, opened 2021-03-21
Add a bunch of missing html titles

Earlopain:
Some pages I frequently use like uploading and sets were missing their titles. This adds them and a bunch of other missing ones as well. Some of them are deprecated or unused like Artist Commentaries but I added them anyways for completeness sake.

The first commit removes `app_name` from every `page_title` and instead adds it back later in a helper function. The only page which didn't have `app_name` at the end was the root page so I special cased that in the helper.
Overall this commit is not really nessesary but I thought it makes the code a bit cleaner.

The second commit adds the missing titles and adds useful information to a few existing ones, like the tag name when you are editing a tag. Most existing titles already had this kind of information.

The third commit doesn't really fit with the other two, I just added it here to not spam MR (tell me I should do that instead). It replaces the plain 404 error page with the already used one in `application_controller`. That way you don't get blinded by accidentally misstyping a url.
Not sure if it's important but I made it to only respond as html which keeps it in line that no valid json will be served, even if the request ends with `json`. You only get valid json if you are on an existing route.

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -10,7 +10,7 @@ class StaticController < ApplicationController
   end
 
   def not_found
-    render plain: "not found", status: :not_found
+    render "static/404", formats: [:html], status: 404
   end
 
   def error
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,7 +4,6 @@ class ApplicationController < ActionController::Base
   skip_forgery_protection if: -> { SessionLoader.new(request).has_api_authentication? }
   before_action :reset_current_user
   before_action :set_current_user
-  before_action :set_title
   before_action :normalize_search
   before_action :api_check
   before_action :set_variant
@@ -12,6 +11,7 @@ class ApplicationController < ActionController::Base
   after_action :reset_current_user
   layout "default"
 
+  include TitleHelper
   include DeferredPosts
   helper_method :deferred_post_ids, :deferred_posts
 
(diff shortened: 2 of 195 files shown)


app/controllers/static_controller.rb
```diff
@@ -10,7 +10,7 @@ def accept_terms_of_service
   end
 
   def not_found
-    render plain: "not found", status: :not_found
```
kiranoot, line 13: I remember there being something surrounding error/404 templates and exceptions related to template rendering, but I think they might have been resolved at some point along with the changes to error handling. I'll make sure that this works as intended on my side.

kiranoot:
:100: Looks good to me. I'll do a checkout on this and play around with it, verify everything and then merge it when I can.

app/controllers/static_controller.rb
Earlopain: Do you mean what happenes when the template might have a syntax error or something?
`application_controller` has a generic `rescue_from` while it's missing from the static one. This shows the error stack and other stuff to the user. I don't know how it might look in production but I think it's probably the same?

Should I had a error handler to the static controller as well?

Edit: Whoops, I was on the wrong branch, the exception gets thrown inside the exception handler which apparently means it won't be catched again, but If you don't have one everything is fine?

Compare http://e621.local/test with http://e621.local/posts/-1 when the 404 template is malformed

kiranoot: The old behavior is gone now, so this should be safe. No changes needed. If anything crops up I'll address it as needed.

PR #249, opened 2021-03-22
[UI] Fix multiple static navbar entries being active

Earlopain:
Previously the controller was used exclusivley to check if a route
was active. Because static entries all use the same controller it was
possible for multiple navbar entries to be active at the same time
if you were on a static page.

This fixes that by first checking for route equality and ignoring the
static controller later on.

Before: 
image
After:
image
image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -266,6 +266,9 @@ module ApplicationHelper
 
 protected
   def nav_link_match(controller, url)
+    # Static routes must match completely
+    return url == request.path if controller == "static"
+
     url =~ case controller
     when "sessions", "users", "maintenance/user/login_reminders", "maintenance/user/password_resets", "admin/users", "dmails"
       /^\/(session|users)/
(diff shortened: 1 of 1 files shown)


kiranoot:
Unfortunately this breaks an intended behavior of the site as well as fixes two of the menu items. The "More >>" menu item is supposed to be highlighted when the controller is not in the list of known tabs. With this patch it no longer highlights.

The whole system and how it interacts with the menu highlight needs some kind of rework to resolve the ambiguity between "more >>" and "discord" as they are part of the same controller.

Earlopain:
> Unfortunately this breaks an intended behavior of the site as well as fixes two of the menu items. The "More >>" menu item is supposed to be highlighted when the controller is not in the list of known tabs. With this patch it no longer highlights.

I didn't realise that this was a thing. I'll take a look if I can't come up with a better way of handling the navlinks

Earlopain:
I changed the code so that the fix doesn't break old behaviour. No rework or anything but at least this doesn't fix the bug just to introduce a new one in the same step...

kiranoot (approved):

PR #252, opened 2021-04-03
Autocomplete fixes

### LINKED ISSUE #251: Autocomplete middle click action always opens a tag search
There are a few pages which feature autocomplete, most importantly when you are searching for a post.

But Wikis, Pools, Artists and probably a few more also feature autocomplete. Currently these don't to what I expect them to do on middle clicking:

- Wikis: Opens a tag search for the page I wanted to look at
- Pools: Opens a tag search for the pool name, which will probably result in no posts being found
- Artists: Opens a tag search for the artist

It seems like they all just treat the returned text as a tag. If I am already on the dedicated page for those I expect to also be redirected to them on search. Heres an example for wiki autocomplete:

image

Is it ok if I open a PR which changes this behaviour?

Earlopain:
Redirect the user to the specific entry when middle clicking on autocomplete entries instead of always starting a tag search.

Adds the id to the autocomplete entry visible to javascript which will then be used to construct the href. Also adds the id to tags, which don't necessarily need them but that way it's uniform. The id was already returned in the api call but was always null.

It seems like there was other logic for autocomplete, `tag_autocomplete.rb` but I believe that was unused, so I removed it.

Fixes #251


kiranoot:
:100:

PR #254, opened 2021-04-05
Improve initial vagrant setup

Earlopain:
This does a few things:

- Set the default user for ssh back to `danbooru`. This was uncommented a while ago but I don't know why, there probably was a reason. Because ruby-installs are local to a user the vagrant user does not have the correct ruby version to chroot to. This might be fixable by doing a system wide install but this seemed like the easier approach
- Don't copy over the local config if it already exists
- Seed the posts with data from the live site. Instead of putting everything in the yaml just use the api to fetch the info. This also gives access to a few things which were previously not seeded like sources, description and tag types. I also replaced a few deleted posts while I was at it
- Fix `can_upload_with_reason` not accounting for `disable_throttles`. Noticable when seeding because only 30 posts where uploaded. This worked previously but at some point the hourly upload limit was enforced on all users, including admins. This still means that during seeding `disable_throttles` must be set to true. Not sure what the best solution is here

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,7 +25,7 @@ Vagrant.configure('2') do |config|
   # end
 
   VAGRANT_COMMAND = ARGV[0]
-  # config.ssh.username = 'danbooru' if VAGRANT_COMMAND == 'ssh'
+  config.ssh.username = 'danbooru' if VAGRANT_COMMAND == 'ssh'
 
   config.vm.define 'default' do |node|
     node.vm.hostname = 'e621.local'
@@ -34,4 +34,3 @@ Vagrant.configure('2') do |config|
 
   config.vm.provision 'shell', path: 'vagrant/install.sh'
 end
-
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -557,15 +557,15 @@ class User < ApplicationRecord
     end
 
     def can_upload_with_reason
-      if hourly_upload_limit <= 0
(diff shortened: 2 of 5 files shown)


kiranoot:
Sorry it took me so long to get back to this. The last week has been pretty hectic on my end.

The official danbooru source uses rbenv, but I was told by @liamwhite not to use it, and to use chruby instead for reasons surrounding stubs and portability.

I would like to keep the development machine as close as possible to how the production machines are set up so that there are not significant setup differences. Since the production machines use chruby and ruby-install, perhaps we can troubleshoot what is going wrong and why it isn't being used for the initial part of the install. My first glance suggests that it only gets activated if it was installed before the script is run the first time, so it may be as easy as adjusting the script to run chruby after it is installed.

Earlopain:
I'll take a look getting it to work with chruby/ruby-install. I oriented myself at the other existing install scripts (`app_server.sh`, `INSTALL.debian`, `INSTALL.ubuntu`) which install rbenv but I guess they aren't used anymore.

Earlopain:
Turns out I was trying to fix a non-issue. All that needed to be done is set the login user to `danbooru`. ruby-installs are user-local which means the vagrant user had no chance to pick them up and changing users with i.e. `sudo su danbooru` didn't pick up on chruby, probably because that doesn't start a loginshell.

Once chruby is loaded and the correct version is installed cding to the folder sets the correct ruby version automatically because of `chruby/auto.sh`

Was there a reason why `danbooru` was changed for the default user?

kiranoot (approved):

PR #255, opened 2021-04-05
UI tweaks

Earlopain:
This tweaks a few things:

- Remove useless entries when creating forum categories
- Hide non visible forum categories when creating a forum topic
- Hide buttons on upload whitelist page when they won't do anything for you
- Put the footer always at the bottom of the viewport

While doing this I also stumbled onto a few missed exceptions for post/topic creation. Those were already checked in https://github.com/zwagoth/e621ng/blob/c5eb6ce2c343ef57a4805734bc6d16746e9cd87a/app/models/user.rb#L518-L519
which means that now it will work with only `disable_throttles` set to true instead of also having to also set `disable_age_checks`


kiranoot:
Most of these changes look fine. Tests for the change in throttle behavior would be nice, just so that something exists to ensure they don't get broken in the change. I'd request that you drop the change involving always showing the footer near the bottom of the page. I don't want to change that behavior, especially because it touches the style for the base element.

Earlopain:
I fixed the tests for forum post/topic and users, added a test for the throttle changes and removed the always on bottom behaviour.

Seeding now adds the alias category from the config. Some tests assumed this will always exist (and failed) and some created it themselves. Other code like creating tag aliases also require it. This should work out of the box now.

The throttle functions have their limit memoized which interfered with the tests. I added that they don't get memoized when running in test mode but I don't really like that. Is there maybe someway to disable memoization when running tests? I look it up briefly but didn't find anything.

kiranoot:
It looks like memoist supports a `flush_cache` method on objects that contain memoized values.

I'm looking through and trying to figure out why throttle values are memoized. I know the upload limit one is done because the upload pages are a mess and check it frequently and multiple times per page. I don't see any obvious cases where memoization would be of great benefit to this area. I feel like I added it defensively instead of analyzing the true usage per call.

Earlopain:
Should I just remove the memoization then? I checked too but the only place the functions are used is during validation. One is in `CommentVote` which is already covered by the validation in the model, I removed it and added it to the tests, still seems to work fine.

PR #256, opened 2021-04-06
Remove useless config entries

Earlopain:
They were probably useful at some point, but now they don't do anything


kiranoot:
💯 Thank you.

PR #264, opened 2021-05-03
Post set UI

Earlopain:
Let's start with a few images
Post set maintainer old:
image
Post set maintainer new:
image
Maintainers for set old:
image
Maintainers for set new:
image

All other things I know of use the table view, this one had it's own thing. In addition to being badly aligned it also had a bunch of links which are supposed to order but had no effect.

I merged the different status views together which I think is fine, especialy when ordered by last modified. I didn't add search functionality to the table because I honestly don't know if that would even be needed.

On the way I fixed a bunch of smaller things
- Unnecessary hrefs on the set index page. Those were supposed to order but they don't work because the var is missing [search]. I removed that altogether because no other table has clickable header links
- Add autocomplete to the maintainer input and handle non existant users gracefully
- Fix wrong order parameter value on search form for post sets


app/views/post_set_maintainers/index.html.erb
```diff
@@ -1,68 +1,21 @@
...
-      <% @invites.where(status: 'blocked').each do |m| %>
-        <% if m.post_set.is_public %>
-          <tr>
-            <td><%= link_to m.post_set.name, post_set_path(m.post_set) %></td>
-            <td><%= m.status.capitalize %></td>
-            <td>
-              <%= link_to "Unblock", deny_post_set_maintainer_path(m), 'data-confirm': 'Are you sure you want to unblock invites for this set?' %>
-            </td>
+            <td><%= compact_time m.updated_at %></td>
+            <td><%= invite_links m %></td>
```
kiranoot: Generally I dislike helpers that generate large amounts of html in rails because they are an easy way to make XSS injection mistakes through use of html_safe. This could be done using a partial and passing the invite member as a local to the view template. Something like `<%= render "post_set_maintainers/invite_links", invite: m %>`

If you're open to converting this, that's cool. If not, I'm open to accepting it as is too.

Earlopain: No issue with changing that, thanks for the input (:

PR #265, opened 2021-05-03
[NewsUpdate] Show date and format dtext

Earlopain:
- Add date column
- Hide last th if not moderator, like it does for tr
- Format dtext news message

### DIFF
--- a/app/views/news_updates/index.html.erb
+++ b/app/views/news_updates/index.html.erb
@@ -6,15 +6,19 @@
       <thead>
         <tr>
           <th>Creator</th>
+          <th>Date</th>
           <th>Message</th>
-          <th></th>
+          <% if CurrentUser.is_moderator? %>
+            <th></th>
+          <% end %>
         </tr>
       </thead>
       <tbody>
         <% @news_updates.each do |news_update| %>
           <tr id="news-update-<%= news_update.id %>">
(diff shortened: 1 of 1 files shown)


app/views/news_updates/index.html.erb
```diff
@@ -6,15 +6,19 @@
...
+          <% end %>
         </tr>
       </thead>
       <tbody>
         <% @news_updates.each do |news_update| %>
           <tr id="news-update-<%= news_update.id %>">
             <td><%= link_to_user news_update.creator %></td>
-            <td><%= news_update.message %></td>
+            <th><%= compact_time news_update.updated_at %></th>
+            <td><%= format_text news_update.message %></td>
```
kiranoot: There needs to be a div surrounding formatted dtext with the class `styled-dtext` for most dtext elements to render properly. Otherwise this looks great.

kiranoot (changes requested):

app/views/news_updates/index.html.erb
Earlopain: Thanks for the input, I changed it. I also noticed I put a `th` instead of `td` in the body (whoops, my bad), changed that too.

kiranoot (approved):

PR #266, opened 2021-05-03
Update some help links

Earlopain:
Some pages link to wiki pages with the help prefix but their only content was a link to a wiki page with e621 as the prefix.
Also removes something related to autotagging which isn't used anymore, I just removed it because of the wiki link.

`help:home` and `help:flag_notice` actually exist, `help:appeal_notice` and `help:replacement_notice` don't but theres also no e621 entry for them


kiranoot:
Thanks again for all the help.

PR #275, opened 2021-06-03
[PostVersions] Fix display for parent/rating

Earlopain:
The diff algorithm has trouble with parent/rating values. Rating and parent changes are always strikethrough, if no parent is present on upload it will simply show `parent:`. This aims to fix that. A sideeffect of this change is that parent/rating will also be displayed when it didn't change from the previous version but I believe that should be fine since they will be appended at the end of the list anyways.

Old              |  New
:-------------------------:|:-------------------------:
image  |  image

This has no effect on undoing versions. Undoing happens in `changes` while `diff` is only responsible for visualizing it.

### DIFF
--- a/app/models/post_archive.rb
+++ b/app/models/post_archive.rb
@@ -215,12 +215,12 @@ class PostArchive < ApplicationRecord
     if post.nil?
       latest_tags = tag_array
     else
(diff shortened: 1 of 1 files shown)


kiranoot:
Seems reasonable enough.

PR #277, opened 2021-06-14
Improve user listing page

Earlopain:
First commit:
https://e621.net/users?search[order]=post_upload_count where users are supposed to be ordered by upload count. What happens instead is that I can see a trend in the uploads but it's not what I would call sorted. The problem here is that the number that gets displayed is doing a count over the `posts` table while sorting happens through `user_statuses`. The same is true for deleted posts and note changes. Fixing this is fairly simple, just replace the count with the value from `user_statuses`. This will also probably improve performance since all those counts won't need to happen anymore. Technically the current value is more correct but every other place also uses `user_statuses` like the profile page

Second commit:
Adds a few more sortable stats to the listing: Wiki Changes/Artist Changes/Pool Changes/Forum Posts/Comments. Since they all go over `user_statuses` there won't be a performance impact.
I find it interesting to look at where I am in relation to other users and who are the powerusers. Since some stats were already visible/sortable I don't see why the rest shouldn't be.


kiranoot:
I'll have to check some of this on a larger dataset. Right now the join and sort based on user_statuses doesn't perform as well as I'd like it to. So expanding which fields are used to sort by is likely to result in those timing out before returning results. But using user_statuses for the display values should be a significant improvement on loading times.

Earlopain:
I though it would be fine since sorting by post count was already possible but I guess they would all be equally slow. Though you can't sort by more than one column at a time and since `post count` works right now the rest should too? `post_count` doesn't have an index (the rest doesn't either), it doesn't have some advantage over the other fields.
Or do you mean that the query is almost timing out and probably won't work in the near future?

kiranoot:
That the current query is currently taking a long time to run, and I'm not sure that it will continue to work properly in the future. Indexes could be added, but postgresql can be annoying about when it does, and does not use indexes. It should in this case, but I'll have to do testing.

Earlopain:
I added indices to the sortable columns, I don't know if they will get used though. On my dev db they don't but I also only have 2 rows.

PR #278, opened 2021-06-15
Order artist urls on artist page

Earlopain:
Currently artist urls are displayed as users add them. I myself try to group them somewhat together, but only if I have a link to add.
If an artist has especially many urls they'll be pretty much mixed with inactive urls sprinkled in. Instead order them after a fixed pattern:
* Put urls with actual uploadable content to the top, like furaffinity/twitter etc.
* Put livestream/donation after that
* End with a few miscellaneous domains like linktr/trello
* Alwasy put inactive urls below active ones

Obviously I didn't put in every domain but this should catch the large majority. Unknown domains will simply be put at the end of active urls, right before the inactive ones.

| Before  | After |
| ------------- | ------------- |
| image | image 
| image | image 

Looks much cleaner in my opinion and the links which I'm actually interested in are at predictable locations.

### DIFF
--- a/app/models/artist.rb
+++ b/app/models/artist.rb
@@ -192,7 +192,7 @@ class Artist < ApplicationRecord
     end
 
     def sorted_urls
-      urls.sort {|a, b| a.priority <=> b.priority}
+      urls.sort {|a, b| b.priority <=> a.priority}
     end
 
     def url_array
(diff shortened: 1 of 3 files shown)


kiranoot:
Looks good to me!

PR #279, opened 2021-06-16
[Vagrant] Expose postgres server to host

Earlopain:
This is something I added manually everytime I after I destroyed the image. Since this could probably be usefull to others I though why not add it to the setup script.

Should the port already be occupied it will simply not map, informing the user during `vagrant up` and continue on with the setup.


Earlopain:
After some discusison I removed the port forwarding in the vagrantfile. For some reason it didn't occur to me that you are simply able to access it via the ip.

PR #284, opened 2021-06-27
Add replacement tracking for uploader

Earlopain:
Adds the possibility for users to see which of their posts got replaced. For a given replacement the uploader at the time of approval gets saved which can later on be used to filter for specific users. I also added the option of penalizing the original user for the upload which can be toggled later on. When a replacement gets accepted you can choose if it should affect the users upload limit or not. I made the actions use ajax which feels better in my opinion. Getting thrown of the page doesn't feel that great in my opinion.

Another thing I noticed was that replacements don't hide their thumbnails when the post is deleted because they don't behave like post thumbnails. I checked if the post is currently deleted and if it is display the deleted thumbnail. There are still other problems like that the thumbnails don't listen to other things like the blacklist or safeblock but to fix that things would need to change a bit more.

I created a fixer script which tries to fill the newly added fields from the currently available data, but if a post has more than two replacements it's not really possible. I imagine the most only got replaced once though.

### DIFF
--- a/app/controllers/post_replacements_controller.rb
+++ b/app/controllers/post_replacements_controller.rb
@@ -25,7 +25,7 @@ class PostReplacementsController < ApplicationController
 
   def approve
     @post_replacement = PostReplacement.find(params[:id])
-    @post_replacement.approve!
+    @post_replacement.approve!(penalize_current_uploader: params[:penalize_current_uploader])
 
     respond_with(@post_replacement, location: post_path(@post_replacement.post))
   end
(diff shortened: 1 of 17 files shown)


db/migrate/20211017091212_add_replacement_audit_stats.rb
```diff
@@ -0,0 +1,18 @@
+class AddReplacementAuditStats < ActiveRecord::Migration[6.1]
+  def up
+    add_column :post_replacements2, :uploader_id_on_approve, :int
```
kiranoot: Please add these two new post replacement attributes to the hidden attributes list for json. It would also be nice to have :protected marked as hidden as well.

Earlopain: Sure thing, see newest commit.

PR #297, opened 2021-07-22
Misc again

Earlopain:
Another one of those pr where I put a bunch of smaller stuff together.

1. Repeatedly clicking on edit on a post would grow the tagbox continuously in size. Fix this by simply resizing the textarea just once on the initial open. Coincidentally this also fixes the tagbox not being correctly resized on initial open because vue completely replaces the element it gets mounted on, throwing the first resize away.
2. Update embed title from `drawn by`to `created by` to account for 3d artwork. I saw this brought up in the discord at some point.
3. Fix some wrong table column on the ip adresses page.
~~4. Fix html encoding on the ticket index page. For some reason stripping first and truncating after fixes the issue, like it already happens for the title.~~
image
Stuff still gets correctly html encoded.
~~5. Editing an artist without a wiki page would always say that note changes happenend. This probably happenend because `notes` is nil while `text` was the empy string.~~
6. Add to set would always flash up on the posts index, even when not selected. The other two select boxes already behaved correctly. This also adds some padding, like on the other inputs. The code already handles showing the box correctly when it is selected on load.
7. Fix typo on `respond_to` on tag type versions.
8. Make the source buttons on the uploader a bit larger. I always found them a bit to hard to hit.
image
image
8. Remove appeals, add stats, db export and avoid posting and link to pools gallery on the sitemap.
9. Remove highchart references from the stats page. Some time ago you removed the needed vendor scripts but the text and some js was still left.
10. Fix wrong wiki link on the uploader. Seems like I fucked this one up im my uploader pr.
11. Filter out duplicate tags on the tag preview. Consider the following: A -> B -> C but also A -> D -> C. C now appears twice in the list.
image
image
12. Correctly display implications in the tag preview. Character implies Copyright but both tags use the same tag category.  For this to work I obviously need to load the category for both tags and not only for b.
image
image
13. Increase vagrant RAM. The machine was swapping quite a bit, 700mb after bringing the machine up and loading one page.
image
image
14. Simple fix for #298, looks like the returned data changed structure but that was not accounted for in all places. This seems to be the only place in js where this might happen.


app/models/artist.rb
```diff
@@ -342,7 +342,7 @@ def notes
     end
 
     def notes=(text)
-      if notes != text
+      if notes != text && !text.empty?
```
kiranoot: This removes logging of when a note is removed. If empty notes are being logged every time then a special check needs to be made.

app/views/tickets/index.html.erb
```diff
@@ -44,7 +44,7 @@
 				<% if !ticket.can_see_reason?(CurrentUser.user) %>
 					<td><span style="cursor:help;" class="redtext" title="Due to privacy concerns, this information is confidential">Confidential</span></td>
 				<% else %>
-					<td class="ticket-subject" title="<%= h(truncate(strip_dtext(ticket.reason), length: 200)) %>"><%= link_to h(strip_dtext(truncate(ticket.subject, length: 200))), action: "show", id: ticket.id %></td>
+					<td class="ticket-subject" title="<%= h(truncate(strip_dtext(ticket.reason), length: 200)) %>"><%= link_to h(truncate(strip_dtext(ticket.subject), length: 200)), action: "show", id: ticket.id %></td>
```
kiranoot: I'd ask that you drop this commit for right now and I will investigate and address it further. Subjects probably shouldn't be allowed html, or dtext for that matter.

kiranoot (changes requested):

Earlopain:
I dropped them both for now. I'll take a look at the artist one at a later date

PR #302, opened 2021-08-05
Revert "[NewsUpdate] Fix new entries not being displayed immediately"

kiranoot:
Reverts zwagoth/e621ng#299

This was causing an exception related to cached array values in production, and needs to be further debugged. For the time being this is a revert action to get things going again.


Earlopain:
Didn't think about this to be honest. I switched it up because only one value gets accessed at most anyways. Sorry about that.

The cache would need to be flushed but that proved to be difficult because of the underlying problem. Should I just redo it without changing the cache structure?

PR #304, opened 2021-08-14
Display duplicate tickets on create

Earlopain:
Show the user a notice when they are reporting something that is already reported, if they are able to view the tickets. User reports and other restricted tickets will not be shown. Only tickets directly visible to the user are considered.

image

Also makes it possble to view forum post tickets after the forum post has been hidden. Previously you could'nt even view your own tickets after they have been handled (and from other users as well). This matches the behavious of the other ticket types for public resources.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -12,6 +12,8 @@ class TicketsController < ApplicationController
   def new
     @ticket = Ticket.new(qtype: params[:type], disp_id: params[:disp_id])
     check_new_permission(@ticket)
+    @duplicates = Ticket.open_duplicates(params[:type], params[:disp_id])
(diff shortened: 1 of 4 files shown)


kiranoot:
This creates a privacy and permissions bypass for ticket types where it isn't possible for users to see the target of tickets, such as user complaints. I am rejecting this as duplicates are primarily a concern for administration handling tickets and not users filing reports. Right now they are only displayed so that they can be quickly grouped and handled in bulk instead of having to find the related tickets one at a time. Internally there are benefits for allowing duplicates like this.

Earlopain:
Sure. User complaints and other non visible tickets were filtered out thought because of `duplicates.select {|t| t.can_see_details?(CurrentUser.user)}` in the controller.
I just thought it might be useful. It's not a large change so I'm not sad it hasn't been accepted. Otherwise I would have talked with you abou it anyways.

PR #306, opened 2021-08-14
[NewsUpdate] Update delay fix, take two

Earlopain:
Redo of #299 and undo of #302
This caused  problems because the previous cache contained an array and
the new one does not. Changing the cache key should prevent trouble when
deploying this to production

This is identical to the previous attempt, with the exception of the cache key being changed.


kiranoot:
I'm just going to avoid this because it doesn't seem like it ever comes up.

PR #309, opened 2021-08-24
[DText] Prevent content from overflowing

Earlopain:
Prevent dtext from overflowing out of its content box with deeply nested dtext or spooky zaglo text.

Here's some before/after footage. Ignore the pause in the middle, I'm restarting the server.
https://user-images.githubusercontent.com/14981592/130659211-4d679a81-1ee0-4f17-be4c-a527381e5f44.mp4

Stuff seems to work fine with this change, though testing this extensively is not that easy for me since I'm missing some real world data. I'll test this some more myself at a later date.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -82,6 +82,14 @@ module ApplicationHelper
   end
 
   def format_text(text, **options)
+    raw %(<div class="dtext-content">#{dtext_ragel(text, options)}</div>)
+  end
+
+  def strip_dtext(text)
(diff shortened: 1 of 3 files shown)


kiranoot:
Were there any additional changes that needed to be done to this? I notice it is still a draft.

Earlopain:
There are still a bunch of things that don't look right. Especially on smaller screens the divs in tables don't extend as much to the right anymore, resulting in unnecessarily long and hard to read text. For some reason the text in tables also isn't centered anymore but aligned top instead.

PR #317, opened 2021-09-26
Add iqdb/iqdbs to vagrant image

### LINKED ISSUE #315: Add iqdb/iqdbs to docker
There's a bunch of stuff I'd like to do (#300, #313, https://e621.net/forum_topics/30819) but since it's not added to the image I don't actually have the ability to work with it.

Earlopain:
This seems to work but there are a few things I'm not entirely certain about:
* iqdb requires `libgd-dev` which seems to have provided `gdlib-config` at some point. This has since been removed and `make`ing iqdb complains about not finding it. That doesn't seem to matter all that much though because the build finishes successfully and png/jpg/gif are possible to reverse serch.
* I made a patch to the iqdb source code because it wouldn't build otherwise. Just removing a line seems strange, maybe it wasn't used or something.
* I had to make the default  storage manager hierarchical since that is how iqdbs expects the files to be.

I'm also installing iqdb from iqdb.org/code right now which I'd like to replace with a git repository under your account in case this ever goes down. It's licensed under GNU GPLv2 so it should be fine.

I replaced the hardcoded database file location in the iqdbs `Procfile` to `$IQDB_DATABASE_FILE` from the .env file since the path there uses the home directory of the vagrant user which the danbooru user does not have access to. This env variable has to be set anyways for stuff to work so it should already be set anyways. I created https://github.com/zwagoth/iqdbs/pull/1 with this change.
It might be possible that there are two seperate databases in which case this will not work, but I don't know enough about iqdb to say if thats the case.

Closes #315


Earlopain:
Doesn't seem worth to merge this with the bunch of hacks and the breaking config change I needed to make. I'll leave the branch up though since it does work. I don't want to figure all this dumb shit out again when I want to to something with iqdb/s again.

PR #320, opened 2021-09-30
Add subheader menus to the pages that are missing them

bitWolfy:
This is a pretty simple fix that restores subheaders on the pages that were missing them.

comparison

It's not a critical issue, but it's somewhat irritating from a user experience.
Without this, the site layout shifts up and down when navigating between certain pages.

Most affected pages get a blank header, just to properly space things out:
- `iqdb_queries`
- `post_appeals`
- `static/404`
- `static/access_denied`
- `static/bookmarklet`
- `static/contact`
- `static/discord`
- `static/error`
- `static/keyboard_shortcuts`
- `static/site_map`
- `static/terms_of_service`
- `static/theme`
- `stats`

Some pages inherited sheaders from other relevant pages:
- `maintenance/user/api_keys`  [users]
- `maintenance/user/count_fixes` [users]
- `maintenance/user/deletions` [users]
- `maintenance/user/dmail_filters` [dmails]
- `maintenance/user/email_changes` [users]
- `maintenance/user/passwords` [users]
- `moderator/post/posts/confirm_delete` [posts]
- `moderator/post/posts/confirm_move_favorites` [posts]
- `post_events` [posts]
- `static/takedown` [takedowns]
- `tag_corrections` [tags]
- `tag_type_versions` [tags]
- `users/upload_limit` [users]

Other pages were given custom subheaders:
- `admin/alias_and_implication_imports`
- `admin/dashboards`
- `admin/exceptions`
- `comment_votes`
- `deleted_posts`
- `edit_histories`
- `mod_actions`
- `moderator/dashboards`
- `moderator/ip_addrs`
- `moderator/post/disapprovals`
- `post_approvals`
- `post_flags`
- `post_replacements`
- `post_votes`

### DIFF
--- a/app/views/admin/_secondary_links.html.erb
+++ b/app/views/admin/_secondary_links.html.erb
@@ -1,8 +1,12 @@
 <% content_for(:secondary_links) do %>
   <menu>
     <%= subnav_link_to "Dashboard", admin_dashboard_path %>
+    <li>|</li>
     <%= subnav_link_to "Exceptions", admin_exceptions_path %>
     <%= subnav_link_to "Staff Notes", admin_staff_notes_path %>
     <%= subnav_link_to "BUR Import", new_admin_alias_and_implication_import_path %>
+    <li>|</li>
+    <%= subnav_link_to "Post  Votes", controller: "/post_votes", action: "index" %>
+    <%= subnav_link_to "Comment Votes", controller: "/comment_votes", action: "index" %>
   </menu>
 <% end %>
--- a/app/views/admin/_secondary_links.html.erb
+++ b/app/views/admin/_secondary_links.html.erb
@@ -1,0 +1,8 @@
+<% content_for(:secondary_links) do %>
+  <menu>
+    <%= subnav_link_to "Dashboard", admin_dashboard_path %>
+    <%= subnav_link_to "Exceptions", admin_exceptions_path %>
+    <%= subnav_link_to "Staff Notes", admin_staff_notes_path %>
+    <%= subnav_link_to "BUR Import", new_admin_alias_and_implication_import_path %>
+  </menu>
+<% end %>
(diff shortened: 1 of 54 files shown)


kiranoot:
I am mixed on this idea. While it offers uniformity, it also needlessly pushes page contents down when it doesn't need to, and looks really weird when nothing is offered to the user in that location.

bitWolfy:
The subheader does not push the page down by much. In my opinion, not having the page shift around when navigating the site is worth it.

To prevent there just being a huge blue blob with nothing in it, I could just put the page title in there.

sitemap

It does not really do much for the user either, but it does fill the empty space.

bitWolfy:
Most non-static pages should have proper filled-in subheaders now.
The only exceptions are `post_appeals` (because it's broken) and `iqdb_queries` (because I'm not sure where it should go).

I've played around with static pages a bit, and had the idea of maybe linking other static pages in the subheader
sitemap
Does not look half bad, I think.

Earlopain:
In my opinion adding useful links to already existing headers or adding headers with links to pages which didn't have them before is fine. But I do agree that the blank header doesn't look very good.

Personally I didn't notice (or mind) that some pages have headers and some not. To be fair I don't use most of these pages regularly but some I do use more often. Pages like `confirm_delete` or `count_fixes` also don't really need a header since I'm there to do one thing and one thing only.

I'm also not sure if the sitemap needs a header. Every link there is already present on that page anyways. "Contact" and "Terms of Service" are also already present in every footer.

bitWolfy:
I am not arguing with the fact that the empty header does not look very good. That much is indisputable.

Should I just remove them empty headers altogether then?
Honestly, the rest of them seem pretty reasonable to me. While pages like `confirm_delete` don't necessarily _need_ a subheader, having one does not hurt them either. The links in the subheader are at least _somewhat_ relevant to the page.

PR #324, opened 2021-10-01
Flag page redesign

bitWolfy:
This is a redesign of the flag page that I mocked up and proposed a few months ago.

flagpage update

This adds detailed explanations of what each flag reason means specifically, as well as streamline the whole page somewhat.
My hope is that this will reduce the number of people who use flags incorrectly.

Since the header text is imported from a [wiki page](https://e621.net/wiki_pages/help:flag_notice), rather than specified in the settings, that page will have to be edited manually to complete the change if this pull request is accepted.
```
[quote]
h3. Abusing this system by willfully flagging posts for incorrect reasons may lead to disciplinary actions against your account.
[/quote]

Select an option below that most closely matches the reason why you believe this post should be deleted. Keep in mind that personal dislike of specific themes that may appear in images – such as gore, rape, or cub – are not valid reasons to flag the post on their own. Use the [[e621:blacklist|blacklist]] to filter out images that contain these tags.

If a reason is not listed here, check the options listed on the Report page.

If you are the owner of this submission, and would like to have it deleted (and not just replaced with an updated version), then please file a "takedown request":/static/takedown instead.
```

### DIFF
--- a/config/danbooru_default_config.rb
+++ b/config/danbooru_default_config.rb
@@ -681,6 +681,10 @@ fart'
       ]
     end
     
+    def flag_reason_48hours
+      "If you are the artist, and want this image to be taken down [b]permanently[/b], file a \"takedown\":/static/takedowns instead.\nTo replace the image with a \"fixed\" version, upload that image first, and then use the \"Duplicate or inferior version\" reason above.\nFor accidentally released paysite or private content, use the \"Paysite, commercial, or private content\" reason above."
+    end
+    
     def deletion_reasons
       [
         "Inferior version/duplicate of post #%PARENT_ID%",
@@ -711,9 +715,6 @@ fart'
         "[[conditional_dnp|Conditional DNP]] (Only the artist is allowed to post)",
         "[[conditional_dnp|Conditional DNP]] (%OTHER_ID%)",
       ]
-    end
-    def flag_reason_48hours
-      "If you are the artist, and want this image to be taken down [b]permanently[/b], file a \"takedown\":/static/takedowns instead.\nTo replace the image with a \"fixed\" version, upload that image first, and then use the \"Duplicate or inferior version\" reason above.\nFor accidentally released paysite or private content, use the \"Paysite, commercial, or private content\" reason above."
     end
 
     # Any custom code you want to insert into the default layout without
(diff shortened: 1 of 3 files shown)


kiranoot (approved):
:100: lgtm @Earlopain this look good to you too?

config/danbooru_default_config.rb
```diff
@@ -649,14 +649,47 @@ def enable_signups?
...
+          {
+            name: 'inferior',
+            reason: "Duplicate or inferior version of another post",
+            text: "A superior version of this post already exists on the site.\nThis may include images with better visual quality (larger, less compressed), but may also feature \"fixed\" versions, with visual mistakes accounted for by the artist.\nNote that edits and alternate versions do not fall under this category.",
+            parent: true
+          },
       ]
     end
+    def flag_reason_48hours
+      "If you are the artist, and want this image to be taken down [b]permanently[/b], file a \"takedown\":/static/takedowns instead.\nTo replace the image with a \"fixed\" version, upload that image first, and then use the \"Duplicate or inferior version\" reason above.\nFor accidentally released paysite or private content, use the \"Paysite, commercial, or private content\" reason above."
```
Earlopain:
```suggestion
      "If you are the artist, and want this image to be taken down [b]permanently[/b], file a \"takedown\":/static/takedown instead.\nTo replace the image with a \"fixed\" version, upload that image first, and then use the \"Duplicate or inferior version\" reason above.\nFor accidentally released paysite or private content, use the \"Paysite, commercial, or private content\" reason above."
```
The current link 404s, the same is true for the wiki page template. Once that's fixed this should be all good.

kiranoot:
With the other pull request, this one now has a conflict with the default config. Could you please rebase this back onto master to resolve that please?

Earlopain (changes requested):

config/danbooru_default_config.rb
bitWolfy: Should be fixed in 0c99e1209507bf633783b3e2b8214ef0ec50b855.

Earlopain:
> :100: lgtm @Earlopain this look good to you too?

Should be good to go

PR #325, opened 2021-10-03
[Search] Add pending replacements field

Earlopain:
Makes it possible to search for pending replacements with the search system via `pending_replacements:true`.

This adds a new elasticsearch index to store this information. There is also a migration script which adds this field, that should be run after this is deployed. To my understanding the script can be run while the site is operating. There don't seem to be any negative effects if the index is missing in the documents.

I only added the new field to the `ElasticPostQueryBuilder`, `PostQueryBuilder` looks like it should not recieve new features https://github.com/zwagoth/e621ng/commit/d8cffc4be132c8dc6f79fe4f1e556e28e1e91d45

### DIFF
--- a/app/indexes/post_index.rb
+++ b/app/indexes/post_index.rb
@@ -63,6 +63,7 @@ module PostIndex
         indexes :pending,         type: 'boolean'
         indexes :deleted,         type: 'boolean'
         indexes :has_children,    type: 'boolean'
+        indexes :has_pending_replacements, type: 'boolean'
       end
     end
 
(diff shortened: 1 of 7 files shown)


app/indexes/post_index.rb
```diff
@@ -258,6 +266,7 @@ def as_indexed_json(options = {})
       pending:        is_pending,
       deleted:        is_deleted,
       has_children:   has_children,
+      has_pending_replacements: options.key?(:has_pending_replacements) ? options[:has_pending_replacements] : replacements.pending.any?
```
kiranoot, line 269: Is there a reason that this has an override that bypasses the normal lookup? Is this an optimization?

Earlopain: I just did it the same way as the other fields which don't have direct database fields. The query gets executed during a single update through normal site usage while the optimized version is used during bulk importing. Since this is a boolean it doesn't look quite the same as the other ones.

This might actually just be unessesary since the replacements table probably won't grow to that large of a size. If you think so I can just remove it.

PR #326, opened 2021-10-03
Add webm uploader preview

### LINKED ISSUE #319: The upload preview does not work for videos
Currently it seems to try to put it into a `` which obviously will not work.

Earlopain:
This renders either a `` or `<video>`, depending on the guessed filetype.

This is my first time doing anything of substance with vue. There are probably things you don't do the way I did them here, which is why I'd like to get some input on this first.

Fixes #319


kiranoot:
It's bad practice in Vue to reach up into the parent element. It's designed as an escape hatch when passing props isn't working for some reason, or there is a direct need to interact with the parent/root as a Vue class object.

You should instead pass props into the element with what is needed from the parent(where the data currently is). If you want to group them into a single prop to pass you can either group them in the source data, or you can use a Vue computed property to have it pass a new object down and still obtain reactivity.

Elements are automatically re-rendered when any of their props change, and computed properties are automatically regenerated(and treated like they updated themselves) when any of the referenced variables change within the data section. It's a bit like magic, but it frees you from having to track data changes yourself.

This however is primarily done by tracking when variables are touched/read through `this` so don't bypass the data store if you can avoid it.

PR #330, opened 2021-10-07
Add a deletion reason selector

bitWolfy:
post_delete

A selector is created on the confirm_delete page with a bunch of prebuilt reasons.
I find a feature like this to be extremely useful, as it simplifies the deletion process and reduces the number of potential typos.

When the user clicks on a reason, it is added to the input field at the top. Clicking on the same reason again removes it.
Some deletion reasons have inputs, allowing for easier customization.

I have also moved the "merge favorites / tags / sources" checkboxes to be side by side with the parent preview, to save some space.

### DIFF
--- a/app/controllers/moderator/post/posts_controller.rb
+++ b/app/controllers/moderator/post/posts_controller.rb
@@ -11,7 +11,7 @@ module Moderator
       def confirm_delete
         @post = ::Post.find(params[:id])
         @reason = @post.flags.where(is_resolved: false)&.last&.reason || ''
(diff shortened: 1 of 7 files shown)


kiranoot:
Is there a reason that this was closed? I was going to confer with the other janitors about this, but it seems like a useful feature.

bitWolfy:
Feel free to re-open it if you want.
I don't really care about sharing this feature with others too much – everyone else has been doing fine without it. I can still use it through my userscript, so it does not affect me.

I simply figured that maybe reducing the number of open pull requests from me might mean that _some_ of them would have a chance to be accepted.
Oh, and the commits also got fucked up slightly.

PR #332, opened 2021-10-11
[Autocomplete] Open the tag's wiki page on middle click

bitWolfy:
This is a pretty simple change, made by request.

mammal

Instead of opening a tag search when middle-clicking on a tag in autocomplete, a corresponding wiki page is opened instead.
This is done by listening to an `auxclick` event in order to preserve the original link, so that it would still be possible to right-click it, and open the search for that tag in a new tab.

### DIFF
--- a/app/javascript/src/javascripts/autocomplete.js.erb
+++ b/app/javascript/src/javascripts/autocomplete.js.erb
@@ -256,6 +256,15 @@ Autocomplete.render_item = function(list, item) {
   $link.on("click.danbooru", function(e) {
     e.preventDefault();
   });
+  
+  // Open the wiki page on middle-click
+  if(item.type === "tag") {
+    $link.on("auxclick", function(e) {
+      if (e.button !== 1) return;
+      e.preventDefault();
+      window.open("/wiki_pages/show_or_new?title=" + encodeURIComponent(item.value), "_blank");
+    });
+  }
 
   if (item.antecedent) {
     var antecedent = item.antecedent.replace(/_/g, " ");


kiranoot:
I would prefer to have the behavior be consistent, whatever it is. Overloading default browser actions is considered bad practice, and we're already doing too much of it here with the link action being different from the click action. I agree that for the sake of auto complete it might make more sense to point them at the wiki page, but I also have to consider the areas that these can show up on the site, which is from the search dropdown, from tag entry fields on the uploader/edit page, and potentially some other tag edit areas.

bitWolfy:
The tag autocomplete appears in the following areas:
- General tag search form (`app/views/posts/partials/common/_search.html.erb`)
- Mobile-specific search form (`app/views/posts/index.html.erb`)
- Home page search form (`app/views/static/home.html.erb`)
- Tag script input (`app/views/posts/partials/index/_mode_menu.html.erb`)
- Post editing form (`app/views/posts/partials/show/_edit.html.erb`)
- Tag editing form (`app/javascript/src/javascripts/tag_editor.vue`)
- Upload form (`app/javascript/src/javascripts/uploader.vue`)
- Artist search form (`app/views/artists/_quick_search.html.erb`)
- Artist URLs search (`app/views/artist_urls/index.html.erb`)

Would it make sense to just replace the default link for the tag autocomplete with a wiki link here?https://github.com/zwagoth/e621ng/blob/ffd2de03e927610bd88090a0170a7c8bb54a2944/app/javascript/src/javascripts/autocomplete.js.erb#L305-L317
Something like:
```JS
case "tag":
    return "/wiki_pages/show_or_new?title=" + encodeURIComponent(item.value);
```

The result is pretty much the same, the only difference is that the ability to open the search with that tag in a new tab by right-clicking the item in the dropdown is lost. But then again, being able to open the wiki page is more important than that, in my opinion.

Earlopain:
I thought about this myself sometime ago and few of those places would definitely be improved if they link to a wiki page instead, especially when uploading or editing tags. I always had to open the wiki from the search page when uploading, which was also the reason I made it possible to open the wiki page from the uploaders tag preview window.

Not sure about the rest, I don't think I personally click on them all that often when simply  searching, or if anyone else does. Opening the wiki looks like the more useful version to me.
You can still click on "view posts" from the wiki page to start a search if you want.

PR #333, opened 2021-10-14
Add pre-built deletion reason selector

bitWolfy:
post_delete

A selector is created on the confirm_delete page with a bunch of prebuilt reasons.
I find a feature like this to be extremely useful, as it simplifies the deletion process and reduces the number of potential typos.

When the user clicks on a reason, it is added to the input field at the top. Clicking on the same reason again removes it.
Some deletion reasons have inputs, allowing for easier customization.

I have also moved the "merge favorites / tags / sources" checkboxes to be side by side with the parent preview, to save some space.

### DIFF
--- a/app/javascript/src/styles/specific/post_delete.scss
+++ b/app/javascript/src/styles/specific/post_delete.scss
@@ -1,4 +1,4 @@
-#reason { width: 100%; }
+#delreason-input { width: 100%; }
 #delreason-clear { width: 5rem; }
 
 .post_delete_options {
--- a/app/javascript/src/styles/specific/post_delete.scss
+++ b/app/javascript/src/styles/specific/post_delete.scss
@@ -1,62 +1,69 @@
-#delreason-input { width: 100%; }
-#delreason-clear { width: 5rem; }
+div#c-confirm-delete {
+  
+  // Raw text input and clear-all button
+  #delreason-input { width: 100%; }
+  #delreason-clear { width: 5rem; }
 
-.post_delete_options {
-  display: flex;
-}
+  // "Move to parent" checkboxes
+  .post_delete_options {
+    display: flex;
+  }
 
-#delreason-prebuilt {
-  a.delreason-button {
-    display: block;
-    cursor: pointer;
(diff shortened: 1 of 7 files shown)


app/views/moderator/post/posts/confirm_delete.html.erb
```diff
@@ -7,31 +7,56 @@
...
+            .gsub("%OTHER_ID%", "<input type=\"text\" />")
+            .html_safe
+      %>
+    </a>
+  <% end %>
+<% end %>
+</div>
+
+<div id="deletion-reason-suggestions"></div>
+<%= javascript_tag nonce: true do -%>
```
kiranoot: In order to keep on page javascript to a minimum, the usual method of doing this is to check the page ids on page ready and determine if the code should run. The reason this is done is because the main pack might load after the the page has finished parsing, and if that happens then this code runs before the main pack exists and has executed.

Trying to prepare for deferred loading of JS as much as possible.

The other routines have example code for how this is done. It's a bit messy and has some problems with execution order for shared pieces, but it keeps JS all in one place.

kiranoot (changes requested):
LGTM except for a small change related to how JS is initialized.

app/views/moderator/post/posts/confirm_delete.html.erb
bitWolfy: Should be fixed in e5a19272bd87a086f7d18de092a2e0dcc642bdaa.

app/javascript/src/styles/specific/post_delete.scss
```diff
@@ -0,0 +1,62 @@
+#reason { width: 100%; }
```
kiranoot: Can this ID be changed to something less generic, so that it doesn't cause surprises in the future if another element gets attached to an ID like this.

bitWolfy: Of course, my bad.
I'll also wrap the whole thing in `div#c-confirm-delete {}`, just in case.

PR #339, opened 2021-10-24
[Posts] Make the search terms in the post navbar into a link

bitWolfy:
navbar

The search terms in the navbar on the post page are currently being returned as plain text.
This change turns them into a link back to the search page with the specified tags.

It's especially useful on the mobile version of the site, since it lacks the searchbar on the post page.


Earlopain:
Shouldn't `Search:` also be part of the link? It's that way for pools and sets.
image

bitWolfy:
That's a good point. Although, I don't know if it would be better to include "Search" in the link, or to exclude "Pool" and "Set".

rabvan

Is there any specific reason why they would be included in the link?
Does it even make a difference?

Earlopain:
For me it can go this or that way, I just wanted to point it out. If you pressured me to favor one or the other I would probably go with the current version since it colors the text uniformly.

PR #344, opened 2021-11-01
[UI] Add formatting buttons to DText inputs

bitWolfy:
This adds a wrapper for DText-enabled textareas, with formatting buttons and built-in preview.

formatting-helper

The wrapper is generated via JavaScript. In order to add it to a textarea, simply add the `dtext-previewable` class to it.
This approach simplifies things, especially when dealing with special cases, like the upload page.

There are 8 formatting buttons total, although new ones can be added fairly easily:
* Bold
* Italics
* Strikethrough
* Underline
* Header
* Spoiler
* Code
* Quote

This wrapper also includes a character counter, and a way to notify the user that they are approaching the limit.
Previously, all limits were set up using magic numbers in the models. Those have been moved to the config file instead.


kiranoot:
Only change I would request is that when clicking to add a formatting tag, the tag is not selected upon insertion, and the cursor is placed between the open and close tag.

Issue #318, opened 2021-09-28
Blacklist does not work on videos


Earlopain:
They are  filtered out fine when searching but if you somehow end up on a post anyways toggling the blacklist simply pauses the video and seeks to the start instead of showing the blacklist placeholder.

bitWolfy:
This is actually the exact opposite problem to #319, funnily enough.
Here, the blacklist tries to jam an image file into a `<video>` element.

IMHO, it would be easier if the blacklisting method was jus replaced wholesale.
Instead of trying to change the image / video source, just hide that element, and display the placeholder image in its own dedicated `<img>`.

PR #345, opened 2021-11-03
Fix blacklist not applying on videos

Earlopain:
Like bitwolfy sugested in #318. This removes a bunch of code related to switching back and forth the `src` attribute. Instead it simply hides the image and displays the blacklist notice seperatly. Should you for some godforsaken reason have no javascript available all posts will simply display, just like it was beforehand.

Seems like I also accidentally fixed blacklist toggling on wiki pages with this. Maybe also in other places, who knows. Also fixes a display error when hiding deferred posts, they were not accounted in the blacklist counter.

### DIFF
--- a/app/javascript/src/javascripts/blacklists.js
+++ b/app/javascript/src/javascripts/blacklists.js
@@ -1,5 +1,6 @@
 import Utility from './utility'
 import LS from './local_storage'
+import Post from './posts';
 
 let Blacklist = {};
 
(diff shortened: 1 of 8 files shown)


config/webpacker.yml
```diff
@@ -10,7 +10,8 @@ default: &default
 
   # Additional paths webpack should look up modules
   # ['app/assets', 'engine/foo/app/assets']
-  additional_paths: []
+  additional_paths:
+    - public
```
Earlopain, line 14: For some reason needed for `background: url("images/blacklisted-preview.png");` in scss.

app/javascript/src/javascripts/blacklists.js
```diff
@@ -192,8 +117,8 @@ Blacklist.postHide = function (post) {
 
 Blacklist.postShow = function (post) {
   const $post = $(post);
-  Blacklist.postRestoreSrc(post);
-  $post.addClass("blacklisted").removeClass("blacklisted-active").removeClass("blacklisted-active-visible");
```
Earlopain, line 196: There didn't seem to be any reason for having seperate classes when an item can be blacklisted and when it actually is, so I removed them.

app/javascript/src/javascripts/blacklists.js
```diff
@@ -192,8 +117,8 @@ Blacklist.postHide = function (post) {
...
+  $post.removeClass("blacklisted");
+  Post.resize_notes();
```
Earlopain, line 121: When the blacklist is active on load notes are scaled against 0x0 so I resize them here after the post has it's size restored.

kiranoot:
It seems like this would break pages like pool pages and pool order pages where there is a reliance on the elements being visible for understanding by users.

Earlopain:
My bad, I didn't think about them. I'll take a look at how it behaves currently and bring that behaviour over.

Earlopain:
Should be good now. My understanding is that blacklisted posts should only be hidden on the post index page. I don't really see a point for that to happen on other pages. This makes it easier to fix the pool issue you described, since both pages use `.post-preview`

PR #347, opened 2021-11-05
Use .gitattributes to force the equivalent of core.autocrlf=false

ShadyGuy123:
I spent almost a week trying to get vagrant to work, and this was the (main) problem.
Thanks to @Earlopain for the help.
To avoid anyone else having this problem, it's possible to override the core.autocrlf behaviour. I learned about it from [Gang-Garrison-2/.gitattributes](https://github.com/Gang-Garrison-2/Gang-Garrison-2/blob/master/.gitattributes), the explanation is taken verbatim from there too

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# This "unsets" the git "text" attribute on all files, which tells git to
+# not do any line-ending conversion between LF and CRLF on check-in and
+# check-out. This is like core.autocrlf=false, but doesn't require the user to
+# edit their git config, instead it overrides it.
+* -text


Earlopain:
While we're at it, shouldn't we just force lf line endings no matter which OS you are on? My understanding of this is that if someone where to commit crlf line endings they would be taken as such.

There aren't many files with crlf line endings in the repo right now, only a few migration scripts and two or three actual source files. I googled a bit and I this should do the trick `* text=auto eol=lf`, but after newly commiting one of those crlf files they still had their original line endings. There's  a `.editorconfig` specifying lf but actually forcing it by git sounds better.

ShadyGuy123:
Sure, we can force eol=lf just as you said, then we can renormalize all the line endings with `git add . --renormalize`.
These files will get renormalized:
- Troubleshooting.md
- app/javascript/src/javascripts/post_delete.js
- public/vendor/jquery-3.4.1.min.js &nbsp; &nbsp; &nbsp;and&nbsp; &nbsp; &nbsp; public/vendor/vue.js
- All scripts in script/e6migration except for&nbsp; &nbsp; &nbsp; 01_artists,&nbsp; &nbsp; &nbsp; 11_posts,&nbsp; &nbsp; &nbsp; 13_users,&nbsp; &nbsp; &nbsp; 22_favorites

bitWolfy:
> These files will get renormalized:
> 
> * Troubleshooting.md
> * app/javascript/src/javascripts/post_delete.js

These might be my fault, I apologize.
Forcing the lf line endings seems like a good idea to me too.

PR #348, opened 2021-11-06
[DText] Prevent zalgo from overflowing

Earlopain:
Basically the same as #309 but it doesn't try to adress deeply nested dtext. I also renamed the css class which I didn't do previously for some reason.

Zalgo no longer extends top/bottom and a long string of `a`s no longer extends to the right. Trying to fix the nested dtext introduced more problems that I was willing/capable to fix.

image

#344 or this should be rebased once one gets merged as I don't know how well it will play together.


Earlopain:
Seem to work fine with the formatting changes. I just needed to replace the text class and it was good to go.

PR #356, opened 2021-11-20
[DText] Fix the bugged preview window in the upload form

bitWolfy:
This is a relatively minor fix to a bug that exists on the upload page.
The existing style sets all textarea fields on that page to `display: inline-block`, which prevents it from being hidden with `display: none` when looking at the preview.

unknown

This fix simply adds `!important` to `display: none` to prevent issues like this.

### DIFF
--- a/app/javascript/src/styles/common/dtext.scss
+++ b/app/javascript/src/styles/common/dtext.scss
@@ -285,7 +285,7 @@ div.dtext-formatter {
         background: #00000040;
       }
     }
-    div.dtext-formatter-preview { display: none; }
+    div.dtext-formatter-preview { display: none !important; }
     div.dtext-formatter-charcount { color: black; }
   }
   &[data-editing="false"] {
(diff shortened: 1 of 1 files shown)


Earlopain:
Aren't textareas `inline-block` by default? I just played around with it for a bit, there doesn't seem to be any difference when removing that instead.

bitWolfy:
They should be, but they get reset to `display:block` by [forms.scss](https://github.com/zwagoth/e621ng/blob/master/app/javascript/src/styles/common/forms.scss#L12-L16).
I am not entirely sure why. Meanwhile, [upload.scss](https://github.com/zwagoth/e621ng/blob/master/app/javascript/src/styles/specific/uploads.scss#L90) resets it back. It even says that setting it to `block` breaks everything, although I can't say for sure what exactly is broken.

I chose this solution instead of simply removing the `display` style from uploads.scss because it would work in all cases where the textarea is given a specific display style, not just on the upload page.

kiranoot:
The display: block in the forms forces form elements to be on their own line, everywhere, which is not typically how it works. So it created chaos when you attempt to use them inline, or even just have behavior that matches how they normally behave. There is a class to force them inline, but it sets them as inline and not inline-block and that is also the wrong behavior. It's a mess, and I didn't want to bother sorting it all out.

PR #357, opened 2021-11-24
Danbooru cleanup

Earlopain:
This looks rather large but not much is actually happening here. I always hated how the tests are mostly useless since there were so many failing you had no chance to see if you actually broke something. Looking at the assertion count also doesn't help you because test failures vary from run to run.

Here are some numbers where I am at right now. Execution time stayed mostly the same. I'll update this as I repair more tests. 
master:  `1554 runs, 1994 assertions, 166 failures, 394 errors, 30 skips`
this pr:  `1249 runs, 2268 assertions, 0 failures, 0 errors, 0 skips`

In order to make this work I removed a bunch of danbooru specific code and features that are disabled or unused, including their tests. A large portion of these changes are just removing unsused functionality, removing now unused functions, removing views, removing even more newly unused functions etc.

I removed:
- Source Strategies
- Recommender Service
- Reportbooru
- Reports
- Unused maintanance tasks
- Artist Commentaries
- Post Appeals
- Spam detector
- Ugoira
- Artist Groups
- And a bunch of other smaller things

I image that most of these didn't work anymore anyways since it has been two years since anything has been done to them while the codebase kept evolving. Removing their tests reduced failures/errors by a significant amount already.

Elasticsearch indicies are now prefixed with the environment when not running in production. Sometimes my search broke on development which I think was because it shared the index with test. This might fix this, though I'm not 100% sure. It does at least fix some tests which were failing previously.
I'm not sure how you are supposed to test against elasticsearch, there are still tests which I believe are failing because of it. Droping/Recreating the index on each test would certainly work but it also increases the run duration by way to much.

If you have a local config it now doesn't apply to tests anymore. Disabling limits for development should not lead to broken test.

I also added a new docker service you can use to run the tests more conveniently. Might also be usable on some kind of CI, who knows
`docker-compose run test` runs the whole suite.
`docker-compose run test TEST=unit/models/post_test.rb` runs the tests of only a single file.
`docker-compose run test TEST=unit/models/post_test.rb:42` runs the specific test at that line.


Earlopain:
Finally got it all working, I've now got zero errors/failures when running all tests, consistently too. Test time increased by ~30 seconds overall for me to 220 seconds for a full run. Probably because of waiting for ealsticsearch when creating/destroying a post which didn't happen before.

image

app/logical/sources/alternates/pixiv.rb
```diff
@@ -14,6 +14,10 @@ def domains
         ["pixiv.net", "pximg.net"]
       end
 
+      def headers
```
kiranoot, line 17: This value is filled, but does not appear to be used within this commit.

app/logical/downloads/file.rb
```diff
@@ -28,8 +26,7 @@ def size
 
     def download!(tries: 3, **options)
       Retriable.retriable(on: RETRIABLE_ERRORS, tries: tries, base_interval: 0) do
-        file = http_get_streaming(uncached_url, headers: strategy.headers, **options)
-        return [file, strategy]
+        http_get_streaming(uncached_url, **options)
```
kiranoot, line 31: strategy no longer providers headers?

app/logical/d_text.rb
```diff
@@ -2,23 +2,11 @@
 require 'uri'
 
 class DText
-  MENTION_REGEXP = /(?<=^| )@\S+/
```
kiranoot, line 5: Part of wrong commit?

app/models/post.rb
```diff
@@ -96,16 +96,6 @@ def move_files_on_undelete
       Danbooru.config.storage_manager.move_file_undelete(self)
     end
 
-    def distribute_files(file, sample_file, preview_file)
```
kiranoot, line 99: Pretty sure this is still used by the uploader.

test/unit/post_flag_test.rb
```diff
@@ -4,144 +4,79 @@ class PostFlagTest < ActiveSupport::TestCase
...
-
-        assert_difference(-> { PostFlag.count }, 0) do
-          as(@bob) do
-            @post_flag = PostFlag.create(post: @post, reason: "aaa", is_resolved: false)
-          end
-        end
-        assert_equal(["Post is deleted"], @post_flag.errors.full_messages)
+      as(@users.first) do
+        @flag1 = create(:post_flag, post: @post)
+        puts @flag1.errors.full_messages
```
kiranoot: leftover debug code?

docker/danbooru_local_config.rb
```diff
@@ -1,5 +1,5 @@
 module Danbooru
-  class CustomConfiguration < Configuration
```
kiranoot: This changes the semantics of how this class works. It no longer has all of the methods from the base configuration and moves all the logic into the surrounding wrapper. It seems like it might make sense to allow configuration even during tests, but to spot fill areas that specifically would cause tests to fail during global setup.

test/functional/admin/users_controller_test.rb
```diff
@@ -17,13 +17,22 @@ class Admin::UsersControllerTest < ActionDispatch::IntegrationTest
 
     context "#update" do
       context "on a basic user" do
-        should "succeed" do
-          put_auth admin_user_path(@user), @mod, params: {:user => {:level => "30"}}
-          assert_redirected_to(edit_admin_user_path(@user))
+        should "change the leve" do
```
kiranoot: typo

kiranoot (changes requested):
Need to finish up this but sleep first.

app/models/post.rb
Earlopain: The uploader uses `Utils.distribute_files`, not the member method from the post model. This function was only used in rake tasks and fixer scripts.

app/logical/sources/alternates/pixiv.rb
Earlopain: It's used in https://github.com/zwagoth/e621ng/blob/65fed7a800ca5e3adc9f22255cc136ccc290213a/app/logical/downloads/file.rb#L80. It was still called strategy there but I renamed it in my latest commit.

app/logical/downloads/file.rb
Earlopain: The header variable was simply unused. The headers are set in `http_get_streaming` and are provided by  `httparty_options`, which includes the headers

app/logical/d_text.rb
Earlopain: Seems like it. Should I move it out or just leave it?

docker/danbooru_local_config.rb
Earlopain: I reverted this. Fixing this up would probably be another larger undertaking.

PR #364, opened 2022-01-07
Post events log

### LINKED ISSUE #346: The post events page is incomplete
I'm talking about pages like this one: https://e621.net/posts/897473/events

It would be really nice to have **all** the moderation events listed there, with timestamps.
* Post deleted
* Post undeleted
* Post approved
* Post unapproved
* Post rating locked
* Post flag created

All of those are already available through mod actions, but it's not possible to search for individual posts there.
I believe that this information should be publicly accessible, both for transparency / accountability, and to be able to establish a timeline of events if there is some confusion about what happened to a post.

Earlopain:
Fixes #346 

Moves a bunch of post related events from ModActions into their own table, where they can be identified by post_id. I also added a few things which were not logged before, like unflagging. We can also expose this information through json, since the only thing that needs to be hidden is the creator on flags, which is easy to implement. Note that there is no foreign keys to posts in order to store information about destroyed posts.

Contains a fixer script which tries to fill the table with information from current data, like approvals/flags/deletions and a few modactions. When inserted ordering by id != ordering by created_at,  I still need to think about how to do that. Maybe a temporary table would be best for that.

image


app/models/post_event.rb
```diff
@@ -1,71 +1,78 @@
...
   end
 
-  # XXX can't use hidden_attributes because we don't inherit from ApplicationRecord.
-  def serializable_hash(**options)
-    hash = super
-    hash = hash.except(:creator_id) unless is_creator_visible?
-    hash
+  def hidden_attributes
+    hidden = super + [:extra_data]
+    hidden += [:creator_id] unless is_creator_visible?(CurrentUser.user)
```
kiranoot: This creates a strange change in attribute shape based on type. It would be better to define a serializer and defined the shape and fill this with nulls when it isn't exportable.

app/models/post_event.rb
```diff
@@ -1,71 +1,78 @@
-class PostEvent
-  include ActiveModel::Model
-  include ActiveModel::Serializers::JSON
-  include ActiveModel::Serializers::Xml
+class PostEvent < ApplicationRecord
+  belongs_to :creator, class_name: "User"
+  before_validation :initialize_creator, on: :create
```
kiranoot: We should start normalizing passing user objects into models when possible. This pattern creates too many implicit sets, and complicates tests.

app/models/post_event.rb
```diff
@@ -1,71 +1,78 @@
...
-    event.try(:is_resolved) || false
-  end
+    if params[:post_id].present?
+      q = q.where("post_id = ?", params[:post_id].to_i)
+    end
 
-  def creator_id
-    event.try(:creator_id) || event.try(:user_id)
-  end
+    if params[:creator_name].present?
```
kiranoot, line 52: These two search fields bypass the protection on flags. The objects are still returned, but the field would be missing visually.

kiranoot (changes requested):

app/models/post_event.rb
Earlopain: This is not true. When someone is searching by username/id and cannot see the flagger, those actions are filtered out directly via sql. If you are only searching by post they are returned, but the name is not displayed. I have also written tests in `post_events_controller_test.rb` to confirm this behaviour for both the html and json response. I added a few more tests for that behaviour in my latest commit.

https://github.com/zwagoth/e621ng/blob/c8bbfb93b9a18671e1100a8f6791e9fdaba8e161/app/models/post_event.rb#L37-L43

app/models/post_event.rb
Earlopain: Done. They are still comming in from the calling model, but changing that all the way to the top is something for another day

kiranoot (approved):

kiranoot:
If you are happy with the ordering of the events in this. I am ready to pull it.

kirwinia (approved):
#364 Great changes, I definitely want these!

script/fixes/105_modaction_to_postevents.rb
```diff
@@ -62,3 +78,7 @@ def create(post_id, creator_id, created_at, action, extra_data = {})
     end
   end
 end
+
+ActiveRecord::Base.connection.execute('INSERT INTO post_events (creator_id, post_id, "action", extra_data, created_at)
```
kiranoot: This needs a without timeout block around it, or it's likely to fail in production. The prod server has fairly strict global statement timeouts applied to it.

script/fixes/105_modaction_to_postevents.rb
```diff
@@ -2,8 +2,24 @@
...
+);')
+
+ActiveRecord::Base.connection.execute('CREATE INDEX index_post_events_temp_on_created_at ON public.post_events_temp USING btree (created_at);')
+
 def create(post_id, creator_id, created_at, action, extra_data = {})
-  PostEvent.create!(post_id: post_id, creator_id: creator_id, created_at: created_at, action: action, extra_data: extra_data)
+  PostEventTemp.create!(post_id: post_id, creator_id: creator_id, created_at: created_at, action: action, extra_data: extra_data)
 end
 
 PostApproval.in_batches do |batch|
```
kiranoot: This should likely be `find_in_batches` otherwise it will emit multiple queries per batch. the `find_in_batches` variant calls `in_batches` with load set to true so that it doesn't return an ActiveRecord::Relationship, but an array of results.

Earlopain: My bad, changed it to `find_in_batches`. After a quick google search it seems that in_batches runs two queries per batch.

kirwinia (approved):

PR #387, opened 2022-04-03
Theme support for related tags, autocompletion and backtraces

faucetlol:
While using the hotdog theme on April Fools', something hurt my eyes more than the contract between the red and the yellow - the fact parts of the website still show in the default blue. A before and after example is shown below.

e6

I also did the same for the autocompletion widget and backtraces. This is my first PR so I'm sorry if I did anything wrong :)


kiranoot:
LGTM

PR #391, opened 2022-04-07
[Uploader] Add trio to upload form

faucetlol:
It's been [at least a year](https://e621.net/forum_topics/27724) since the trio tag became a thing, and while it has over 45,000 usages it's currently underused due to the fact people select "group" on the upload form.

Trio


Earlopain:
Thank you

PR #392, opened 2022-04-09
Remove testparser

Earlopain:
I have no idea what this is for but there aren't much uses left, and they can all be removed over time. This is only used for a posts `tag_index` and wiki pages `body_index`.

This has to happen in order for `tag_index` to be removed:

- [ ] Remove all usages of `Post.sql_raw_tag_match`
- [ ] Remove all usages of `PostQueryBuilder`

I don't see a reason why searching wiki pages should behave differently from any other text search, so that can safely be removed in my opinion. Since elastic should be the only way to search for posts `tag_index` should be phased out either way.

TODO: Add a script to reindex `body_index`


kiranoot:
Danbooru figured out a solution for this that preserves the semantics but without any compiled extensions. Let me see if I can find the commit for it.

kiranoot:
https://github.com/danbooru/danbooru/commit/d50cfdb8560eba80069639cbe8bcb6772a6ad146

Earlopain:
I'm not in a hurry to get this in, removing the two things I listed is something I want to do sometime anyways. Since this doesn't need to be changed right now I'll just let this sit here until then.

PR #394, opened 2022-04-13
[Pools] Use the same search as other pages + add a heading

faucetlol:
The search form used on the pools page has been different to the other pages and it's been bothering me - so this PR brings it up to date with the other ones. This also fixes a bug where "Creator" currently uses pool autocomplete.

Also added a `<h2>` tag to the page, just like the post sets page.

image
image


Earlopain:
You don't actuallty need to write `value` and `selected` explicitly. This is something I made to happen automatically for you when using `form_search`. `value` how you did it doesn't actually do anything, since it would need to be inside `input_html`.
 https://github.com/zwagoth/e621ng/blob/master/app/inputs/search_form_builder.rb

faucetlol:
If they're not actually doing anything, I've gone ahead and removed it.

PR #396, opened 2022-04-25
Ruby 3.1

Earlopain:
Tests seem fine, but otherwise I didn't look into it yet.

Things of note:
* Bundler got udpated to the newest versions. Older versions print an annoying deprecation warning.
* net-smtp is not bundled by default anymore. Check if that's still needed with rails 7 (when we get there).
* Webrick is not bundled by default anymore. A new rails project would uses puma, so I did that here too. Only relevant for development, since prod used unicorn

I also removed a bunch of non-functional binstubs and everything from capistrano which I believe isn't used, correct me if I'm wrong.
@zwagoth iqdbs also uses some ruby version, does that need upgrading too or is that not important?


kiranoot:
iqdbs should be okay, it's own little thing. The prod setup allows for multiple ruby versions without issue.

Sidekiq and the jobs runners are the ones that need attention first.

We don't use capistrano because it doesn't fit our needs, it's fine to remove it.

It should be okay to include puma. Ideally it would use unicorn, but I understand that unicorn isn't as well suited for development as the built in support for puma.

PR #397, opened 2022-04-29
Disable downvoting own comment

JasonTheKitten:
Currently, users are not able to make upvote their own comments.
However, they _are_ able to downvote their own comment.

This behaviour seems inconsistent and does not make much sense, so I made a PR that also disabled the downvote button on your own comment, in order to make things more consistent.

I also made it so that the buttons are not shown on a user's own comment, because it doesn't make much sense to show the buttons if you are not able to use them.

P.S.
I was not able to perform unit tests because it complained that I had not set up the testing DB, and there were no instructions on how to do so.
I manually did testing via the localhost site though.

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -4,7 +4,7 @@ module CommentsHelper
     vote_score = voted ? vote.score : 0
     comment_score = comment.score
 
-    if CurrentUser.is_member? && comment.creator != CurrentUser.user
+    if CurrentUser.is_member?
       up_tag = tag.li(tag.a('&#x25B2;'.html_safe, class: 'comment-vote-up-link', 'data-id': comment.id),
                       class: confirm_score_class(vote_score, 1, false),
                       id: "comment-vote-up-#{comment.id}")
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -4,7 +4,7 @@ module CommentsHelper
     vote_score = voted ? vote.score : 0
     comment_score = comment.score
 
-    if CurrentUser.is_member?
+    if CurrentUser.is_member? && comment.creator != CurrentUser.user
       up_tag = tag.li(tag.a('&#x25B2;'.html_safe, class: 'comment-vote-up-link', 'data-id': comment.id),
                       class: confirm_score_class(vote_score, 1, false),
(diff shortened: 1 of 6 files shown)


Earlopain:
Hi there, I think it's best to still show the vote buttons for you own comments, even if they don't do anything. This makes it consistent with all the other comments, and displaying the error on vote makes it clear that it's intentional instead of some error.

Another nitpick: I would have merged the two branches `is_positive` and `is_negative`, I don't see the need for two seperate error messages in those cases. This is also how I did it for voting on your own forum requests.

No idea why tests don't run. I probably set that up manually a while ago. It should work after `docker-compose run e621 RAILS_ENV=test bin/rails db:setup`, or maybe it was just `db:create`. I can see if I can get that to happen automatically, that would be nice. Don't mind all the failing tests, I'm in the process of cleaning that up. You can run an individual test with `docker-compose run tests test/functional/file_name.rb:line_number`

JasonTheKitten:
For the buttons that don't do anything:
How about graying them out? Would that be a good idea?

JasonTheKitten:
The command you gave (`docker-compose run e621 RAILS_ENV=test bin/rails db:setup`) had to be altered some. `docker-compose run -e RAILS_ENV=test e621 bin/rails db:setup` ended up being the command that made unit tests work properly.

After doing this, the command that I was previously running works as expected.

JasonTheKitten:
I've run into a problem
Some other unit tests for other things rely on being able to downvote your own comment, it seems.

Earlopain:
I would prefer to just leave the buttons as is to be honest.

Whoops, yeah. That's not how you start docker commands with environment variables. But you seem to have figured it out anyways.

Earlopain:
> I've run into a problem
> Some other unit tests for other things rely on being able to downvote your own comment, it seems.

Either change the vote to a positive one, or just remove the test completely if it doesn't test anything else. You might also need to add a few more comment and votes for things to still make sense. I can fix up the the tests for this if you would prefer.

JasonTheKitten:
Ok, I got unit tests working too!

Earlopain:
There are still a few tests failing now:
* test/functional/comment_votes_controller_test.rb:19
* test/functional/comment_votes_controller_test.rb:26
* test/unit/user_test.rb:71
* test/functional/moderator/dashboards_controller_test.rb:95

After you fix those it'll be good to go.

PR #398, opened 2022-04-29
[API] Hacky optimization of API requests

kiranoot:
I'm not proud of this, but it works.

### DIFF
--- a/app/logical/post_sets/base.rb
+++ b/app/logical/post_sets/base.rb
@@ -40,6 +40,26 @@ module PostSets
       nil
     end
 
+    def fill_tag_types(posts)
+      tag_array = []
+      posts.each do |p|
+        tag_array = (p.tag_array + tag_array).uniq
+      end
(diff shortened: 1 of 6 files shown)


app/logical/post_sets/base.rb
```diff
@@ -40,6 +40,26 @@ def best_post
       nil
     end
 
+    def fill_tag_types(posts)
```
Earlopain, line 43: Shouldn't a call to just `Tag.categories_for` be enough already? This function already does category caching, though maybe you did it this way to also prevent a bunch of cache accessing? Is`inject_tag_categories` needed?

kiranoot: Tag.categories_for makes requests out to memcached, so there is latency for each post, called in a loop this results in 320 requests to the cache at about 0.5-1ms each. With an aggregate request, the cache fetch takes about 2-3ms and applies to all the posts in the response.

Earlopain (approved):

PR #403, opened 2022-05-05
[Posts] Use will-change: transform; when using vertical/horizontal fit

faucetlol:
Fixes poor image scaling on Chrome/Edge, and probably every other Chromium based browser. See also [forum #33513](https://e621.net/forum_topics/33513).

This probably isn't the most elegant solution, but it seems to fix the issue in Chromium-based browsers and doesn't seem to cause any issues in other browsers. Safari on iOS and OS X has not been tested

**Examples**
Left is original, right is with `will-change: transform;`. Make sure it's viewed at 100% size to actually see a difference.
examples
(image credit: [post #2042556](https://e926.net/posts/2042556), [post #692610](https://e926.net/posts/692610), [post #864033](https://e926.net/posts/864033), [post #896454](https://e926.net/posts/896454))

### DIFF
--- a/app/javascript/src/styles/specific/posts.scss
+++ b/app/javascript/src/styles/specific/posts.scss
@@ -491,10 +491,12 @@ textarea[data-autocomplete="tag-edit"] {
 #image {
   &.fit-window {
     max-width: 100%;
+    will-change: transform;
   }
   &.fit-window-vertical {
     max-width: 100%;
     max-height: 95vh;
+    will-change: transform;
   }
 }
 


Earlopain:
I don't think I understand this change. https://developer.mozilla.org/en-US/docs/Web/CSS/will-change describes it as an performance hint to the browser (though it may also affect the appearance, such as in this case?), and also heavily discourages using it. This can't be the solution to the problem, maybe there is something else worth checking out?

faucetlol:
> I don't think I understand this change. https://developer.mozilla.org/en-US/docs/Web/CSS/will-change describes it as an performance hint to the browser (though it may also affect the appearance, such as in this case?), and also heavily discourages using it. This can't be the solution to the problem, maybe there is something else worth checking out?

I found [a related question on Stack Overflow](https://stackoverflow.com/questions/37906602/blurry-downscaled-images-in-chrome) which suggests a few other solutions I'll check out shortly, but many of them just seem as hacky as this one and a few have replies saying it no longer works.

faucetlol:
I tested most the methods suggested on Stack Overflow

- `image-rendering: -webkit-optimize-contrast`: working in Chrome & Edge. Comments say that it ruins the image in Safari but I can't test that, however [somebody did use media queries to unset it on Safari](https://stackoverflow.com/a/69137897).
- `transform: translateZ(0);`: appears to have no effect at all in either Chrome or Edge
- `transform: translateZ(0) scale(0.999999);`: this... for some reason works on Chrome & Edge.
- `transform: translateZ(1px);`: this does something, but never really seems as sharp as the other options.
- `will-change: transform;`: also works in both, a comment says "This feature uses GPU to render images so it may slow down the browser." which might also be a concern for some of the larger images hosted on e621.

None of these seem to negatively impact Firefox users at all. Safari remains untested.

PR #407, opened 2022-05-14
[Uploader] Correct path to user upload limit

faucetlol:
Reported at [forum #334620](https://e621.net/forum_topics/25734?page=24#forum_post_334620). 

In one case the uploader will link to a non-existent help page rather than the users upload limit path, this PR fixes it.


Earlopain:
There's https://e621.net/wiki_pages/upload_limit which the help page probably linked to at some time, but this is better.

PR #410, opened 2022-05-16
Edit Basque, Lithuanian, Latvian welcomes

heyspacebuck:
1) Basque "ongi" would be like saying just "well" in English, it doesn't work as a greeting unless it's "Ongi etorri". Source: asked a native Basque speaker.
2) Latvian "gaidīts" means "welcome" as an adjective, but can't be used on its own as a greeting. The closest greeting would be "Esi gaidīts" but other options exist: "Laipni lūdzam" (more formal), "sveiki" (general greeting). Source: Asked a non-native but fluent Latvian speaker.
3) "Gaidīts" is not a Lithuanian word, Lithuanian doesn't use "ī". "Sveiki atvykę" is an option, or more casually "Sveiki" (but "sveiki" on its own is more like "hey" than "welcome"). Source: My uncle, don't tell him why I was asking about this <____<

Also, as far as I can tell, "Taliga" does not mean "Welcome" in Samoan. It means "ear." But I don't have anyone to confirm with so I'm not going to commit it here. source: http://www2.hawaii.edu/~leighj/SAMOAN/dictionary/

Thanks, takk, merci, danke!


Earlopain:
First two seem ok, but `Sveiki atvykę` means `Welcome to the` according to google translate. I'm going to trust your uncle on this

PR #413, opened 2022-05-19
Replacements uploader part 1

Earlopain:
First part of replacing the current replacements upoader with parts of the normal uploader. I didn't quite figure out how to properly integrate it yet, but I also haven't looked into it since we last talked about it.

Functionally this should work the same like it did before, it just has a few more components so that they may be reused in other places. I'm mainly looking for feedback on this code in general since I'm still new to vue.


app/javascript/src/javascripts/uploader/file_preview.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <div class="upload_preview_container" :class="classes">
+    <div class="box-section sect_red" v-show="overDims">
+      One of the image dimensions is above the maximum allowed of 15,000px and will fail to upload.
+    </div>
+    <div v-if="!failed">
+      <div class="upload_preview_dims">{{ previewDimensions }}</div>
+      <preview-video v-if="isVideo" :url="finalPreviewUrl"
```
kiranoot: Due to the tightly coupled nature of the isVideo and url variables, and the extreme similarity of the way the preview is displayed. I think it makes more sense to package these two variables up into an object that is emitted and then consumed by a single component, which internally makes the decision to display a video or image tag.

By having two variables you may run into subtle race conditions(it seems unlikely, but it isn't guaranteed not to happen), where passing around a new instance of an object each time would result in a single prop that changes instead of the internal members. This will always arrive as a single piece and be processed in the same render tick.

app/javascript/src/javascripts/uploader/preview_image.vue
```diff
@@ -0,0 +1,13 @@
+<template>
```
kiranoot: combine these two templates into a single component. Accept an object prop with the shape `{isVideo: boolean, url: string}`

kiranoot (changes requested):
Seems mostly fine. Some simplification can take place as noted above.

app/javascript/src/javascripts/uploader/preview_image.vue
Earlopain: Done, but I just moved both into the existing component. There wasn't much use for it to be separate anyways.

app/javascript/src/javascripts/uploader/file_preview.vue
Earlopain: I didn't know that, good to know. I added a validator to check the objects shape. Also makes it more explicit in the code what properties are usable/needed.

kiranoot (approved):

PR #416, opened 2022-05-21
Vue 3

Earlopain:
Because https://v2.vuejs.org/ keep reminding me that Vue 2 is in use. I followed the guide on https://v3-migration.vuejs.org/migration-build.html during the upgrade.

There were very few things that actually needed much changing. I converted the `tag-preview-tag` functional component into a sfc since functional components are removed. This could apparently have been a plain function, but I had trouble accessing the tag prop, so I did it this way (also easier to read in my opinion).

Vue 3 has a new functionallity called `<Teleport>` which might allow me to easily integrate the uploader for replacements. https://vuejs.org/guide/built-ins/teleport.html. I haven't looked to closely at it yet, but it sounds promising.


kiranoot:
I have only one hesitation on this project. Vue 3 has essentially dropped support for webpack for SFC when packaging projects and moved entirely over to vite as their build platform. This means that as this code ages, the ability to build it is going to degrade and eventually cease to exist if we use the SFC system. Unless there is something wrong with Vue 2, or some mythical security problem shows up in it, we should likely stick to 2 until Vite for rails starts to show better support.

I hate the JS ecosystem so much.

Earlopain:
After some discusion this is fine as is and we'll look into vite in the near future.

PR #421, opened 2022-07-05
Show all users upload whitelist on sitemap, add "Requests" and "Help" to the secondary links on the upload whitelist page.

### LINKED ISSUE #420: Upload Whitelists isn't on the sitemap
The [Upload Whitelists](https://e621.net/upload_whitelists) page does not appear on the sitemap, making it difficult to find the page.

Would've made a PR but didn't know what category it would fit under the best.

faucetlol:
Sitemap
* The upload whitelist link has been moved outside the `if CurrentUser.is_admin?` block so regular users can see the link too.

Upload whitelist
* A "Requests" subnav link, which links to the upload whitelist request forum.
* A "Help" link is also added, which links to the relevant help page.

This closes #420

### DIFF
--- a/app/views/static/site_map.html.erb
+++ b/app/views/static/site_map.html.erb
@@ -120,10 +120,10 @@
           <li><%= link_to("Alias & Implication Import", new_admin_alias_and_implication_import_path) %></li>
           <li><%= link_to("Danger Zone", admin_danger_zone_index_path) %></li>
         <% end %>
-          <li><%= link_to("Upload Whitelist", upload_whitelists_path) %></li>
         <% if CurrentUser.is_moderator? %>
           <li><%= link_to("IP Bans", ip_bans_path) %></li>
         <% end %>
+        <li><%= link_to("Upload Whitelist", upload_whitelists_path) %></li>
         <li><%= link_to("Mod Actions", mod_actions_path) %></li>
         <li><%= link_to("Bulk Update Requests", bulk_update_requests_path) %></li>
         <li><%= link_to("Takedowns", takedowns_path) %></li>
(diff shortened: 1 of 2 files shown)


app/views/static/site_map.html.erb
```diff
@@ -116,11 +116,11 @@
         <% if CurrentUser.is_admin? %>
           <li><%= link_to("Admin Dashboard", admin_dashboard_path) %></li>
           <li><%= link_to("Email Blacklist", email_blacklists_path) %></li>
-          <li><%= link_to("Upload Whitelist", upload_whitelists_path) %></li>
           <li><%= link_to("Post Report Reasons", post_report_reasons_path) %></li>
           <li><%= link_to("Alias & Implication Import", new_admin_alias_and_implication_import_path) %></li>
           <li><%= link_to("Danger Zone", admin_danger_zone_index_path) %></li>
         <% end %>
+          <li><%= link_to("Upload Whitelist", upload_whitelists_path) %></li>
```
Earlopain: Not properly indented. Also, please move it down to the other member visible links, just so they are more grouped together in the code.

app/views/upload_whitelists/_secondary_links.html.erb
```diff
@@ -3,6 +3,9 @@
     <%= subnav_link_to "Listing", upload_whitelists_path %>
     <% if CurrentUser.is_admin? %>
       <%= subnav_link_to "New", new_upload_whitelist_path %>
+    <% else %>
```
Earlopain: I would have just added this link for anyone regardless of user level.

Earlopain (changes requested):
Small nitpick. I won't actually be able to merge this until sometime next week, since I'm on vacation at the moment.

app/views/upload_whitelists/_secondary_links.html.erb
faucetlol: My initial thought was that having both "New" and "Request" on the menu at the same time would be a little confusing, but it makes complete sense that an admin would also want to check that forum to check if any new requests have been made. I removed the else and renamed it to "Requests" which seems a little more fitting.

app/views/static/site_map.html.erb
faucetlol: Sorry, done!

Earlopain (approved):

PR #422, opened 2022-07-10
[Cleanup] Remove pool is_deleted

Earlopain:
Pools are hard-deleted now. There are also a few pools with this flag
which can't be properly deleted at the moment


app/decorators/mod_action_decorator.rb
```diff
@@ -19,8 +19,6 @@ def format_description
       ### Pools ###
     when "pool_delete"
       "Deleted pool ##{vals['pool_id']} (named #{vals['pool_name']}) by #{user}"
-    when "pool_undelete"
```
kiranoot, line 22: Better to leave existing mod actions as there are often existing entries for them. Without this decorator they display as errors.

Earlopain: Yeah, but there is only 1 entry for this action so I thought it wasn't  worth it. I can add it back if you like though

PR #424, opened 2022-07-13
[Pools Gallery] Default to any category when none is selected

faucetlol:
Mentioned in [topic #25734 - post #303846](https://e621.net/forum_topics/25734?page=17#forum_post_303846) and [topic #34432](https://e621.net/forum_topics/34432) that by default the pools gallery only shows series when the search isn't used. 

This is contradictory to how it works when using the search - leaving the category as blank while searching will return both series and collections but the index page without any searching only returns series. This doesn't match up with the behaviour of the regular (non-gallery) listing of pools which shows both collections and series at the same time.

Without this, as far as I can tell, there's no way of viewing both series and collections at the same time without filtering any out, apart from manually altering the URL to `category=any`.

### DIFF
--- a/app/controllers/pools_controller.rb
+++ b/app/controllers/pools_controller.rb
@@ -28,7 +28,7 @@ class PoolsController < ApplicationController
 
   def gallery
     params[:limit] ||= CurrentUser.user.per_page
-    search = search_params.presence || ActionController::Parameters.new(category: "series")
+    search = search_params.presence || ActionController::Parameters.new(category: "any")
 
     @pools = Pool.search(search).paginate(params[:page], :limit => params[:limit], :search_count => params[:search])
     @post_set = PostSets::PoolGallery.new(@pools)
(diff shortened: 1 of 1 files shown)


Earlopain:
Shouldn't you just be able to remove this line for it to work as expected? I hope the search function doesn't only returns all results unless you explicitly tell it to do so.

faucetlol:
> Shouldn't you just be able to remove this line for it to work as expected? I hope the search function doesn't only returns all results unless you explicitly tell it to do so.

I did try that, after removing the line it returns the following error
```undefined local variable or method `search' for #<PoolsController:0x0000000002f9e0> @pools = Pool.search(search).paginate(params[:page], :limit => params[:limit], :search_count => params[:search]) ^^^^^^```

Earlopain:
The new, modified variable was named `search`, derived from `search_params`, try that. The base method is defined in the `ApplicationController` and allows all parameters, and a few controllers override this method in order to restrict certain search parameters like ip depending on the users level.

faucetlol:
Yeah that works just fine, would've helped if I actually read the error properly...

It looks like that line was added to intentionally exclude collections with 29fdfc42a4f90f32fc1f3f005b4d4ff90c6202b1 where I'd previously just assumed some sort of oversight, but it does make a little more sense on Danbooru because their "collections" include what e621 uses post sets for.

PR #425, opened 2022-07-13
Simple form improvements

Earlopain:
* Remove duplicate new/edit form code
* Use a custom form builder everywhere
* Use the custom form builder to add the autocomplete html property  through a shortcut
* Add a method to the custom form builder to create this:
image

This is good to go, but I want this as part of the next deploy since master has enough changes already.


kiranoot (approved):

Earlopain:
Did some more things. Moved `dtext_field` to be an simple form input and simplified/improved the newly added `button_select`.

Also did this for post edits. Why not use it when I already build it.
image


There were a few classes/ids being explicitly added to the dtext_field input. [I removed them](https://github.com/zwagoth/e621ng/pull/425/commits/bd563562b07b1e242227ebed526f71762fd4ae90) and changed the code to account for the new selectors. There was only one place I could find relying on them, that being blips. Regardless of that, any external scripts can still use the ids/classes added just because it's now using `simple_form`.

@zwagoth some of these selectors were added by yourself, any reason for that? I didn't find anything using them. If not I'll merge this.

Issue #373, opened 2022-02-04
Cleanup


Earlopain:
Tracking issue for me which things from #357 are in master right now. I will add a few commits each deploy. Commits which only change tests can probably merged whenever I feel like it. One commit which requires migration per deploy.

- [x] Fix tests. Status: 1254 runs, 2341 assertions, 0 failures, 0 errors, 0 skips

Removal only:
- [x] [Remove danbooru cloudflare purge cache functionality](https://github.com/zwagoth/e621ng/pull/357/commits/dcc2a8f1ecefe3d69a0979cb6f6a9819d5174cfb)
- [x] [Remove dead source normalization code](https://github.com/zwagoth/e621ng/pull/357/commits/48913549a4c28976f1f53cd5bbaba3a957077cfa)
- [x] [Remove sources controller](https://github.com/zwagoth/e621ng/pull/357/commits/3bb30538551db91b6f2a68b376716bc660503f6e)
- [x] [Remove strategies](https://github.com/zwagoth/e621ng/pull/357/commits/65fed7a800ca5e3adc9f22255cc136ccc290213a), break this up. Also [Rename strategy to alternate](https://github.com/zwagoth/e621ng/pull/357/commits/a057a3ea5e51c29773dd612c7203a045e4d258e5) EDIT: Mostly done in ac5e8673c,  2d94167fa, 3aa731c40, 92a0a3658, 2a4f97610, a0b2f4639
- [x] [Remove dead related tags code](https://github.com/zwagoth/e621ng/pull/357/commits/f3dc55ea0480b86dfc79126da743e2b8d8e3de0d)
- [x] [Remove more danbooru specific code](https://github.com/zwagoth/e621ng/pull/357/commits/0add842a9244c4b40ade7d9aa366f62b028d3613)
- [x] [Remove legacy api code](https://github.com/zwagoth/e621ng/pull/357/commits/71c4e9f55a6127b5d6566a8160a9fc68a5f96268)
- [x] [Remove recommender service](https://github.com/zwagoth/e621ng/pull/357/commits/3e0755f29497f1db2ebd48c1d00ef7210b5df271)
- [x] [Remove reportbooru](https://github.com/zwagoth/e621ng/pull/357/commits/5f109228b757daee5cb4760c7f6e99c0de6847ec)
- [x] [Remove donmai specifics](https://github.com/zwagoth/e621ng/pull/357/commits/2a66b6fa4f876799181187e01805f8bc6c96501a)
[...]

PR #426, opened 2022-08-06
Source alternate handlers & cleanup: FA, DA, IB, Twitter, Weasyl

Deer-Spangle:
As discussed in this thread here:
https://e621.net/forum_topics/34730

I've improved the FurAffinity source handler, and added source handlers for Deviantart, Twitter, Inkbunny, and Weasyl.
I've added to the `Source::Alternate` handlers, and added new ones. I've add unit tests for them, and added a fix script to update a large amount of broken sources.

Full list of the checks added:
- Base: Convert known secure domains to https [also in fix script]
- Deviantart: Add gallery link via submission link
- Deviantart: Convert old format links (artist.deviantart.com) to new format (deviantart.com/artist) [also in fix script]
- Furaffinity: Convert old/broken cdn links (d.facdn.net, d2.facdn.net) to new domain (d.furaffinity.net) [also in fix script]
- Furaffinity: Convert `/full/` links to `/view/` links [also in fix script]
- Furaffinity: Remove `?upload-successful` queries from URLs [also in fix script]
- Furaffinity: Remove comment anchors `#cid:` from URLs [also in fix script]
- Inkbunny: Remove page anchors `#pictop` from URLs [also in fix script]
- Twitter: Convert mobile links (mobile.twitter.com) to main domain (twitter.com) [also in fix script]
- Twitter: Convert [FixTweet](https://github.com/dangeredwolf/FixTweet) links (fxtwitter.com, ayytwitter.com, vxtwitter.com, etc) to main domain (twitter.com) [also in fix script]
- Twitter: Remove tracking data from links [also in fix script]
- Twitter: Convert old direct links (`file.jpg:orig`) to new format (`file?format=jpg&name=orig`) [also in fix script]
- Weasyl: Add gallery links from submission links

And other changes:
- Moved `force_https` check after parsing URL, to allow base handler to force https on other domains without a specific handler
- Moved source length truncation to post.rb, so all source links get truncated, not just unhandled ones
- Updated direct image link parser in FurAffinity source handler
- Stopped FurAffinity source handler from duplicating direct link
- Fixed adding direct_url to list of sources
- Moved null source strategy test into `tests/unit/sources/strategies` so that alternates tests can go in a sibling directory
- Converting some post tests to use https scheme. Fixing one of them

I've tested this with the included unit tests, and deployed a local instance with these changes and ensured new sources get modified as planned. Also tried spinning up a local instance on the `master` branch and running the fix script, which behaved as designed.
Running the full suite of tests, there are a bunch of failures, but all look unrelated to this. 
`1319 runs, 2322 assertions, 32 failures, 10 errors, 0 skips` 
I checked a couple that looked like they could be relevant, but they were not. Fixed one of them anyway.


app/logical/sources/alternates/furaffinity.rb
```diff
@@ -13,11 +13,36 @@ def domains
...
           @gallery_url = "https://www.furaffinity.net/user/#{$1}/"
-          @direct_url = @url
         end
       end
+
+      def original_url
+        # Handle old CDN or old broken CDN
+        if ["d.facdn.net", "d2.facdn.net"].include?(@parsed_url.host)
+          @parsed_url.host = "d.furaffinity.net"
+          @url = @parsed_url.to_s
```
Earlopain: @url = @parsed_url.to_s seems to happen very often. It only needs to be done once, at the end of the method. @url = @parsed_url.to_s automatically returns the result of the asignment.

Even better would be if that happends somewhere else automatically, mabye in the base class or something. But theres no need to do that now if you don't feel like it.

app/logical/sources/alternates/twitter.rb
```diff
@@ -0,0 +1,58 @@
...
+          @url = @parsed_url.to_s
+        end
+        # Remove tracking data from links
+        if @parsed_url.domain == "twitter.com" and not @parsed_url.query.nil?
+          @parsed_url.query = nil
+          @url = @parsed_url.to_s
+        end
+        # Update old direct image URLs
+        if @parsed_url.host == "pbs.twimg.com"
+          query_values = @parsed_url.query_values
```
Earlopain: query_values = @parsed_url.query_values || {}

script/fixes/108_source_cleaning.rb
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env ruby
```
Earlopain, line 1: This seems to reimplement what you already did in the source alternates. Why not simply call `strip_source` on the post instead? Later on you can save with `post.save(validate: false)`(I think) to skip creating a post version.

This would also allow us to rerun this later on, when potential new source cleanup happens. Also makes it easier to review, since I don't have to check everything twice and removes some risk, since this code doesn't run through the tests you wrote.

test/unit/sources/alternates/deviantart_test.rb
```diff
@@ -0,0 +1,56 @@
+require 'test_helper'
```
Earlopain, line 1: These test cases would benefit from a shared assert function. I have something in mind like danbooru does: https://github.com/danbooru/danbooru/blob/master/test/unit/sources/deviant_art_test.rb
https://github.com/danbooru/danbooru/blob/9cab67c0ac72a8c52289302c519715ceec2372d9/test/test_helpers/source_test_helper.rb#L8

This obviously does more than what is needed here, but should get my point across.

test/unit/sources/alternates/twitter_test.rb
```diff
@@ -0,0 +1,143 @@
...
+    context "An old twitter direct image link" do
+      setup do
+        @site = Sources::Alternates.find("https://pbs.twimg.com/media/E8v96meVgAkTKDE.jpg:orig")
+      end
+
+      should "be handled by the twitter strategy" do
+        assert(@site.is_a?(Sources::Alternates::Twitter))
+      end
+
+      should "convert to use query parameters" do
```
Earlopain: Why would you check against two possible matches? Only one of these should be the correct one. Same below.

app/logical/sources/alternates/deviantart.rb
```diff
@@ -0,0 +1,40 @@
...
+        true
+      end
+
+      def domains
+        ["deviantart.com"]
+      end
+
+      def parse
+        # Add gallery link from old-style subdomain links
+        unless BASE_HOSTS.include?(@parsed_url.host) or @parsed_url.host.nil?
```
Earlopain: Please don't use the written form for logic operators. Same thing for the other places you used them.

Also, personally I find `unless` harder to read than a simple if, so I try to invert the condition if easily possible. Like `if BASE_HOSTS.exclude?(@parsed_url.host) && @parsed_url.host.present?`. Most ruby/rails methods have an inverse of them to make this possible.

Earlopain (changes requested):
Just some things I noticed from a quick glance. Haven't checked it out yet, just went through the changes on github itself.

I myself might take a while to look it over properly, sorry for that in advance. I'm also going to somehow have to run this against the current post data, to verify that there aren't any funny edge-cases.

Regarding the failing tests: That is the number of tests that are failing for me on master right now, plus one more. I've been meaning to fix them up, but haven't gotten to them yet. I keep track of my progress in #373, where you can check the failing count as it was for me at the moment. That one more you've got is probably timing dependant.

app/logical/sources/alternates/deviantart.rb
Deer-Spangle: Ah, I'm used to python, missed that ruby has both syntax for boolean operators. And didn't realise the opposites exist, I've never used Rails before.

app/logical/sources/alternates/twitter.rb
Deer-Spangle: Sounds good

script/fixes/108_source_cleaning.rb
Deer-Spangle: That makes sense, I wasn't sure whether I could just do that. Awesome! That'll add missing galleries too

test/unit/sources/alternates/deviantart_test.rb
Deer-Spangle: Hmmm, interesting! Yeah, that could work. I'll try that out. It does lose the descriptive test names though.
Should I put the helper alongside these, or over in `tests/helpers`?

test/unit/sources/alternates/twitter_test.rb
Deer-Spangle: Hmmm, I wasn't sure whether the query order would be guaranteed? The Addressable::URI docs weren't clear, and I wasn't sure Ruby's standards about that.
Trying it both ways around, and throwing some random query values in, it seems like it's encoded alphabetically? At least in the times I tried. I'll just switch stuff to that then

app/logical/sources/alternates/furaffinity.rb
Deer-Spangle: Hmm, yeah, it could be done in a wrapper method or something, but that does seem like it might be a tricky reworking of things? Possibly, baving a base method which returned `@parsed_url.to_s` instead of `@url` might prevent any other source handler doing string based url changes.
I can imagine how it could be done in different languages, but I'm not confident enough in Ruby to be sure the best way here. I suspect it would be more lines than just having it at the end of this method though, and potentially more confusion.

I'll just move it to the end in the handlers which do this often

Deer-Spangle:
Made those changes, tests still passing ^_^
There's still a couple `unless` in there, for the optional named arguments in `alternate_should_work()`  because `false.present?` is falsy, so not a true inversion of `.nil?`.

And yeah, not sure whether to move that source_test_helper.rb to the tests/helpers directory, or keep it where it is used. Not sure it'll be useful elsewhere

test/unit/sources/alternates/twitter_test.rb
Earlopain: Looks like it's sorted alphabeticaly when passing a hash, and respects the ordering when it is an array, like `[["b", "y"], [""a", "z"]]`, `?b=y&a=z`

app/logical/sources/alternates/furaffinity.rb
Earlopain: Yeah, it's fine like this

test/unit/sources/alternates/deviantart_test.rb
Earlopain: You could pass the description to the method, like 'alternate_should("remove tracking data", ...)`.

The more correct place for this would be in the helper directory, and then include the module into the correct test type in the global `test_helper.rb`. I'm not that big on ruby/rails tests myself, so I can mostly just parrot what I saw in this repo.

Earlopain:
Thank you, it looks much better already. I'm going to take a more in-depth look some other time, and try to run it through the database export so see if anything unexpected happens.

Earlopain:
As per #427: What is the actual usage of adding a gallery link from the submission url? The gallery link is already easily visible, and it just duplicates the information. And like you said, the username doesn't even have to be correct. I knew that it works like that at twitter, but I didn't know that that's also the case for deviantart/wesyl.

Deer-Spangle:
I had much the same opinion, yeah. But I've had numerous folks say it was helpful, and I had never witnessed anyone doing that to Weasyl and DA links (and had not tested it until I mentioned in the other ticket!)
But yeah, people say gallery links are useful when galleries are nuked, and [the upload form on the site ](https://e621.net/uploads/new)also emphasizes this:
> You should include: A link to the artists page where this was obtained, and a link to the submission page where this image was obtained. No available source should ONLY be used if the content has never been posted online anywhere else.

Earlopain:
I brought this up in the staff chat, automatically adding the gallery link when you can infer it from the submission url is not something that we want to do

Deer-Spangle:
Okay! I only implemented it in the weasyl handler, because it is already present in the FA handler on master.
Shall I remove it from all of them?

DonovanDMC:
For the whole "alternate twitter" bit, I've seen another one floating around: nitter

Main url I've seen is `nitter.moomoo.me`, and it's mostly from [one user](https://e621.net/post_versions?lr=457918&search%5Bupdater_id%5D=457918).

They've got a [list of instances](https://github.com/zedeus/nitter/wiki/Instances) if you want to go that far

Earlopain:
Just the one you added is fine

Deer-Spangle:
Alright, removed gallery parsing from deviantart, weasyl, and the note from the twitter handler. Removed the weasyl handler entirely, as gallery parsing was all it did. (Added it to base handler's list of known https domains)

> For the whole "alternate twitter" bit, I've seen another one floating around: nitter

Oh yeah! Nitter was [in my original cleanup script ](https://github.com/Deer-Spangle/e621_source_cleanup/blob/master/e621_source_cleanup/checks/twitter.py#L15)too, can't believe I forgot to port it over.
Though, adding the full list of nitter domains feels like overkill... And there is an issue now, that the handler-finder only checks domains, not subdomains, so for that one, I would need to add a check for "moomoo.me", which might be imprecise.
I'll do some extra full-domain checks, maybe.

Ranking of nitter domains in use:
> 00 = {tuple: 2} ('nitter.net', 401)
> 01 = {tuple: 2} ('nitter.moomoo.me', 55)
> 02 = {tuple: 2} ('nitter.kavin.rocks', 29)
> 03 = {tuple: 2} ('nitter.it', 14)
> 04 = {tuple: 2} ('nitter.domain.glass', 7)
> 05 = {tuple: 2} ('nitter.ca', 5)
> 06 = {tuple: 2} ('nitter.fdn.fr', 4)
> 08 = {tuple: 2} ('nitter.ir', 1)
> 09 = {tuple: 2} ('nitter.cc', 1)
> 10 = {tuple: 2} ('nitter.snopyta.org', 1)
> 11 = {tuple: 2} ('nitter.tedomum.net', 1)
> 12 = {tuple: 2} ('nitter.eu', 1)
> 13 = {tuple: 2} ('twnitter.com', 1)


I'll add the ones with more than one use, and manually fix the others now

Deer-Spangle:
Ah, darn, I need to handle nitter media links ( e.g. https://nitter.net/pic/media%2FCTNngvZW4AAHvGM.jpg%3Fname%3Dsmall ) a bit better than that!
Will fix this evening, as my train just arrived at work.

It just needs to strip the "/pic" from the start of the path, switch host to pbs.twimg.com, and urldecode the rest.
How is best to do a URI decode on that last bit? Google suggests `URI.decode_www_form_component(encoded_uri)`, but is that the preferred way here, or is there another magic Rails solution perhaps? I can't find any other examples in the repo

Earlopain (changes requested):

Earlopain:
Please don't forget to update your tests, like when you added `NITTER_HOSTS`. You also left a reference to the weasyl class in `alternates.rb`, the weasyl tests are also still there.

Deer-Spangle:
I can't seem to get things working with moving the test helper. For some reason, imports are failing. Will investigate further when I can

Earlopain:
Use `extend` instead of `include`, that should work. I looked at how danbooru does it.

app/logical/sources/alternates/twitter.rb
```diff
@@ -0,0 +1,67 @@
...
+  module Alternates
+    class Twitter < Base
+      TWITFIX_DOMAINS = %w[fxtwitter.com ayytwitter.com twitter64.com vxtwitter.com pxtwitter.com twittpr.com nitter.net]
+      NITTER_HOSTS = %w["nitter.net nitter.moomoo.me nitter.kavin.rocks nitter.it nitter.domain.glass nitter.ca nitter.fdn.fr"]
+
+      def force_https?
+        true
+      end
+
+      def _nitter_domains
```
Earlopain: I thought I doubl posted the comment and deleted, but apparently I didn't.

There's no need for the underscore to be here.

Deer-Spangle: I figured it would be best as a private method, and thought that's how those were denoted. I'll rename

Earlopain: Ruby does have private methods. https://github.com/zwagoth/e621ng/blob/d86112d1d3f119d23b2a4f66294c1727b9b1df8d/app/inputs/search_form_builder.rb

Everything after the private keyword is, well, private. Something to keep in mind is that this only properly works for instance methods. It does not work like this for class methods.

test/unit/sources/alternates/source_test_helper.rb
```diff
@@ -0,0 +1,35 @@
+
```
Earlopain: Accidental newline.

I've put a rubocop config in the repo some time ago and have been chipping away at the violations as I encounter them around code I modify. The plan is to have it run as part of CI sometime, but that is still miles away.

New code should _mostly_ adhere to the style guide. I haven't updated the file in a while, so if there's a suggestion you find utterly stupid don't do it. I mostly added it to find stuff like this, or code smells like `var&.method.abc` for example.

test/unit/sources/alternates/null_test.rb
```diff
@@ -6,18 +6,24 @@ class NullTest < ActiveSupport::TestCase
       alternate_should_work(
         "http://oremuhax.x0.com/yoro1603.jpg",
         Sources::Alternates::Null,
-        "http://oremuhax.x0.com/yoro1603.jpg",
-        force_https: false
+        "http://oremuhax.x0.com/yoro1603.jpg"
       )
+
+      should "not convert to https" do
```
Earlopain: What exactly are you testing here? This is already covered by the alternate check above, since the link will or will not be https. Same below.

Deer-Spangle: True, yeah. I think it was different before? I forget. I can just rip these out then, that's cleaner

app/logical/sources/alternates/twitter.rb
Deer-Spangle: Oh yeah, of course. I forgot that. It's in the pixiv helper too, my bad, was thinking in python this morning

Deer-Spangle:
> Use `extend` instead of `include`, that should work. I looked at how danbooru does it.

Hmmm, was having some odd behaviour when doing that too (4 arguments instead of 3?)
I was trying to mirror what was going on with the test/helpers/upload_test_helper.rb, putting it alongside, adding it to test/test_helper.rb as an include, or an extend, right next to it.

My laptop just ran out of power though, so I'll need to give it another few attempts once I get home (train delay today gave me a bit of extra time to explore at least!) ^_^

Earlopain:
It seems to work fine for me with `extend`. Maybe there were some issues when you removed the gallery url parameter?

You will need to remove the setup block when moving it over there, and the instance variable too. Otherwise the test class will have multiple setup blocks, and every class will use the `@site` of the last defined test, resulting in basically all tests being broken. At least that was the behaviour I was observing, might be something different.

Also, don't feel like you need to hurry. Kira will do a deploy this week, and I'm planning for this to be part of the one after that, or even one more after that. Time between deploys is typically about 2-4 weeks, depending on how active (or not) the repo is. I try to not put too many changes in at once.

Deer-Spangle:
Ah, I saw the "checking against the last site" thing too! Okay.
I guess I should just do the URL parsing in each check then, fair enough. There's only 2 checks left in there anyway!

Yeah, I figured it might take a while to land, I'm not worried about that, just trying to keep my momentum before I dig into other stuff and forget where this was ^_^
(Probably the next thing for me will be throwing some script together to go hunt those gallery-only links and fixing them up through the API?)

Earlopain:
`alternate_should_work` is just a normal function being evaluated top-to-bottom, which means that you can put plain local-scoped variables in there. Just move the instance variable out of the current setup block and loose the `@`, that should work. No need to parse the url twice.



If you mean a seperate project to fix it up, sure. I don't think that that would be something that should be part of the main repo. One thing that comes to mind is doing md5 checks. That would yield a 0% error rate, while missing higher/lower quality submissions though. Someone did this already, I remember seeing post edit reasons mentioning exact fa md5 matches dating back a few years now I believe.

Deer-Spangle:
Ohhh, of course! I get a bit muddled thanks to Ruby's lack of parentheses around arguments, and the blocks. It has a magical way of making functions look like keywords to me.
That works! Just need to tweak some things with the nitter image parsing, now.

And yes, definitely meant a separate project! Definitely would not be something for the main repo. Would need to be querying some other stuff. I've got a database of all the perceptual image hashes across most the big furry sites now, along with a lot of other metadata, so should be able to whip up a python script using that.

Deer-Spangle:
Awesome, there we go

```
spangle@kobold:~/e621ng$ docker-compose run tests test/unit/sources/
[+] Running 4/0
 ⠿ Container e621ng-redis-1      Running                                                                                                                                                           0.0s
 ⠿ Container e621ng-elastic-1    Running                                                                                                                                                           0.0s
 ⠿ Container e621ng-memcached-1  Running                                                                                                                                                           0.0s
 ⠿ Container e621ng-postgres-1   Running                                                                                                                                                           0.0s
Waiting for elastic to come up
== Seeding database with sample content ==
Run options: --seed 47641

# Running:

..........................................................

Finished in 0.413580s, 140.2390 runs/s, 145.0748 assertions/s.
58 runs, 60 assertions, 0 failures, 0 errors, 0 skips
```

DonovanDMC:
A removal of the `/photo/X` part from Twitter links could also be done
It does nothing for the actual link, Twitter opens the tweet all the same with or without it

In fact, it's removed when you open the url
So it's useless outside of twitter

Deer-Spangle:
Hmmm, I don't know about that actually. It's a designator for the specific image of the tweet, and is used when doing previews on various platforms, I believe? (Though, maybe that's just fxtwitter, actually)
I think it makes some sense to keep it there, as a pointer to the precise image the post is sourced from, but on the other hand, it allows duplicates where something could link both the tweet, and the individual photo. And as you say, going to the link just strips it out. (Which does seem odd.)

I could be convinced either way on that, I think

DonovanDMC:
Doesn't change the display on Discord, Telegram, Guilded

(in fact, guilded shows a reply as the image)
Screenshot from 2022-08-23 20-11-27
image

The only place it seems to make a difference is on slack, and that's iffy at best

~~[the tweet, if you're curious](https://twitter.com/Gaokunx3/status/1559632481623711747)~~

Deer-Spangle:
Fair enough! I originally thought those links actually went to the pic in the tweet, was surprised to discover they just redirect to the tweet, and they don't seem to do anything on major platforms either.
Added a check to remove those too, and a test

Earlopain:
Did you commit the right thing? Only the test is present, and something else unrelated in an ancient fixer script seems to have slipped in

Deer-Spangle:
... I did not commit the right thing! Sorry, I was rushing toward the end of my lunch break there, and ticked the wrong files in RubyMine. I'll get that corrected in the evening!
(The todo was just a note from real early, so I knew an example fix script to check when writing my one. Should have cleared that on my local by now maybe)

Earlopain:
Could you also do some rebasing? The commit history is pretty messy right now. If you don't want to bother, that's fine. I can just squash merge instead.

Deer-Spangle:
Sure! I've reorganised a local branch to have a cleaner history, but I see most commits have a prefix, what are the recommended prefixes for this? `[SourceAlternate]` and `[Tests]` or something like that?

Earlopain:
Hm, I would just go with `[Posts]` in this case I think. It's not an exact science, but I try to relate my commits to a rails model. Since this improves a posts sources I would go with that. 

Though I also sometimes just use `[Tests]` when just modifying tests for example. Or `[Cleanup]` when I think the model doesn't matter much.

Just do what you think is best, it doesn't really matter much in the end. Though I would go with `[Alternates]`. It's shorter and pluralized.

Deer-Spangle:
That sounds good! Alright, lets see how that force push goes..

Deer-Spangle:
How's that? ^_^

Earlopain:
Much better, thank you

Deer-Spangle:
By the way, did this require further changes? Or just waiting for a quiet moment?

(I mean, it sounds like it may need a little rebase now)

Earlopain:
Changes aren't required, no. I only need to get of my ass and properly look at this. Really sorry about the awful turnaround time on this.

Don't worry about the merge conflics, the least I can do is to resolve them myself.

Deer-Spangle:
I was hoping this wouldn't be too much of a big or difficult change to implement!
Would it be easier split up into smaller parts? Or to rebase for you?

Can it be untagged as changes requested, at least?

faucetlol:
I guess this will need changes now that twitter.com is x.com - if it's even ever going to get merged.

app/logical/sources/alternates/furaffinity.rb
```diff
@@ -1,7 +1,7 @@
 module Sources
   module Alternates
     class Furaffinity < Base
-      IMAGE_TO_ARTIST = /facdn\.net\/art\/([0-9a-zA-Z_.~\-]+)/
+      IMAGE_TO_ARTIST = /d2?\.(?!facdn|furaffinity)\.net\/art\/([0-9a-zA-Z_.~\-\[\]]+)/
```
DonovanDMC, line 4: This is using a negative lookahead (`?!`) rather than the non capturing group I assume you're going for (`?:`), so this will never match

PR #433, opened 2022-08-26
[Posts] Add Yandex reverse search link

faucetlol:
Requested in [topic #35042](https://e621.net/forum_topics/35042) - Yandex's reverse search can be pretty useful.

image

I put it beneath Google because I wasn't really sure if there was actually any order to this list.


Earlopain:
This is not something we want to do at the moment.

PR #436, opened 2022-09-30
[Posts] Don't hide page nubers of pools behind title attribute

naf-0:
This has been annoying me ever since Firefox "Daylight" for Android stopped displaying the title of links and called it a "feature".

## Features

This PR adds the page number of a post in a pool next to the pool links above the post. Previously the page numbers were hidden behind the `title` attribute of the pool link.

Clicking on the page number will take you to the page in the pool gallery that contains the post.

## How it looks

on desktop:
image

on mobile:
image

I reused and renamed the `last-page` css rule for styling this. As a bonus the `:hover` color of these links will actually show now, because it previously used the wrong css variable name.

## Tests

I barely tested this and couldn't find any UI tests in this repo. I hope i didn't break any CSS elsewhere.
The link may or may not actually take you to the correct page.

### DIFF
--- a/app/javascript/src/styles/base/_colors.scss
+++ b/app/javascript/src/styles/base/_colors.scss
@@ -26,7 +26,7 @@ $darken-background-50: rgba(0,0,0,0.5);
 // Links
 $link-color: #b4c7d9;
 $link-hover-color: #2e76b4;
-$link-last-page-color: #3e9e49;
+$link-page-number-color: #3e9e49;
 $highlight-color: lighten($link-color, 45%);
 
 // Sections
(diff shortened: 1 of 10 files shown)


faucetlol:
It makes sense - Android Chrome apparently does the same too now, and it was just generally a hassle to see title text on mobile to begin with.

I instead think it should be toggleable though, perhaps on [/static/theme](https://e621.net/static/theme) so it's a per-device setting, because I browse e621 on both mobile and desktop and would prefer not to have the clutter on desktop where the mouseover text works just fine.

There's other places that could benefit from adding a toggle to show title text without having to mouseover, the score on the posts page for one example:

image image

Sindrake:
This is all well and good when the pool name is extremely short like in your examples.
When it's longer... well, the page numbers are not gonna be on the screen anymore.

PR #442, opened 2022-10-15
Dynamic Mascots

Earlopain:
Moves mascot stuff from the config into the site itself, in order for it to be more easily changed. Removes the need for a second manually blurred image. CSS is capable of doing this.

Mascot files cannot be: Larger than 0.5MB, or larger than 1000 pixels in either direction. That is what the current mascots adhere to, with a bit of a breather.

I will add seeding to dev for the mascots after this has been deployed to the live site.

image


kiranoot:
Would it be possible to add a "safe_mode_only" flag for mascots? There are a few mascots that are supposed to be visible only on e9. The poof mascot is supposed to be visible on e9 but I keep forgetting to update the config and files required.

Earlopain:
Sure, added.

Issue #446, opened 2022-11-01
e6 docker service doesn't wait until elastic is ready to accept connections


yggdrasil75:
I went ahead and tried to set this up. in the instructions is listed the the default account is admin and e621test, but I tried logging in with that and it did not work. it appears that there are no accounts set up right now at all, how can I either add an admin account from terminal or elevate an account made normally to admin?

DonovanDMC:
The normal setup makes both an admin and auto_moderator account
The latter is vital to functionality
Rerun setup and watch for errors, and make sure you aren't randomly disabling something

As for elevating an account, without one already existing you can't do it through the ui, so just register an account then connect to the database server, and edit the `level` of the account you want to be an admin to **50**

You can see a list of levels and their corresponding numbers [here](https://github.com/zwagoth/e621ng/blob/master/config/danbooru_default_config.rb#L84-L92)

yggdrasil75:
<details>
<summary>
Error Log
</summary>

```
== Preparing search indices ==
/usr/local/lib/ruby/3.1.0/socket.rb:1214:in `__connect_nonblock': Failed to open TCP connection to elastic:9200 (Connection refused - connect(2) for 172.19.0.4:9200) (Faraday::ConnectionFailed)
        from /usr/local/lib/ruby/3.1.0/socket.rb:1214:in `connect_nonblock'
        from /usr/local/lib/ruby/3.1.0/socket.rb:56:in `connect_internal'
        from /usr/local/lib/ruby/3.1.0/socket.rb:137:in `connect'
        from /usr/local/lib/ruby/3.1.0/socket.rb:642:in `block in tcp'
        from /usr/local/lib/ruby/3.1.0/socket.rb:227:in `each'
        from /usr/local/lib/ruby/3.1.0/socket.rb:227:in `foreach'
        from /usr/local/lib/ruby/3.1.0/socket.rb:632:in `tcp'
        from /usr/local/lib/ruby/3.1.0/net/http.rb:998:in `connect'
        from /usr/local/lib/ruby/3.1.0/net/http.rb:976:in `do_start'
        from /usr/local/lib/ruby/3.1.0/net/http.rb:965:in `start'
        from /usr/local/bundle/gems/faraday-net_http-1.0.1/lib/faraday/adapter/net_http.rb:138:in `request_via_get_method'
        from /usr/local/bundle/gems/faraday-net_http-1.0.1/lib/faraday/adapter/net_http.rb:129:in `request_with_wrapped_block'
[...]

PR #447, opened 2022-11-02
Update README.md

yggdrasil75:
fix order of commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ To mitigate this you can install a WSL distribution and clone the project inside
 1. Uncomment the `COMPOSE_PROFILES` variable if you wish to use solargraph. Doesn't work on Windows without WSL.
 1. Run the following commands:
     ```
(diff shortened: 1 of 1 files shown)


Earlopain:
Right, I this makes some sense. The problem you probably ran into is that there are no healthchecks for the services at the moment. While the e6 service depends on elastic, it doesn't wait until it's ready to accept connections.

I regularly run into this when restarting and navigating to the UI, where I have to reload a bunch when I'm too fast, until elastic is done setting itself up. I worked around this manually for the test running in 36df31eb7477b8c3d723ccc23e0b3c78839fa36a, but what I really should have done is add proper healthchecks for everything. Unfortunatly the compose `healthcheck` property is not usable for this, I will have to expand the wait script for the other services as well and use it in the required places.

While changing the order fixed it for you, if you have a particularly slow PC and don't wait long enough between commands, the issue is still present. I reopened #446 and will fix it properly sometime, or you can do it yourself if you want.

PR #455, opened 2022-12-18
Correctly Truncate Artist Other Names & URLs To 25

DonovanDMC:
This is a really nitpicky one, but I noticed artists allowed 26 other names & urls when testing them. I was unsure if I should subtract one from the `MAX_URLS_PER_ARTIST` constant when using it, or inline the number. I've gone with subtracting one to leave as little a footprint as possible.


Earlopain:
The better solution in my opinion would be to call `first`, like `uniq(&:url).first(MAX_URLS_PER_ARTIST)`. That way you don't have to subtract 1 when slicing. You can do the same with strings, so `other_name[0..99]` would become `other_name.first(100)`

DonovanDMC:
That does look cleaner. I've also moved the other names limit out into a constant in the `NameMethods` module to match how urls is.

PR #456, opened 2022-12-18
[Tickets] Fix paths for sets and pools

faucetlol:
Fixes the path used on sets and pools when displaying a ticket.

* The link on a set report ticket currently goes to `/set/show/id`, resulting in a 404.
* The link on a pool report ticket currently goes to `/pools//id`, which does actually work, but looks bad.

These changes instead result in the ticket linking to `/post_sets/id` and `/pools/id` respectively.

Also, added a space after the time ago on the new pool ticket to prevent this happening:
image


Earlopain:
I actually fixed this locally already but haven't pushed yet, sorry. Someone brought it up in staff chat when Donovan played around with tickets. Once Kira rolls out the pending changes from e9 to e6 I will push.

PR #458, opened 2022-12-24
[Artists] Add (view all) to recent posts header

faucetlol:
A few artists that I've uploaded for have expressed confusion how to view the rest of the posts when I've linked them their artist page. This would make it much clearer than knowing they have to click the artist name on the main header, and creates a little more visual consistency with wiki pages as an additional bonus.

Before (top) and after (bottom)
image

A wiki page, for comparison
image

### DIFF
--- a/app/views/artists/_show.html.erb
+++ b/app/views/artists/_show.html.erb
@@ -16,7 +16,7 @@
     <%= yield %>
 
     <div class="recent-posts">
-      <h1>Recent Posts (<%= link_to "view all", posts_path(:tags => @artist.name) %>)</h1>
+      <h1>Recent Posts (<%= link_to "view all", posts_path(tags: @artist.name) %>)</h1>
 
       <%= render "posts/partials/common/inline_blacklist" %>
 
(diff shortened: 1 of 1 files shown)


Earlopain:
Looks good, but please use the new ruby hash syntax for added code.

faucetlol:
Oops, will push that shortly 😅

While we're on the topic of the artists page, how would you feel about including a link to the comments somewhere? A few artists also wished it was easier to read the comments on all of the posts with their artist tag. `comments_path(group_by: :comment, search: {post_tags_match: @artist.name})` seems to work ideally, perhaps beneath the "View wiki page" link?

Earlopain:
> While we're on the topic of the artists page, how would you feel about including a link to the comments somewhere

Lets not. The tag search doesn't work properly, and I don't want to put it in more places when it doesn't work as expected. It's not that big of an issue for artist searches I believe, but if you search for more common tags like `anthro` for example you will see what I'm talking about.

faucetlol:
Not sure I can see what's going wrong here, but I'll take your word for it, thanks for the input anyways!

Earlopain:
The amount of results doesn't go past page one, or similar low pages depending on your limit setting. The actual result count should be magnitudes larger.

PR #469, opened 2023-02-05
[Notes] Improve search UI for notes and note changes

faucetlol:
Adds some additional search options to the notes and note changes pages, that already worked but weren't exposed in the UI.

**Note changes**
image
Previously had no UI-exposed search despite accepting the parameters. I intentionally excluded `is_active` because the results will likely confuse users - deleted notes still appear in the search because they were active at the time they were updated.

Also added the ability to search note changes by username, see `app/models/note_version.rb` because I'm not convinced that was the best solution. Doesn't support being comma-separated like other values on this page, but I don't see any other pages allowing to search for comma-separated usernames either.

**Notes**
image
Nothing major here, just added `post_id` and `is_active` to the search UI.

Original suggestion: [forum #37234](https://e621.net/forum_topics/37234)

### DIFF
--- a/app/controllers/note_versions_controller.rb
+++ b/app/controllers/note_versions_controller.rb
@@ -11,7 +11,7 @@ class NoteVersionsController < ApplicationController
   private
 
   def search_params
-    permitted_params = %i[updater_id post_id note_id is_active body_matches]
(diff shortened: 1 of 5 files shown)


Earlopain:
Thanks!

PR #470, opened 2023-02-06
Allow Privileged+ To Bypass Rating Locks

DonovanDMC:
As the title says, this allows privileged+ users to edit the rating of a post even if it has been locked. As I see it, the necessity to unlock then relock the rating (if needed) is an unnecessary burden that could be forgone.

As far as I tested, this should make no difference in functionality beyond allowing Privileged+ to bypass rating locks. This also changes the error to a more informative "is locked and cannot be changed.", as only users who cannot lock ratings would have the possibility of seeing it.

I also want to make a sidenote that editing a rating locked post via api methods produces no error, and the result looks as if the change has succeeded (the result has the provided rating, but nothing was actually changed).

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1590,11 +1590,8 @@ class Post < ApplicationRecord
     end
 
     def updater_can_change_rating
-      if rating_changed? && is_rating_locked?
-        # Don't forbid changes if the rating lock was just now set in the same update.
-        if !is_rating_locked_changed?
-          errors.add(:rating, "is locked and cannot be changed. Unlock the post first.")
-        end
+      if !CurrentUser.user.is_privileged? && rating_changed? && is_rating_locked?
+        errors.add(:rating, "is locked and cannot be changed.")
       end
     end
 
--- a/app/views/posts/partials/show/_edit.html.erb
+++ b/app/views/posts/partials/show/_edit.html.erb
@@ -29,7 +29,7 @@
     <%= f.input :locked_tags, label: "Locked Tags", autocomplete: "tag-edit", input_html: { value: (post.locked_tags || ""), spellcheck: false, size: "60x2", disabled: !CurrentUser.is_moderator? } %>
   </div>
 
-  <% if post.is_rating_locked? %>
+  <% if !CurrentUser.user.is_privileged? && post.is_rating_locked?  %>
     <div class="input">
       This post is rating locked.
     </div>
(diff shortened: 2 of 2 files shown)


faucetlol:
Little conflicted on this one, I've attempted to change the rating through the API and once or twice it hasn't changed _despite providing no error, which is definitely a problem_. Upon visiting the post page to unlock the post, taking a closer look at the image I realize it's actually been locked to that rating for a good reason. It _is_ annoying as hell when you need to unlock a post first, so I can see good reason for this.

https://github.com/e621ng/e621ng/blob/edbffeb25207a966374e87fb2120dc75a480dc27/app/views/posts/partials/show/_edit.html.erb#L32-L38

The current if/else probably isn't the clearest way to represent this, it should still show the fact the rating is locked even if the user has the ability to override it, to prevent people accidentally overriding the rating unaware of the fact it was actually locked. 

Technically you can see that the "Locked" box is checked, but I think something like this would be clearer:
image

kiranoot:
This is intentional to prevent mistakes where higher access levels attempt to do things without intending to, and end up vandalizing things in the process. The extra step forces people to step back and review what is going on before continuing.

DonovanDMC:
Making it more obvious it's still locked I can get. As for preventing unintentional abuse, how about instead of outright bypassing it, a checkbox is added that would force the change? It could show next to "this post is rating locked", and could show the boxes when checked (and in the api, a boolean option). This way it's still obvious (in the ui at least), that the rating is locked, and you need to intentionally bypass it.

I might need some help making that possible without making shit code

Earlopain:
Having a checkbox will be same as not having one, it's just an extra step that will happen automatically without really thinking about it. Needing to unlock first is much more explicit.

I would much rather tackle why you think this is a change that's needed. Rating locks are supposed to be when there can be no debate on it, or when the rating can easily be confused. In both cases the one who locks the rating should be confident that it's correct what they're doing. But if you're constantly stumbling over incorrect locked ratings then it should be looked into why that happens in the first place.

DonovanDMC:
The need to unlock ratings is more common than it would first seem as I cover a wide array of 100% explicit tags
The majority of the time it's an accident, someone uploads a very obviously explicit post as q/s and locks it upon upload

It's exceedingly common when I pick up a new tag that no one regularly upkeeps

I'll just do something clientside to "bypass" it ¯\\_(ツ)_/¯

PR #472, opened 2023-02-13
Short Circuit On Null updater_id In Pool Versions

### LINKED ISSUE #471: Pool Versions With Null updater_id Error
While looking through the autogenerated pools (created via `newpool:`, as you do) I stumbled across one that randomly errored, I can't really tell why it's erroring, but it's only erroring on its first version.

The pool is [#1059](https://e621.net/pools/1059)
Visiting any pool versions releted to this pool where the first version would be shown will error, as can be shown by [this](https://e621.net/pool_versions?search%5Bpool_id%5D=1059), and [this](https://e621.net/pool_versions?limit=1&page=11&search%5Bpool_id%5D=1059), the latter of which specifically goes for `limit=1&page=11` to only show the first version. Each other version works as they should.

This pool and its versions are from the time where there are nulls everywhere (2011), so I'm not really sure if this is even worth worrying about, but since it has an exception and not just nulls everywhere I figured I'd make an issue for it. I'd look into it myself but I can't really tell what's happening with it.

Also a side note, the pool versions show that pool is locked, despite seemingly anyone being able to edit it. (I haven't tested editing it myself, but the most recent edit of 7 days ago was by a normal user)

DonovanDMC:
When visiting a pool version with a null updater id, an exception will be thrown. This can be seen via [this](https://e621.net/pool_versions?limit=1&page=11&search%5Bpool_id%5D=1059) pool.

The exception can be seen below:

I've added a check for nil before we bother attempting to fetch anything to short circuit if the updater id is nil. As far as I can tell, this remedies the issue. Fixes #471.

### DIFF
--- a/app/models/pool_version.rb
+++ b/app/models/pool_version.rb
@@ -101,6 +101,7 @@ class PoolVersion < ApplicationRecord
   end
 
   def updater
+    return nil if updater_id.nil?
     User.find(updater_id)
   end
 
--- a/app/models/pool_version.rb
+++ b/app/models/pool_version.rb
@@ -101,8 +101,7 @@ class PoolVersion < ApplicationRecord
   end
 
   def updater
-    return nil if updater_id.nil?
-    User.find(updater_id)
+    User.find_by(id: updater_id)
   end
 
   def updater_name


Earlopain:
The nicer way would be just `User.find_by(id: id)` which doesn't throw. But I'm probably just going to backfill the nulls with user id 1 so that this isn't necessary. I'm not exactly happy with these data inconsistencies myself and I'm now in a position to do something about it myself. When I checked there were about 60 pool versions where this is the case.

DonovanDMC:
Fixing the many inconsistencies would be lovely, it's so annoying traversing over that data and seeing nulls left and right.

Earlopain:
I mean the inserted data wouldn't really make sense, but at least expectations about the structure wouldn't be broken.

Earlopain:
I filled the few cases where this is null with user id 1 so this is no longer needed. There were about 40 or so.

PR #474, opened 2023-02-17
[PostSets] Don't require wildcard for name search

faucetlol:
Don't require wildcards when searching post sets, this behavior matches the pool search. Also fix a typo in the search partial that prevented any autocompletion for usernames.

[topic #37381](https://e621.net/forum_topics/37381)


Earlopain:
Added a commit so that pools and sets share their implementation.
There was also an issue with spaces/underscores which you copied over from pools. There was no way to search for post sets with spaces in them since those got converted to underscores in the normalization method. Just passing the search string through as is solves that problem.

PR #475, opened 2023-02-22
Rubocop ci

Earlopain:
Let's see if this works.


Deer-Spangle:
Are you actually planning to delete all the old fix scripts? Wouldn't it be better to exclude the directory from rubocop, that way they can still serve as guides for future fix scripts? (I know I found them helpful)

Earlopain:
I could exclude them, sure. But most of these scripts are ancient and completly broken anyways. Did you see that I left the scripts explicitly written for e6 in?

Deer-Spangle:
Ah, fair enough! I hadn't been sure why some were kept while most were removed. Makes sense that the kept ones are since the e6 split

PR #482, opened 2023-03-14
[Themes/Comments] Theme stickied comments

faucetlol:
Somehow I've never encountered a sticky comment before now.

Before/after
image

Uses lighten instead of saturate because of adverse effects on greyscale colours (Bloodlust)


Earlopain:
Looks fine, thanks.

PR #484, opened 2023-03-16
[HelpPages] Add default value for related & title

DonovanDMC:
While messing around with various json routes in my never ending quest to break everything I normally can't touch, I came across this little bug. Not providing `related` or `title` when creating a help page results in null values being inserted, which then breaks displaying any help pages. Not providing `name` also results in an exception being thrown due to attempting to normalize a nil value. 

This pr:
- makes `related` & `title` non-nullable
- fixes the presence check for `name`
- adds presence checks for `related` & `title`

### DIFF
--- a/db/migrate/20230316084945_help_pages_add_default_values.rb
+++ b/db/migrate/20230316084945_help_pages_add_default_values.rb
@@ -1,0 +1,8 @@
+class HelpPagesAddDefaultValues < ActiveRecord::Migration[7.0]
+  def change
+    change_column_null :help_pages, :related, false
+    change_column_default :help_pages, :related, from: nil, to: ""
+    change_column_null :help_pages, :title, false
+    change_column_default :help_pages, :title, from: nil, to: ""
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -863,8 +863,8 @@ CREATE TABLE public.help_pages (
     updated_at timestamp without time zone NOT NULL,
     name character varying NOT NULL,
     wiki_page character varying NOT NULL,
-    related character varying,
-    title character varying
+    related character varying DEFAULT ''::character varying NOT NULL,
+    title character varying DEFAULT ''::character varying NOT NULL
 );
 
 
@@ -4627,6 +4627,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20230221153458'),
 ('20230226152600'),
 ('20230312103728'),
-('20230314170352');
+('20230314170352'),
+('20230316084945');
 
 


DonovanDMC:
Nice github action you've got there 🙃

Earlopain:
The `to_s` during normalization seems left over from before the not null change?

Also, could you follow the projects commit message format please. Mention which fields you made not null so it's easy to follow what happened without looking at the commit.

DonovanDMC:
> The `to_s` during normalization seems left over from before the not null change?

No, this will error if `name` is not provided. Due to this happening before validations, the value not being provided will not be caught beforehand. I looked at the artist model for a reference on how this was solved elsewhere.
https://github.com/e621ng/e621ng/blob/0faf1fa1a33cb55ee8ef4947fb73fb22f95c3afe/app/models/artist.rb#L260

To hammer the point home,

#### With `to_s` / Without `to_s`
&nbsp;

Also, on commit messages, if you do a squash merge instead of a rebase merge you can change the message however you see fit, and any nonsense messages included in the pr will be tossed out (put in the description of the commit, but those can still be removed) as you'll only be merging in one commit.

Earlopain:
> No, this will error if `name` is not provided. Due to this happening before validations, the value not being provided will not be caught beforehand. I looked at the artist model for a reference on how this was solved elsewhere.

Ah, ok. I wasn't sure when the method actually gets called. Just wanted to make sure.


> Also, on commit messages, if you do a squash merge instead of a rebase merge you can change the message however you see fit, and any nonsense messages included in the pr will be tossed out (put in the description of the commit, but those can still be removed).

I normally just do a merge commit and leave it at that. Doesn't really matter here when it's just a single commit, but I'd still prefer a proper message nonetheless. Will merge this later.

Earlopain:
This doesn't behave as expected, it should be possible to leave title and related empty. You should be able to add `allow_blank: true` for this.

DonovanDMC:
`allow_blank` allows nil values. I've taken the liberty of handling related/title being missing to the database by adding a default value of an empty string to each column. This reproduces what is most typically expected, the fields not being required at all. I changed it this way as I could not find an example of requiring a value, but allowing an empty string.

Earlopain:
You have a mismatch between the migration and `structure.sql`. Feel free to add the not null constraints back, that won't hurt. The first changed line  in `help_page.rb` is now just being moved around and is unnecessary.
The `to_s` should also be irrelevant now because the model gets initialized with the default values from the database.

DonovanDMC:
Done it

Earlopain:
Thanks!

PR #485, opened 2023-03-17
[Comments] Include blanks for some fields in comments search

faucetlol:
Use `include_blanks` on hidden, sticky and bumping inputs in the comment search, otherwise the input can't be blanked after being toggled.

Also fixed the rubocop offenses in that file since I was editing it anyways.


Earlopain:
Thanks. I'm a bit surprised with what rubocop wants to do to those arrays, but I will probably tweak that rule in the future anyways. The whole %w %i business seems worthless for short arrays like this.

PR #507, opened 2023-04-16
[Posts]: Hide disabled comments info better

DonovanDMC:
Currently, there's a bit of a mix between making disabled comments hard to find, and making them easy to find, as well as leaking a bit of information disabling comments should hide.

This pr:
- Hides the number of comments for non moderators (shows 0), both in the posts list, as well as in the json api
- Removes the comment_disabled flag in api responses
- Excludes the `comment_disabled` and `comment_enabled` post events from non-moderators, as this makes it very easy to find posts with disabled comments

In theory the comment count can still be sniffed out with a brute force `id:(...) comment_count:`, but I don't think there should really be any concern over that.

### DIFF
--- a/app/serializers/post_serializer.rb
+++ b/app/serializers/post_serializer.rb
@@ -99,7 +99,6 @@ class PostSerializer < ActiveModel::Serializer
         note_locked: nullable_to_truthy(object.is_note_locked),
         status_locked: nullable_to_truthy(object.is_status_locked),
         rating_locked: nullable_to_truthy(object.is_rating_locked),
-        comment_disabled: CurrentUser.is_moderator? ? nullable_to_truthy(object.is_comment_disabled) : false,
         deleted: object.is_deleted
     }
   end
@@ -138,8 +137,7 @@ class PostSerializer < ActiveModel::Serializer
   end
 
   def comment_count
-    return 0 unless CurrentUser.is_moderator? || !nullable_to_truthy(object.is_comment_disabled)
-    object.comment_count
+    object.visible_comment_count(CurrentUser)
   end
 
   attributes :id, :created_at, :updated_at, :file, :preview, :sample, :score, :tags, :locked_tags, :change_seq, :flags,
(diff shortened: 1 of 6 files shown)


app/serializers/post_serializer.rb
```diff
@@ -137,6 +137,11 @@ def duration
     object.duration ? object.duration.to_f : nil
   end
 
+  def comment_count
```
Earlopain, line 139: This would be best in the post model itself. `user_visible_comment_count` or something. You can also drop the nullable check since this is never null. Not sure why I added that. There is also ``post_presenter.rb` which needs the same treatment

Earlopain:
Just get rid of the api field entirely. Not sure what I was thinking including that

app/serializers/post_serializer.rb
DonovanDMC: Done and Done

DonovanDMC:
> Just get rid of the api field entirely. Not sure what I was thinking including that

I had debated doing that, but didn't want to do a breaking change. Though seeing as that isn't a field really anyone should be relying on, I'm realizing that would have been fine.

DonovanDMC:
The test didn't fail locally
ughhh

Earlopain:
I was more meaning something like

```
def visible_comment_count(user)
  (user.is_moderator? || !is_comment_disabled) ? comment_count : 0
end
```

Also not really relevant to this, but I'm trying to at least not introduce new dependencies on `CurrentUser` in anything but controllers. So passing the user into the method is preferable, even if the caller still uses CurrentUser in a non-desireable location. CurrentUser is only set when the controller callbacks fire which messes with a bunch of stuff in jobs, tests and the console.

DonovanDMC:
> I was more meaning something like
> 
> ```
> def visible_comment_count(user)
>   (user.is_moderator? || !is_comment_disabled) ? comment_count : 0
> end
> ```

Oh, sorry. I interpreted that differently, as you saw. I've changed it to that now.

PR #508, opened 2023-04-17
Iqdb upgrade take 2

### LINKED ISSUE #313: Reverse Image Search does not work with alpha channel
I am able to find https://e621.net/posts/2914213 (NSWF) when clicking "Visually similar on E6" but when using a direct url or downloading the file it shows no results.

Quering by id send the preview file to iqdb which has alpha already removed since its always jpg. This is also what is added to iqdb 
https://github.com/zwagoth/e621ng/blob/79b5ed848367e15a7a587b1858d3cd1f3794eae9/app/models/post.rb#L1982-L1987 
https://github.com/zwagoth/iqdbs/blob/97437691d30a86fc89988ea1903018025ba4af21/app/workers/add_image.rb#L10-L16

Apparently it matters if the image you are searching for in iqdb has alpha or not. A solution could be to simply thumbnail the images before sending them to iqdb.

Earlopain:
Fixes #313, previously attempted in #368.

This will run both the old and new version in parallel. The old version will still be the default, the new one will only be used when appending `v2=1` as a parameter. When the results and performance looks good I will switch it over and remove the old version entirely.

This significantly simplifies the current iqdb setup. No more separate sidekiq process, no extra http server. It's all one package now. The sqlite db should also be more resilient that the custom db from before which tended to occasionally vanish.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 server: bin/rails server -p 9000 -b 0.0.0.0
 # server: bundle exec unicorn -c config/unicorn/development.rb
-jobs: SIDEKIQ_QUEUES="low_prio:1;tags:2;default:3;high_prio:5" bundle exec sidekiq
+jobs: SIDEKIQ_QUEUES="low_prio:1;iqdb_new:1;tags:2;default:3;high_prio:5" bundle exec sidekiq
 cron: run-parts /etc/periodic/daily && crond -f
--- a/app/controllers/iqdb_queries_controller.rb
+++ b/app/controllers/iqdb_queries_controller.rb
@@ -40,18 +40,22 @@ class IqdbQueriesController < ApplicationController
 
   def new_version
     if params[:file]
-      @matches = IqdbProxyNew.query_file(params[:file].tempfile)
+      @matches = iqdb_proxy(:query_file, params[:file].tempfile)
     elsif params[:url].present?
       parsed_url = Addressable::URI.heuristic_parse(params[:url]) rescue nil
       raise User::PrivilegeError, "Invalid URL" unless parsed_url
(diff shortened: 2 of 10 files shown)


DonovanDMC:
iqdb v2, pog

Earlopain:
You won't have any luck using this just yet. New posts are added to the db but I still need to import all the other ones.

DonovanDMC:
I figured as much, since it just got merged 🙃
(Not to say that I didn't try it anyways)

Earlopain:
image
This is going to take a while

DonovanDMC:
Better have the fire department on standby for those poor servers

Earlopain:
If the speed stays constant I'm expecting this to take another 3 days

Earlopain:
Import finished about 8 hours go. Gonna have to think about how/when I want to continue with this

faucetlol:
Being able to toggle the score cutoff is incredibly helpful. I've seen barely any false positives even with lower cutoffs like 50, which surprised me considering the amount of images on e6.

Only problem encountered so far is sometimes not getting results when `&v2=1` is set.

Earlopain:
Which images/posts?

faucetlol:
I can replicate it on pretty much every post, sometimes it works sometimes it doesn't. It almost feels like it's dependent on which backend I get connected to or something, not really sure how it all works behind the scenes.

https://e621.net/iqdb_queries?post_id=4001753&v2=1
If it works, retry a few times and you'll probably encounter "No similar posts found." eventually. From my limited testing I'm roughly seeing a 20-50% failure rate on the results.

Earlopain:
My bad, one of the urls was still pointing at the old iqdb server. Should be good now

PR #513, opened 2023-05-06
[TIckets] Expand Accused Search

DonovanDMC:
This pr:
* Expands the "Accused" search option to include the creator of Blips, Forum Posts, Comments & DMails
* Adds an "Accused ID" search field, and properly implements the option (it was an allowed parameter, but never implemented)

My best thought for doing this was a new `accused_id` field for tickets. For simplicity I've also done this for user tickets. I've included the migration, as well as my best attempt at a fixer script for prod.

Currently, it's being added to new tickets via the `initialize_fields` method, as that was the only place I could think of shoving it.

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -2,6 +2,7 @@ class Ticket < ApplicationRecord
   belongs_to_creator
   belongs_to :claimant, class_name: "User", optional: true
   belongs_to :handler, class_name: "User", optional: true
+  belongs_to :accused, class_name: "User", optional: true
   before_validation :initialize_fields, on: :create
   after_initialize :validate_type
   after_initialize :classify
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -195,6 +195,18 @@ class Ticket < ApplicationRecord
 
     def initialize_fields
       self.status = "pending"
+      case qtype
+      when "blip"
+        self.accused_id = Blip.find(disp_id).creator_id
+      when "forum"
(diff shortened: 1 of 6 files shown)


DonovanDMC:
How is master passing, but merging the upstream into this pr which touches nothing related now suddenly fails

Earlopain:
There's a race condition with the elasticsearch indices not being available when needed I haven't looked into yet.

DonovanDMC:
> There's a race condition with the elasticsearch indices not being available when needed I haven't looked into yet.

Ah, makes sense,

Earlopain:
Small issue with the fixer script, not sure how you didn't catch that. Seems to work fine otherwise.

Also, I'm assuming you edited the migration and database manually after changing the column name. The fk name is generated automatically from the table and column name so it didn't match anymore and left me with a different `structure.sql` than what was commited. The best way to ammend a migration is to `bin/rails db:rollback STEP=1` and just run it again.

DonovanDMC:
> Small issue with the fixer script, not sure how you didn't catch that. Seems to work fine otherwise.

I'm not sure either, I swear I tested it. I might have just not had any dmails locally.

> Also, I'm assuming you edited the migration and database manually after changing the column name. The fk name is generated automatically from the table and column name so it didn't match anymore and left me with a different `structure.sql` than what was commited. The best way to ammend a migration is to `bin/rails db:rollback STEP=1` and just run it again.

You are correct there, I had assumed the foreign key name was random, and didn't want to exec back into the container again. I'll do that properly in the future.

PR #514, opened 2023-05-07
[Tickets] Display post report reason

DonovanDMC:
This prepends the Report Reason used when reporting a post to the ticket reason. I didn't feel this required its own field to be shown in, and I feel it's been unshown for far too long.

### DIFF
--- a/app/views/tickets/show.html.erb
+++ b/app/views/tickets/show.html.erb
@@ -65,7 +65,7 @@
           <td><span class='title'>Reason</span></td>
           <td class="dtext-container">
             <% if @ticket.qtype == "post" && @ticket.post_report_reason.present? %>
-                <%= @ticket.post_report_reason.description %>
+                <%= @ticket.post_report_reason.reason %>
              <% end %>
             <%= format_text(@ticket.reason) %>
           </td>
--- a/app/views/tickets/show.html.erb
+++ b/app/views/tickets/show.html.erb
@@ -63,7 +63,12 @@
 
         <tr>
           <td><span class='title'>Reason</span></td>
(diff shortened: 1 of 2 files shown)


app/views/tickets/show.html.erb
```diff
@@ -63,7 +63,12 @@
 
         <tr>
           <td><span class='title'>Reason</span></td>
-          <td class="dtext-container"><%= format_text(@ticket.reason) %></td>
+          <td class="dtext-container">
+            <% if @ticket.qtype == "post" && !@ticket.post_report_reason.nil? %>
```
Earlopain:
```suggestion
            <% if @ticket.qtype == "post" && @ticket.post_report_reason.present? %>
```

You also don't really need the type check since only post reports write to this variable, but it's whatever I guess

DonovanDMC: I figured the check would save an unneeded db call

Earlopain: Oh, you actually display the complete desctiption, I didn't even notice that. Now it also makes sense why you added the association to the ticket model.

I think it would make more sense to just display the reason name. The full desctiption is just noise in my opinion. I think just `report_reason` contains the name?

DonovanDMC: Only the reason id is stored with the ticket, so a lookup is needed. I've switched it to just the reason name, as the description is a bit noisy.

Earlopain: Regarding the db access. It wouldn't have made a difference.

report_reason is always set for post tickets so it is already sufficiently guarded. Rails does cache db lookups as well.

PR #517, opened 2023-05-12
[UserStatuses] Include favorites and replacements

DonovanDMC:
This pr:
* Expands the Count Fixer (`/maintenance/user/count_fixes/new`) to refresh favorites and replacements.
* Rewords the text that's shown when you're on the count fixes page a bit, as it's more than "upload" counts.

I've thought about adding an easier to see "Refresh Counts" link, as the only one on the site currently is nestled away in the sitemap. I considered the secondary links of the user profile, or on the user home page, but didn't want to tackle that in this pr since it isn't necessarily related.


Earlopain:
I know that posts and favorites sometimes get out of sync, but is there evidence so suggest that this happens to the other fields you are all adding, like sets and blips?

DonovanDMC:
> I know that posts and favorites sometimes get out of sync, but is there evidence so suggest that this happens to the other fields you are all adding, like sets and blips?

No, I just slapped in everything. I couldn't even find a place where blips/sets were even used.
I'd imagine `forum_post_count`, `comment_count`, `blip_count` & `set_count` are unlikely to ever get out of sync

Earlopain:
Set count is used somewhere to limit the amount of sets someone can create.

Generally, all the ``user_status_counter` counts are very reliable. It's only the other ones that sometimes give trouble for one reason or another

DonovanDMC:
I've removed these: `post_flag_count`, `forum_post_count`, `comment_count`, `blip_count`, `set_count`

Replacements were specifically asked for, and including the other "edit" stats feels like completeness
I'm not going to fight to keep them though, if you feel they're unneeded

app/models/user.rb
```diff
@@ -762,7 +762,14 @@ def refresh_counts!
           post_count: Post.for_user(id).count,
           post_deleted_count: Post.for_user(id).deleted.count,
           post_update_count: PostVersion.for_user(id).count,
-          note_count: NoteVersion.where(updater_id: id).count
+          favorite_count: Favorite.for_user(id).count,
+          wiki_edit_count: WikiPageVersion.for_user(id).count,
+          note_count: NoteVersion.for_user(id).count,
+          pool_edit_count: PoolVersion.for_user(id).count,
+          artist_edit_count: ArtistVersion.for_user(id).count,
+          own_post_replaced_count: PostReplacement.for_user(id).count,
```
Earlopain: These two are not correct. It should count posts that were uploaded by the user and then replaced by someone else. It needs to be `for_uploader_on_approve` instead.

app/models/user.rb
```diff
@@ -762,7 +762,14 @@ def refresh_counts!
           post_count: Post.for_user(id).count,
           post_deleted_count: Post.for_user(id).deleted.count,
           post_update_count: PostVersion.for_user(id).count,
-          note_count: NoteVersion.where(updater_id: id).count
+          favorite_count: Favorite.for_user(id).count,
+          wiki_edit_count: WikiPageVersion.for_user(id).count,
+          note_count: NoteVersion.for_user(id).count,
```
Earlopain, line 766: Yeah, I don't see a point in the edit ones being here

DonovanDMC: Changed to be just add favorites (as those are known to get out of sync), and replacements

DonovanDMC:
As said in https://github.com/e621ng/e621ng/pull/517#discussion_r1194956988, I've reduced it to just adding `favorite_count` (we know this can get out of sync), and the requested replacement counts

app/models/user.rb
Earlopain: Only the two `own_` were wrong, the last for rejected was correct but is now wrong. Rejected replacements will never be approved and as such will never have an original uploader on approve

DonovanDMC: I changed all 3 in bbb0b77e31745b67aba24f8535256134c4ff9f17, but fixed the rejected one in 659205d878de417bc1bfa7f3aa17577bd111b47c, so it should be correct

Earlopain:
If I remember I will refresh counts for every user. That is a bit overdue, I keep stumbling over older users with negative uploads as well.

PR #518, opened 2023-05-13
[Tickets] Profile Links & Pending Count

DonovanDMC:
This pr:
* Adds a ticket count to user's profile, visible to them and moderator+ (along with pending & accused links)
* Displays the "Accused" user within tickets, along with a count of the "active" tickets against them (pending + partial)
* Replaces the timestamps with the standard `time_ago_in_words_tagged`

Also due to editor magic ✨, I've fixed all of the `Style/HashSyntax` issues in the user presenter. (..it's almost too tempting to go around fixing every tiny error)
The important lines are 137-139.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -766,8 +766,7 @@ class User < ApplicationRecord
           post_count: Post.for_user(id).count,
           post_deleted_count: Post.for_user(id).deleted.count,
           post_update_count: PostVersion.for_user(id).count,
-          note_count: NoteVersion.for_user(id).count,
-          ticket_count: Ticket.for_user(id).count,
+          note_count: NoteVersion.where(updater_id: id).count
         )
       end
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -732,6 +732,10 @@ class User < ApplicationRecord
       user_status.post_flag_count
     end
 
+    def ticket_count
+      user_status.ticket_count
+    end
+
     def positive_feedback_count
       feedback.positive.count
     end
@@ -762,7 +766,8 @@ class User < ApplicationRecord
           post_count: Post.for_user(id).count,
           post_deleted_count: Post.for_user(id).deleted.count,
           post_update_count: PostVersion.for_user(id).count,
-          note_count: NoteVersion.where(updater_id: id).count
+          note_count: NoteVersion.for_user(id).count,
+          ticket_count: Ticket.for_user(id).count,
         )
       end
     end
--- a/app/views/tickets/index.html.erb
+++ b/app/views/tickets/index.html.erb
@@ -50,8 +50,8 @@
             <% end %>
 
             <td class="<%= ticket.status %>-ticket"><%= pretty_ticket_status(ticket) %></td>
-            <td style="cursor:help;" title="<%= ticket.updated_at.strftime("%b %d, %Y %I:%M %p") %>"><%= time_ago_in_words(ticket.updated_at) %> ago</td>
-            <td style="cursor:help;" title="<%= ticket.created_at.strftime("%b %d, %Y %I:%M %p") %>"><%= time_ago_in_words(ticket.created_at) %> ago</td>
+            <td style="cursor:help;" title="<%= ticket.updated_at.strftime("%Y-%m-%dT%H:%M%:z") %>"><%= time_ago_in_words(ticket.updated_at) %> ago</td>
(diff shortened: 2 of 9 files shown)


DonovanDMC:
Oh fuck off rubocop

app/models/user.rb
```diff
@@ -762,7 +766,8 @@ def refresh_counts!
           post_count: Post.for_user(id).count,
           post_deleted_count: Post.for_user(id).deleted.count,
           post_update_count: PostVersion.for_user(id).count,
-          note_count: NoteVersion.where(updater_id: id).count
+          note_count: NoteVersion.for_user(id).count,
+          ticket_count: Ticket.for_user(id).count,
```
Earlopain: I see not reason to fix something that isn't broken in the first place. I have not found evidence that the status counter gets out of sync

app/views/tickets/index.html.erb
```diff
@@ -50,8 +50,8 @@
             <% end %>
 
             <td class="<%= ticket.status %>-ticket"><%= pretty_ticket_status(ticket) %></td>
-            <td style="cursor:help;" title="<%= ticket.updated_at.strftime("%b %d, %Y %I:%M %p") %>"><%= time_ago_in_words(ticket.updated_at) %> ago</td>
```
Earlopain, line 53: Just use the way every other place display time, there's a helper method for that. That would be the proper way

DonovanDMC: > I would have used `time_ago_in_words_tagged` for [this], but that would slightly change the way it works currently. Right now, the title is on the entire column.

If we don't care about that difference, that's all good and well, I just didn't want to change something that might have been a specific way just to have to undo it later

app/models/user.rb
DonovanDMC: With something that will be counted so infrequently, you are probably correct
I had assumed that *everything* would be included in a count refresh, regardless of how infrequent or unlikely they are to be inaccurate

app/views/tickets/index.html.erb
Earlopain: I can't imagine any reason why it should be the way it is currently specifically. Kira was probably just in a rush to get it all working. Sets/Takedowns don't adhere to the general style of the codebase as well. That results in these slight discrepancies with the time formatting you noticed.

app/views/tickets/show.html.erb
```diff
@@ -71,10 +80,10 @@
           </td>
         </tr>
 
-        <% if(!@ticket.response.blank?) %>
+        <% unless @ticket.response.blank? %>
```
Earlopain: I'm not really a fan of `unless` if it can be avoided.

```suggestion
        if @ticket.response.present?
```

app/views/tickets/show.html.erb
```diff
@@ -71,10 +80,10 @@
...
           <tr>
             <td><span class='title'>Handled by</span></td>
-            <% if (!@ticket.handler.nil?)%>
+            <% if !@ticket.handler.nil? %>
```
Earlopain:
```suggestion
            <% if @ticket.handler.present? %>
```

app/views/tickets/index.html.erb
DonovanDMC: Standardization gets no complaints from me

app/views/tickets/show.html.erb
```diff
@@ -83,7 +83,7 @@
         <% unless @ticket.response.blank? %>
           <tr>
             <td><span class='title'>Handled by</span></td>
-            <% if !@ticket.handler.nil? %>
+            <% if @ticket.handler.present? %>
```
Earlopain, line 88: Quick note, you were previously saying you wanted to do a nil check to prevent a database query. I checked and rails is clever enough to not do a database query in this case. It doesn't matter if you do `handler_id.nil?` or go directly for `handler.present?`

DonovanDMC: Oh, nice to know

DonovanDMC: The only time I can remember saving a db call is checking the qtype in https://github.com/e621ng/e621ng/pull/514#discussion_r1189490075

Still, nice to know

Earlopain: Yeah, that's what I was talking about. Same situation here

Earlopain:
Fixer script doesn't work, otherwise good to go

Earlopain:
Please confirm that the fix you are doing is actually fixing things. I can't tell you why it doesn't work or why it didn't work int the first place but `update_all` bypasses whatever rails is doing since it goes straight to the db.

PR #519, opened 2023-05-13
[Blips] Add json response for editing old blips

### LINKED ISSUE #502: Editing a blip older than 5 minutes returns a 302 Found
As the title says, when editing a blip via `/blips/:id.json`, a 302 Found is returned is the blip is too old.

DonovanDMC:
Fixes #502

This pr makes `PATCH /blips/{id}` return a json error instead of a 302 Found redirect when a blip is too old to edit.

### DIFF
--- a/app/controllers/blips_controller.rb
+++ b/app/controllers/blips_controller.rb
@@ -116,10 +116,7 @@ class BlipsController < ApplicationController
         redirect_back(fallback_location: blips_path, flash: { notice: "You cannot edit blips more than 5 minutes old" })
       end
       format.json do
-        render json: {
-          success: false,
-          reason: "You cannot edit blips more than 5 minutes old",
-        }, status: 422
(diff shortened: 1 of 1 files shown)


app/controllers/blips_controller.rb
```diff
@@ -111,7 +111,17 @@ def blip_params(mode)
...
-    redirect_back(fallback_location: blips_path, flash: {notice: 'You cannot edit blips more than 5 minutes old'})
+    respond_to do |format|
+      format.html do
+        redirect_back(fallback_location: blips_path, flash: { notice: "You cannot edit blips more than 5 minutes old" })
+      end
+      format.json do
+        render json: {
+          success: false,
+          reason: "You cannot edit blips more than 5 minutes old",
+        }, status: 422
```
Earlopain: Ah, that's how you do multiline code suggestions
```suggestion
       render_expected_error(422, "You cannot edit blips more than 5 minutes old")
```

PR #520, opened 2023-05-13
[Artists] Respond with json for name lookup & deletion

### LINKED ISSUE #503: Getting a nonexistent artist by name redirects when using JSON
Though not necessarily the intended method of getting an artist by name in the api, `/artsts/[name].json` works. While this works, it redirects to `/artists/show_or_new` if an artist by that name doesn't exist, regardless of response type.
https://github.com/e621ng/e621ng/blob/179bb69f1da33de904cc9c0788e4db3a2fc5ac54/app/controllers/artists_controller.rb#L26-L38

DonovanDMC:
Fixes #503

This makes `GET /artists/{name}` return a proper 404 json response as it should, as well as returning a json response when deleting an artist.

### DIFF
--- a/app/controllers/artists_controller.rb
+++ b/app/controllers/artists_controller.rb
@@ -29,7 +29,14 @@ class ArtistsController < ApplicationController
     else
       @artist = Artist.find_by(name: Artist.normalize_name(params[:id]))
       unless @artist
-        redirect_to(show_or_new_artists_path(name: params[:id]))
+        respond_to do |format|
+          format.html do
+            redirect_to(show_or_new_artists_path(name: params[:id]))
+          end
+          format.json do
+            raise ActiveRecord::RecordNotFound
+          end
+        end
         return
       end
     end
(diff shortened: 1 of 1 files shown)


app/controllers/artists_controller.rb
```diff
@@ -50,11 +57,18 @@ def update
   end
 
   def destroy
-    if !@artist.deletable_by?(CurrentUser.user)
+    unless @artist.deletable_by?(CurrentUser.user)
       raise User::PrivilegeError
     end
     @artist.update_attribute(:is_active, false)
-    redirect_to(artist_path(@artist), :notice => "Artist deleted")
+    respond_to do |format|
```
Earlopain:
```suggestion
    respond_with(@artist) do |format|
      format.html do
        redirect_to(artist_path(@artist), notice: "Artist deleted")
      end
    end
```

DonovanDMC:
Thanks github

Earlopain:
> Thanks github

It's my fault really, for the less than ideal test setup. I did f722dbb4c38c8a8aed06d3948149b8f990a1cb1e which hopefully improves on that a bit for future PRs. Didn't fail immediatly so maybe it actually works.

DonovanDMC:
> > Thanks github
> 
> It's my fault really, for the less than ideal test setup. I did [f722dbb](https://github.com/e621ng/e621ng/commit/f722dbb4c38c8a8aed06d3948149b8f990a1cb1e) which hopefully improves on that a bit for future PRs. Didn't fail immediatly so maybe it actually works.

I was more referring to what happened in d1013724e57eaf982557044250a830dc9afba9e5 (before I force pushed it out) when I commited that suggestion, but it seems like that was actually your doing, since your suggestion added on and didn't remove the rest 🙃

The test fails are annoying yes, but I've more or less learned to accept them as something that will happen

PR #521, opened 2023-05-14
[ModActions] Revise report_reason_update & user_feedback_update

### LINKED ISSUE #488: [ModActions] Inconsistencies and Issues
<details>
<summary><code><strike>artist_page_rename</strike></code></summary>
When renaming an artist, the new name is shown on both sides, making it look like nothing actually happened.

image
</details>

<details>
<summary><code><strike>artist_user_unlinked</strike></code></summary>

The `artist_user_unlinked` modactions do not display a user like they should.

image
image
</details>

<details>
<summary><code>forum_topic_update</code></summary>

`forum_topic_update` is in the search dropdown, but it seems completely unused. In my mind this should be shown when the name of a forum topic is changed, since stick and lock have their own actions, but I can't see anything logged when the name is changed. As far as I can tell (aka I looked at the first commit it appeared in and called it a day), it's been unused since it was introduced in ca5b0d5f807152ad2b2c37f9fa8d7892c25f43e3.
</details>

<details>
<summary><code>report_reason_update</code></summary>

Only the reason is logged in the modactions, which hides changes to the description, making it look like nothing happened.
image
</details>

<details>
<summary><code>user_feedback_update</code></summary>

The `user_feedback_update` does not show if the type of record (positive/negative/neutral) was changed, it only shows the reason, regardless if it was changed. (as can be seen. It *does* log the new type of record, but doesn't say the previous type or call out the changes)

image
</details>

<details>
<summary><code><strike>wiki_page_undelete</strike></code></summary>

Since dc3fa7fd3459f9723472a9901b6e4d006749a30e, deleting a wiki page fully destroys the wiki page, and none of these entries seem to exist, thus making this action irrelevant.
</details>

DonovanDMC:
Fixes #488

This pr:
* Adds more detailed logging for the `report_reason_update` action
* Clarifies details for the `user_feedback_update` action (showing type changes & reason changes separately)

<details>
<summary><code>report_reason_update</code> Example</summary>

image
</details> 

<details>
<summary><code>user_feedback_update</code> Example</summary>

image

Entries without `reason_was` or `type_was` are assumed to be legacy, and use the original format.
</details>

### DIFF
--- a/app/decorators/mod_action_decorator.rb
+++ b/app/decorators/mod_action_decorator.rb
@@ -96,7 +96,18 @@ class ModActionDecorator < ApplicationDecorator
     when "user_feedback_create"
       "Created #{vals['type'].capitalize} record ##{vals['record_id']} for #{user} with reason: #{vals['reason']}"
     when "user_feedback_update"
-      "Edited #{vals['type']} record ##{vals['record_id']} for #{user} to: #{vals['reason']}"
+      if vals["reason_was"].present? || vals["type_was"].present?
+        text = "Edited record ##{vals['record_id']} for #{user}"
+        if vals["type"] != vals["type_was"]
+          text += "\nchanged type from #{vals['type_was']} to #{vals['type']}"
+        end
+        if vals["reason"] != vals["reason_was"]
+          text += "\nchanged reason from \"#{vals['reason_was']}\" to \"#{vals['reason']}\""
+        end
+        text
+      else
+        "Edited #{vals['type']} record ##{vals['record_id']} for #{user} to: #{vals['reason']}"
+      end
     when "user_feedback_delete"
       "Deleted #{vals['type']} record ##{vals['record_id']} for #{user} with reason: #{vals['reason']}"
       ### Legacy User Record ###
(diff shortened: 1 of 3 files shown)


app/decorators/mod_action_decorator.rb
```diff
@@ -325,6 +343,8 @@ def format_description
       "Post rating was #{vals['locked'] ? 'locked' : 'unlocked'} on post ##{vals['post_id']}"
     when "post_unapprove"
       "Unapproved post ##{vals['post_id']}"
+    when "post_owner_reassign"
```
Earlopain: This seems entirely unnessesary to add

DonovanDMC: I only did that due to [a few entries existing](https://e621.net/mod_actions?search[action]=post_owner_reassign)
So if anyone sees them in the future, they don't report them as a bug

Earlopain: I might as well just delete the existing ones. They contain no actual information anyways

DonovanDMC: I'm not opposed to that, in fact that's probably the better solution with them having no information

Earlopain:
I changed how feedback changes get displayed, I wasn't a fan of just throwing it in there. Trailing newlines made it look particulary egregious with the quotes and all.

image

DonovanDMC:
That looks much better than with the quotes.

PR #524, opened 2023-05-17
[ForumPosts] Check for existence of topic & category

DonovanDMC:
This pr prevents an unexpected error being thrown when attempting to create a topic with an invalid `category_id`.

With this, providing an invalid category produces the expected error:
```json
{
  "errors": {
    "category": [
      "must exist"
    ]
  }
}
```


Earlopain:
Can you add a test please

PR #527, opened 2023-06-10
[Tags] Disallow é in tag names

faucetlol:
This exception was originally made for `pokémon` tags, which have now mostly been aliased away to their `pokemon` variants, [but tags are still being created using é](https://e926.net/tags?commit=Search&search%5Bhide_empty%5D=1&search%5Bname_matches%5D=%2A%C3%A9%2A&search%5Border%5D=date) because of this exception.

There's no reason to give é special treatment anymore when no other accented characters are allowed.


Earlopain:
Thanks

PR #529, opened 2023-06-26
[Users] Add avatar search

DonovanDMC:
Adds an avatar search to `/users`.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -136,7 +136,7 @@ class UsersController < ApplicationController
   end
 
   def search_params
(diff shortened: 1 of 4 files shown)


Earlopain:
I don't think the link on every post is needed. Please remove that.
You also don't need to manually cast `to_i`, rails does type conversion automatically.

DonovanDMC:
I figured the link might be interesting for regular users to see, I just threw it in there since I couldn't think of anywhere good to put it

I also thought it maybe useful for staff when looking for alts, clicking into the post then clicking that link to see accounts with that same avatar, like our recent recurring actor

PR #530, opened 2023-06-26
[UserRevert] Fix error when user has uploaded a post

DonovanDMC:
Currently, when performing a user revert, it will throw an error if any of the changes are an upload. This makes use of the `can_undo?` method post versions have to avoid this.


Earlopain:
Done in https://github.com/e621ng/e621ng/commit/42abd4ca333edb683baa3a4918b9d544938e5113 where I went for a more in-depth fix by decoupling the permission check from the validation check.

PR #534, opened 2023-07-07
[Search] Allow ascending order for IDs

clragon:
This MR allows specifying the direction of ordering for IDs on any search endpoint that supports ordering.


Earlopain:
This is pretty cool, thanks for working through with me on this one

PR #535, opened 2023-07-11
[Tags] Use template for MetaSearch relation tables

faucetlol:
I hadn't previously realised that the MetaSearch page didn't only include active relationships, but also the pending and deleted aliases - which makes the data really misleading with no way to tell the status.

I was originally going to just add the status column, but realised I can just use this template that already exists and save a bunch of code duplication. All the additional columns added by the template are probably reasonable to include on the MetaSearch too.


Earlopain:
Nice find

PR #538, opened 2023-07-24
[Uploader] remove "I am the source"

faucetlol:
image

This has came up more than a few times, with many people suspicious it's the reason so many artists/commissioners self-upload without providing any source links. 

I'm not convinced it's not just pure laziness on their part, but it wouldn't hurt to remove the ambiguity from the uploader. Being _the source_ doesn't exclude you from needing to provide source links, unless it's not available, which is already covered by "No available source"


Earlopain:
Sounds reasonable, thanks.

PR #543, opened 2023-08-20
Search improvements

Earlopain:
Allow specifying (most) metatags multiple times and support negation


DonovanDMC:
While you're running around in this area, is there any way to make the `id:` metatag accept more than 100 comma separated ids? it's a minor annoyance really, but it would be nice if its limit were to match the global limit of 320, for fetching posts in bulk.

PR #544, opened 2023-08-24
fix invalid tag category hover color

OpportunePineapple:
I'm not sure what the proper category would be for this fix, so sorry if the commit message isn't properly formatted. Please feel free to change it as you see fit. :p


Earlopain:
Thanks!

Issue #554, opened 2023-10-05
Adding dead sources through the API


faucetlol:
It seems to be impossible to add a dead source link with `post[source_diff]` because the `-` at the start of the URL signifies that it should be removed in the diff.

It's possible to do it with the `post[old_source]` and `post[source]` method, but that's advised against.

PR #558, opened 2023-11-05
[Posts] Add behavior to add sources using '+source' metatag

### LINKED ISSUE #1334: Allow adding sources to posts via tag script
It is currently not possible to add sources to posts via tag script, as the `source` meta tag overwrites all sources.

We could add a new `+source` meta tag to allow adding sources.

mm12:
The `source:` metatag current overwrites all sources when used. This adds behavior for a `+source` metatag that appends a new source to the existing source string, and adds unit tests for it. 

fixes #1334


faucetlol:
Is there any reason to use metatags to edit sources rather than just using `post[source_diff]` in the API? I wasn't even aware that `source:` could actually be used for editing.

It does actually seem to give a bypass for issue #554 though, seeing as I'd just be able to do `+source:-https://brokenlink.com` as a tag edit.

The metatag name `+source` would, to me, also imply the existence of `-source`, which doesn't exist and probably never could because of technical reasons.

mm12:
> Is there any reason to use metatags to edit sources rather than just using `post[source_diff]` in the API? I wasn't even aware that `source:` could actually be used for editing.
> 
> It does actually seem to give a bypass for issue #554 though, seeing as I'd just be able to do `+source:-https://brokenlink.com` as a tag edit.
> 
> The metatag name `+source` would, to me, also imply the existence of `-source`, which doesn't exist and probably never could because of technical reasons.

Yes, tag scripts and similar tools can be used to add the same source to multiple posts. This is useful in instances where a site has multiple images on one post, such as Pixiv and Twitter. There is no way to add a source to multiple posts otherwise.
Originally, I called it "addsource", however, that name is inconvenient. I suppose removing sources the same way would be handy, but how exactly that would work with wildcards is not fun to think about. Other than wildcards though, you could probably just slice it since its a string


Also, it doesn't completely fix #554 since it will add a dead source, even if it already exists, since the source name is technically different due to the `-`

mm12:
> Is there any reason to use metatags to edit sources rather than just using `post[source_diff]` in the API? I wasn't even aware that `source:` could actually be used for editing.
> 
> It does actually seem to give a bypass for issue #554 though, seeing as I'd just be able to do `+source:-https://brokenlink.com` as a tag edit.
> 
> The metatag name `+source` would, to me, also imply the existence of `-source`, which doesn't exist and probably never could because of technical reasons.

I have made a simple example of being able to remove sources. https://github.com/mm12/e621ng/blob/sourceFix_removeExample/app/models/post.rb

However, it doesn't work in the context that adding sources is meant to (tag scripts) - only in the editor. Annoyingly, this is only due to the usage of the "-" character, which has some special case in the tagsript js

mm12:
... ok will fix

mm12:
I found the problem - my branches broke because I messed up (proof of concept code leaked between them causing errors). I have re-synced to master and fixed the branch, and everything is passing now.

mm12:
I just realized I should add this to `https://github.com/e621ng/e621ng/blob/master/app/logical/tag_query.rb` as well?

jmortiger:
Looking at `Files Changed`, it looks like this has been added, just not through merging this PR; is that accurate?

mm12:
> Looking at `Files Changed`, it looks like this has been added, just not through merging this PR; is that accurate?

No, I don't believe it is:  https://github.com/e621ng/e621ng/blob/851cd5c68a922a39b7b300670b0d988b4c41693e/app/models/post.rb#L824-L837

I need to rebase, though.

mm12:
Will try to migrate to `^source:([""]?)(.*)\1$` (or something along those lines) to solve both of these at once. Making the tests a bit better to make sure that it works as intended too.

Sindrake (changes requested):
Looks good, but could use more tests, just for safety's sake.
- `+source:"https://example.com"`
- `+source:none` (would it just append "none" to sources?)
- multiple `+source:` in one string

mm12:
> Looks good, but could use more tests, just for safety's sake.
> 
>     * `+source:"https://example.com"`
>     * `+source:none` (would it just append "none" to sources?)
>     * multiple `+source:` in one string

Well a partial rewrite of a small PR is probably fine

Sindrake (approved):
lgtm

PR #563, opened 2023-12-02
Revamp of replacements index UI

mm12:
By popular request, this is a revamp of the replacements index UI. It makes various properties easier to read. 
In addition, the following was changed: 
1. CodeBehind: Ordering changed from pending -> by id to pending -> by id -> original
2. Sequence number: The sequence number of the replacement is calculated and shown to the user
3. Backups are no longer created unless it is the original post's backup, to avoid duplicates
4. A DB fix to remove existing duplicates is provided

Staff actions and penalty status only appear to staff since that is a planned feature. 

Labeling may need to be changed for consistency with other e621 indexes.


mm12:
This last commit fixes an issue where destroyed Replacements caused the others for the post to have messed up sequence numbers due to how I hacked around it. This way works better.

mm12:
What it looks like
image

mm12:
i just noticed a mistake, image info is wrong. Will have to fix that... and might add accepted dates, if its in there

mm12:
What I wanted: 
- Posts that have never had a replacement should have a virtual 'original' added to the UI
What I did instead: 
- I put back the current-version resolution
----
Changes: 
- [Post] When a post is a video, use "replace video" instead of "replace image"
- [DB FIX] Added string frozen magic comment
- [UI] Fixed 'creator' and 'previous uploader'
- [UI] Added the file info of the current version
- [UI] Added the 'updated-at' field the the UI 
- [UI] Fixed sequence numbers (original always 0)
- [TODO-Replacements] Fix ordering (bottom->top: promoted, rejected, original, accepted, pending)
- [TODO-Replacements] Durations (needs testing)  -> want to consult @Sindrake about this
  - Added commented-out code to enable replacements having a duration property.  
  - Added logic for if-video show-duration

mm12:
image
image


Here is how it currently looks (without the images themselves). Yes, it is information dense. The ordering is fixed. Adding duration will be left for the future.

Sindrake:
Closed in favor of #1048.

PR #566, opened 2023-12-07
[VoteManager]: Make links work

DonovanDMC:
This simply ignores clicking on the row if the element we're clicking on is a link, making the links on `/post_votes` & `/comment_votes` work flawlessly.


Earlopain:
That's nice, thanks

PR #567, opened 2023-12-07
Add some tools that make it easy to convert chains of parent/child into pools

### LINKED ISSUE #1883: Tool to convert chains of parent/child posts into pools
Users tend to chain posts in parent/child relations rather than creating a pool. This makes several things difficult: 
1. Usage of the `inpool` metatag becomes less useful. Using this for approvals to review pools all at once does not work when things are not in pools.
2. Alternate versions vs next image - items that would normally be an alternate version get pushed around because posts can only have 1 parent, but can have many children. Pages have both an alt and a next page in the children, making navigation difficult.

A tool for either staff or priv+ users. This tool should
1. Crawl a chain of posts for related items
2. Ask user to select the last item, from which the parent chain is followed.
  - Allow users to adjust as needed
  - Potentially detect when alt versions exist, eg, an SFW and NSFW version of the same pool.
3. Create a post string and allow creating a new pool from it.

**Describe alternatives you've considered**
A spot in the upload UI (#1882) would help avoid this issue, but not negate it's need entirely.

**Additional context**
A previous attempt at implementation can be found in #567

mm12:
Approvers feature 

- MODEL:POST: Added methods for:
  - Recursively getting the children of the post (with depth)
  - Recursively getting the parents of the post (with last parent)
  - Getting all posts that are parents/children of the post (family)
  - Getting all posts that are children of family members
  - Removing parents from a list of given post IDs
- VIEWS:POST: 
  - Added a 'related' view to view a list of post IDs related to a given post
    - Added a button to create a pool from these IDs
    - TODO: I would like to make this a PostSets, but do not know how to convert an array of post IDs to one
  - Added button to to go the 'related' page from the secondary links bar
  - Added a search partial to support the related view
- CONTROLLERS:POOL:
  - Added capability for URL params to fill in post IDs for new pool creation, to support the 'pool from' button
- CONTROLLERS:POST:
  - Added 'related' method to handle searches for related posts 
- ROUTES:
  - Added the posts/related route
- TESTS: 
  - Added tests for the new `MODEL:POST` methods


mm12:
Current TODOs: 
- Move post parent removal from post model to (post controller? not sure here) since it doesn't pertain to an individual post 
- Write tests for post parent removal 
- Add PostSets presenter to UI
- Write tests for pool with params (maybe) 
- Add tests for underprivileged users 
- Add cases for "parallel" pools (chains with actual alternates) - warn user if multiple children 

It would be nice if this could be done with search metatags, but I cannot figure out the way they work.

mm12:
Another consideration for adding to this, since it has to do with pools: add a field for adding a source to all posts this applies to.

mm12:
This PR does too much. Good feature, bad implementation.

PR #569, opened 2023-12-10
Fixed replacements pending link not filtering with the pending status

S87GMIL:
Fixed the "[Pending]" link for replaced posts, so it now correctly filters the "Pending" status of replaced posts.
Before, the link simply displayed all replaced posts, and didn't set the status filter:
image

After the fix, the link correctly includes the pending status filter in order to only display pending replaced posts:
image

### DIFF
--- a/app/views/users/_statistics.html.erb
+++ b/app/views/users/_statistics.html.erb
@@ -41,7 +41,7 @@
         <span>Replaced</span>
         <span>
           <%= presenter.replaced_upload_count(self) %>
-          [<%= link_to "pending", post_replacements_path(search: { creator_name: user.name }) %>]
+          [<%= link_to "pending", post_replacements_path(search: { creator_name: user.name, status: "pending" }) %>]
         </span>
 
         <span>Rejected</span>


DonovanDMC:
It wouldn't hurt to add a view link or something to the left of pending so we can actually still get to their replacements
Maybe even an entirely new column that actually shows how many accepted replacements you've submitted

S87GMIL:
@DonovanDMC That's a good point, replacements by the user aren't really that visible right now.
I will experiment with the idea, and create another branch and pull request if I find a good solution.

S87GMIL:
Because of the pull request #572 I thought about changing the pending link again, so it only displays posts I uploaded that have pending replacements.
The problem is, that it seems like the original uploader filter simply doesn't work.
In my tests, I replaced a post uploaded by the admin, but it still wouldn't show up in the search, even when searching with the user ID:
image

Sindrake:
I kinda question whether this is necessary.
The pending replacements would get displayed on top of everything else as is, and this change would require the user to make more mouse clicks to see all of the user's replacements.

Sure, the link's label is somewhat misleading, but I don't think it's a huge issue in the long run.

faucetlol:
> I kinda question whether this is necessary. The pending replacements would get displayed on top of everything else as is, and this change would require the user to make more mouse clicks to see all of the user's replacements.

Agreed. A little confusing, but most people actively using the link are going to be janitors that understand how the page works. No need to make them click more to navigate to everything.

The label could perhaps be changed to something like "created" for less ambiguity? I think that makes it a little clearer, also solving the other problem that the replaced number is the number of replaced posts for which I'm the original creator, while [pending] is for replacements of other people's posts that were created by me - not pending replacements for my posts.

PR #572, opened 2023-12-16
Added the count of replaced posts to the user page

S87GMIL:
As suggested by @DonovanDMC in #569 it would be great to see approved and pending replacements created by a user on their page.
With this change, a new replaced post counter will be added to the user stats:
image

### DIFF
--- a/app/presenters/user_presenter.rb
+++ b/app/presenters/user_presenter.rb
@@ -88,6 +88,10 @@ class UserPresenter
     template.link_to(user.post_replacement_rejected_count, template.post_replacements_path(search: { creator_id: user.id, status: "rejected" }))
   end
 
+  def approved_replacements_count(template)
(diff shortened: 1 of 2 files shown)


Sindrake:
This is not a terrible idea, but I see a couple of issues here.

The big problem is that the profile page is incredibly ugly and user-unfriendly due to the sheer number of various stats on display in one giant table. I do want to address it at some point, but I'm not sure if we should be making it _worse_ in the meantime.

A more specific problem is that there are two values in this table: one is labeled "Replaced" (shows how many of this user's uploads had been replaced), and one labeled "Replaced Posts" (shows how many replacements this user created). That would be rather confusing for the users.

DonovanDMC:
At the bare minimum, using "Replacements" rather than "Replaced Posts" would be somewhat less ambiguous
The replacement stats are already haphazardly thrown in, with "Rejected" just hanging out with no qualifier for what's been rejected. It's obviously replacements, but... why isn't that included in the name

PR #526, opened 2023-05-30
EditHistories overhaul & search

DonovanDMC:
### Changes
* [EditHistories] Adds a new `edit_type` column to EditHistory entries. Currently, there's:
  * `original` - `version=1`, nothing special
  * `edit` - any content change
  * `hide`/`unhide` - hiding or unhiding
  * `stick`/`unstick` - sticking and unsticking
  * `mark_warning`/`mark_record`/`mark_ban` - marking for a warning/record/ban
  * `unmark` - removing any mark
* [EditHistoriesController] Adds searching to `/edit_histories` (`versionable_type`, `versionable_id`, `edit_type`, `user_id`, `user_name`, `ip_addr` (admin))
  * `edit_type=original` is excluded, unless searching for a specific edit type (`original` is still not included in the dropdown)
* [Blips/Comments/ForumPosts] Add timestamp for hiding to the "(hidden)" text (`title` attribute, shown when hovering)
  * This does remove the abilitiy of users immediately seeing who hid their comment, though they can still find out by looking at the modactions
  * [Example](https://github.com/e621ng/e621ng/assets/17226394/bc282c9f-06e7-431b-b7ff-c72627428853)
* [EditHistories] Add "Listing" link to "secondary links" to return to index
* [EditHistories#index] Add `|| ""` to subject column to fix `false` spam
* [EditHistories#index] Change "Show" link to actually link directly to that edit (via anchor)
* [EditHistories#show] Add link to "Edits for X #0" header (e.g. "Edits for [Comment #0](http://localhost:3000/comments/0)")
  * This is messy currently, it could potentially be cleaned up if paths can be referenced within models
* [EditHistories#show] Add an `id` to divs (`edit-0`) for linking
* [EditHistories#show] Make time clickable for link, similar to how comments and such behave
* [EditHistories#show] Add numbered paginator
* [WarningNotice] Display timestamp & user inline (when known)
  * [Known Example](https://user-images.githubusercontent.com/17226394/242752573-76a1171f-883d-4309-ac09-eee5eb982cad.png)
  * [Unknown Example](https://user-images.githubusercontent.com/17226394/242752721-c4c6e8c0-42a7-4f55-a6fc-e8117eacda08.png)
  * "Unknown" should only happen on blips marked before this pr, due to `warning_user_id` never being set, and blips not having an `updater_id` we can use for assumptions
* [UserStatistics] Adds link to list forum post/comment edits on user profile

### Tests
[...]

PR #574, opened 2023-12-18
Add searching by disapprovals

mm12:
CHANGES: 
- Added `disapprover:<name|any|none>` metatag - where user has disapproved a post
- Added `disapprovals:<n|any|none>` metatag - where there is a number of disprovals on a post
- Added `order:disapprovals` term, as well as `asc` and `desc` versions - orders by the number of disapprovals on a post, then ID
- Added Unit tests for metatags
- Added post index fields 
- Added DB Fix for post index fields
- Added autocomplete for users

### DIFF
--- a/app/indexes/post_index.rb
+++ b/app/indexes/post_index.rb
@@ -35,6 +35,7 @@ module PostIndex
           tag_count_invalid: { type: "integer" },
           tag_count_lore: { type: "integer" },
           comment_count: { type: "integer" },
+          dis_count: { type: "integer" },
 
           file_size: { type: "integer" },
           parent: { type: "integer" },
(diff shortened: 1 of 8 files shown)


mm12:
oops, i messed up my squashing

mm12:
Okie I think i fixed it... my bad

mm12:
I might be dumb

mm12:
Alright sorry for the mess of a PR

DonovanDMC:
FYI the fixer number in this pr (110) conflicts with #571 (#526), up to 116 is already in used in a pr, so the next one that's unused would be 117
I'm tracking these in Discord: https://discord.com/channels/431908090883997698/1255244495533113354/1264631270403608597

mm12:
I would still like this, but a new implementation seems favorable here. #1718 has a superior implementation and can be used as a base.

PR #576, opened 2023-12-21
[PostSets] require post_ids[] for adding/removing posts

DonovanDMC:
This corrects the error of attempting to map `params[:post_ids]` before verifying it's actually present in `add_posts` & `remove_posts` of the post sets controller.
image
image


Earlopain:
Use required parameters https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require

`permit(ids: []).require(:ids)` To only allow a non-empty array

DonovanDMC:
Since we're doing this at the top level, I had to exclude id/format else it threw an unpermitted parameters error

PR #577, opened 2023-12-21
[ForumTopics] Return no content instead of redirecting when marking all as read

DonovanDMC:


DonovanDMC:
Wish I could force rerun tests since that's one of the bullshit failures

PR #580, opened 2023-12-29
[Artists] Show error when changing notes with locked wiki page

DonovanDMC:
This pr makes an actual error show when editing an artist that has a locked wiki page instead of silently discarding the contents.
image

Includes a test to both ensure the wiki page is not modified, and that the error is generated.


Earlopain:
I think this should just deactivate the input field if you can't edit it in the first place

DonovanDMC:
Updated it to disable the field, and show a little text as to why it's disabled
image

PR #581, opened 2024-01-01
[UserFeedback] Add Updater

DonovanDMC:
This pr adds an updater to user feedbacks, as currently they do not store this information. This results in odd things like the update dmail claiming the original record creator updated the record, even if they were not the updater.

A relevant test has been included.

image
image

### DIFF
--- a/app/models/user_feedback.rb
+++ b/app/models/user_feedback.rb
@@ -2,6 +2,7 @@ class UserFeedback < ApplicationRecord
   self.table_name = "user_feedback"
   belongs_to :user
   belongs_to_creator
+  belongs_to_updater
   validates :body, :category, presence: true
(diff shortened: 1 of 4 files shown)


Earlopain:
Looks good but you have removed the blips constraint from the schema file

DonovanDMC:
Should be fixed out, I thought I had already fixed that with the previous force push, but apparently not

DonovanDMC:
Rubocop is the bane of my existence

Earlopain:
Just a note, adding the updater column for something that shows the update notice will result in all entries to be shown as updated by anonymous. The same thing happenend with blips last time and I had to manually set updater_id = creator_id to fix that. I'll do the same here but still something you should be aware of

DonovanDMC:
I had noticed it, I just forgot to add a fixer for it 🙃

PR #582, opened 2024-01-03
Integrate the avoid posting list into the site

DonovanDMC:
This pr adds an avoid posting feature directly into the site, along with a related history.

### Creating An Entry
Tied to the `is_bd_staff` flag. An artist entry existing is not a requirement
image

### Viewing An Entry
The delete button just deactivates the entry, the destroy button fully removes it.
image
(unprivileged view)
image
Other information such as linked user & other names will be shown, if present
image

### Index
The delete button here deactivates the entry.
image
(unprivileged view)
image

### Static Page
This pr adds a static page (https://e621.net/static/avoid_posting) which would provide the same functionality as our existing wiki page. The top section is pulled from the `e621:avoid_posting_notice` wiki page, and the bottom is constructed dynamically from the avoid posting entries.
image

### Artist
On artists, a bit of text is added that says the artist is on the avoid posting list if they have an active entry, showing the link and the details, if present.
image

Users are prevented from editing the name, other names, or group name of an artist that has an active avoid posting entry. This is tied to the `is_bd_staff` flag.
image

Outside of the pr, I added an `is_staff?` method, intended to be used when we only need to know they're a staff member, and we don't care about their level.

Each of artist & avoid posting entry comes with the ability to rename the other when renaming.
image
image

There are checks to ensure we aren't attempting to override existing artists and avoid posting entries.

A section under the quick mod section was added to posts that lists the avoid posting artists on the post.
image
This shows only on pending posts, though it's currently set to be visible to all users.

Some code has also been added that makes uploads always pending if it contains a dnp artist, unless the uploader is an approver.


Earlopain:
Having this keyed to an artist doesn't really work. There aren't many entries that make trouble, but they do exist. I would just make it a tag text column.

If janitors rely on this info being present there shouldn't be edge cases that don't work, and I don't like potentially creating dummy artist entries to work around that.

DonovanDMC:
I hate merge conflicts

DonovanDMC:
Updated to be based on only the name of the artist
Editing the artist name now also updates the dnp entry

DonovanDMC:
I'm absolutely stumped as to why this isn't working properly. I'm stuck on making the name/other names/group name uneditable.

The errors show up just fine:
image
image


But in the end, the changes still happen:
image
image

DonovanDMC:
I have practically everything done, even tests written, but I'm stuck on this one last thing

Sindrake:
> I have practically everything done, even tests written, but I'm stuck on this one last thing

Are you still stuck on that one thing?

DonovanDMC:
> > I have practically everything done, even tests written, but I'm stuck on this one last thing
> 
> Are you still stuck on that one thing?

Yes

DonovanDMC:
> > > I have practically everything done, even tests written, but I'm stuck on this one last thing
> > 
> > 
> > Are you still stuck on that one thing?
> 
> Yes

I think I've finally gotten it, I resulted to using `before_validation`, though that doesn't feel right. It works though, and that was what I was stuck on.

DonovanDMC:
@Earlopain This should be ready now, whenever you want to review it

DonovanDMC:
image

That commit adds a static page similar to the wiki page we already use, as requested by nimmy

It's formatted entirely within dtext for performance

app/controllers/avoid_postings_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def update
+    @avoid_posting.update(avoid_posting_params(:update))
+    flash[:notice] = @avoid_posting.valid? ? "Avoid posting updated" : @avoid_posting.errors.full_messages.join("; ")
+    respond_with(@avoid_posting)
+  end
+
+  def destroy
+    @avoid_posting.destroy
+    redirect_to artist_path(@avoid_posting.artist), notice: "Avoid posting destroyed"
```
Sindrake: **ActionController::UrlGenerationErro**r
No route matches {:action=>"show", :controller=>"artists", :id=>nil}, possible unmatched constraints: [:id]

app/models/avoid_posting.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
+
+class AvoidPosting < ApplicationRecord
+  belongs_to_creator
+  belongs_to_updater
+
+  has_many :versions, -> { order("avoid_posting_versions.id ASC") }, class_name: "AvoidPostingVersion", dependent: :destroy
+  before_validation :validate_artist_rename_not_conflicting, if: :will_save_change_to_artist_name?
```
Sindrake: Normalize the artist name.
Screenshot 2024-07-29 095057

app/views/posts/partials/show/_avoid_posting.html.erb
```diff
@@ -0,0 +1,33 @@
...
+          <% verified = dnp.artist && dnp.artist.linked_user_id == post.uploader_id %>
+          <li>
+          <% if dnp.details.blank? %>
+            <% if verified %>
+              <span class="artist"><%= link_to dnp.artist_name, avoid_posting_path(dnp) %> <i class="uploader-is-artist fa-regular fa-circle-check" title="The uploader is linked to this artist."></i></span>
+              <span class="separator">&dash;</span>
+              <span class="details">Only the artist is allowed to post.</span>
+            <% else %>
+              <span class="artist"><%= link_to dnp.artist_name, avoid_posting_path(dnp) %></span>
+              <span class="separator">&dash;</span> Avoid posting.
```
Sindrake: Screenshot 2024-07-29 095227
There shouldn't be line breaks here.

app/views/posts/partials/show/_avoid_posting.html.erb
```diff
@@ -0,0 +1,33 @@
+<% if post.avoid_posting_artists.any? %>
+  <div class="notice" id="avoid-posting-notice">
+    <div class="avoid-posting">
+      <h3>Avoid Posting</h3>
+      <ul>
+        <% post.avoid_posting_artists.each do |dnp| %>
+          <% verified = dnp.artist && dnp.artist.linked_user_id == post.uploader_id %>
+          <li>
+          <% if dnp.details.blank? %>
+            <% if verified %>
```
Sindrake: It seems to me that it should display "Only the artist is allowed to post" whenever:
- The DNP details are blank
- An account is linked to the artist page

In your implementation, the DNP reason would change between "Only the artist is allowed to post" and "Avoid posting" depending on whether the artist account ID matches that post's uploader's ID. While technically accurate, I feel like this would create confusion for the janitors.

app/views/avoid_postings/index.html.erb
```diff
@@ -0,0 +1,68 @@
...
+  <div id="a-index">
+    <h1>Avoid Postings</h1>
+
+    <%= render "search" %>
+
+    <table class="striped autofit">
+      <thead>
+      <tr>
+        <th></th>
+        <% if CurrentUser.is_staff? %>
```
Sindrake: You should put artist names first, details in the middle, and everything else after.

Screenshot 2024-07-29 102341

I think the layout makes a lot more sense like that.

app/views/avoid_postings/index.html.erb
```diff
@@ -0,0 +1,68 @@
...
+          <th>Creator</th>
+        <% end %>
+        <th>Status</th>
+        <th>Artist</th>
+        <th>Other Names</th>
+        <th>Details</th>
+        <% if CurrentUser.is_staff? %>
+          <th>Staff Notes</th>
```
Sindrake: I think I would prefer to just put the staff notes in the same column as the DNP details, like this:

Screenshot 2024-07-29 101943

I don't know how common these staff notes are going to be, so I'd rather not add another column just for them right now. We'll see how things go in the future.

Sindrake (changes requested):

app/views/avoid_postings/index.html.erb
DonovanDMC: The staff notes are intended to be private while the details are public, that's why they're separate

app/views/avoid_postings/index.html.erb
```diff
@@ -0,0 +1,68 @@
...
+          </td>
+          <td>
+            <%= link_to avoid_posting.artist_name, show_or_new_artists_path(name: avoid_posting.artist_name) %>
+          </td>
+          <td>
+            <% if avoid_posting.artist.present? %>
+              <%= link_to_artists(avoid_posting.other_names) %>
+            <% end %>
+          </td>
+          <td><%= avoid_posting.details %></td>
```
Sindrake: DText does not work here, is this intentional?

Screenshot 2024-07-29 120826

Sindrake (changes requested):

app/views/avoid_postings/index.html.erb
DonovanDMC: It's intentional on the versions index, but was not intentional here

DonovanDMC:
TiL you can't test multiple conditions with a single string, `assert_difference("ModAction.count AvoidPostingVersion.count")` when evaluated results in code more akin to `ModAction.count(AvoidPostingVersion.count)`, the argument is discarded and we end up only checking `ModAction.count`

I don't see anywhere in the existing code where this issue is present, but it's something to make a mental note of

Sindrake (approved):

Sindrake:
69 commits. Nice.

PR #586, opened 2024-01-09
[Blacklist] Redesign and rework the blacklist

Sindrake:
This pull request replicates re621's extended blacklist to the vanilla site.
This includes both a redesign of the blacklist UI, and adding more usable metatags.

### Sidebar 

The blacklist UI in the sidebar had been made more clear and easy to use.
Before, it was not immediately obvious to some users that the crossed-out filters were disabled.
Muting the colors where appropriate, and adding checkboxes that reflect the filter's state, should help with that.

Additionally, the number of posts next to the blacklist header updates dynamically when individual filters are enabled or disabled, and reflects the total number of posts that had been hidden by the blacklist.

<br clear="left"/>

### Inline

inline

The inline blacklist UI looks more or less the same as before.
Under the hood, it is identical to the sidebar UI, just with some style changes.

### Quick edit form

The quick edit form used to be fairly small and ugly. The actual textarea used to be just 263x138 pixels in size.
That is quite unacceptable – quite a few of my blacklist entries wouldn't fit on one line.

In addition to increasing its size, proper padding had been added to the form's elements.

<br clear="left"/>
<br />

### Persistence

Previous blacklist implementation could only save its state between different pages as all or nothing: either all filters are disabled, or all are enabled.

Now, if an individual filter is disabled on one page, it will remain disabled on another.
"Enable All Filters" clears the saved filter states on all pages. Additionally, the script will check that the filters in the disabled list actually still exist on startup, to prevent entries from getting stuck in the list despite being removed from the blacklist.

#### Search page 

User disables some filters on the search page.
<br clear="left"/>

#### Post page

When the user navigates to a post page, the same filters will remain disabled.
<br clear="left"/>

### Metatags <a id="metatags"></a>
The table below lists the available metatags.
Blue checkmarks indicate previous functionality, green ones – tags that had been added in this PR.

"Equality" indicates an exact match: `score:100`.
"Comparison" is relative: `score:>100`.
"Range" supports a range of values: `score:50..100`.

|                  | Equality | Comparison | Range |
|       ---      |     :---:    |        :---:        |   :---:   |
| id             |     ☑️      |        ✅        |    ✅   |
| status       |     ☑️      |                    |             |
| rating       |     ☑️      |                    |             |
| type          |     ✅      |                    |             |
| width        |     ☑️      |        ✅        |    ✅   |
| height      |     ☑️      |         ✅        |    ✅   |
| filesize <sup>1</sup>    |    ✅      |         ✅        |    ✅   |
| score        |     ☑️      |        ☑️        |    ✅   |
| favcount   |     ✅      |        ✅        |    ✅   |
| fav:me <sup>2</sup>   |     ☑️      |                    |             |
| user          |     ☑️      |                    |             |
| userid       |     ☑️      |         ✅        |    ✅   |
| username |     ✅      |                    |             |
| pool         |     ✅      |                    |             |
| tagcount  |     ✅      |         ✅        |    ✅   |

<sup>1</sup> `filesize` accepts both numerical values, or ones in `kb` or `mb`
<sup>2</sup> `fav:me` can only refer to the user's own favorites, same as before.

### Comments
The blacklist now supports proper comments, both inline and standalone.
Anything after the `#` symbol will be ignored by the blacklist.

All of the following are valid:
```
horse male solo # will remove later
rainbow lizard #scaliespe`
# artists I hate:
##### break #####
```

### Whitelist
Any line that starts with the characters `W! ` will be interpreted as a whitelist filter.
If a post matches the other tags on that line, it will never be hidden by the blacklist, even if it matches other filters.

### Click to Reveal
On the post page, clicking on the blacklist placeholder will reveal the post without disabling any of the filters.

### DIFF
--- a/app/controllers/concerns/deferred_posts.rb
+++ b/app/controllers/concerns/deferred_posts.rb
@@ -7,7 +7,7 @@ module DeferredPosts
 
   def deferred_posts
     Post.includes(:uploader).where(id: deferred_post_ids.to_a).find_each.reduce({}) do |post_hash, p|
-      post_hash[p.id] = p.minimal_attributes
+      post_hash[p.id] = p.thumbnail_attributes
       post_hash
     end
   end
--- a/app/decorators/posts_decorator.rb
+++ b/app/decorators/posts_decorator.rb
@@ -16,32 +16,13 @@ class PostsDecorator < ApplicationDecorator
     klass << "post-rating-safe" if post.rating == 's'
     klass << "post-rating-questionable" if post.rating == 'q'
     klass << "post-rating-explicit" if post.rating == 'e'
-    klass << "post-no-blacklist" if options[:no_blacklist]
+    klass << "blacklistable" unless options[:no_blacklist]
     klass
   end
 
   def data_attributes
-    post = object
-    attributes = {
-        "data-id" => post.id,
-        "data-has-sound" => post.has_tag?("video_with_sound", "flash_with_sound"),
-        "data-tags" => post.tag_string,
-        "data-rating" => post.rating,
-        "data-flags" => post.status_flags,
-        "data-uploader-id" => post.uploader_id,
-        "data-uploader" => post.uploader_name,
-        "data-file-ext" => post.file_ext,
-        "data-score" => post.score,
-        "data-fav-count" => post.fav_count,
-        "data-is-favorited" => post.favorited_by?(CurrentUser.user.id)
-    }
-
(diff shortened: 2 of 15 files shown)


Sindrake:
Back to draft.

Gonna have to rewrite the way the site resolves aliases in the blacklist.
Not only is it ugly, but it also eats comments.

Sindrake:
There is no end to the feature creep.

PR #589, opened 2024-01-11
[Posts] Tweak the janitor toolbar

Sindrake:
Minor changes to the janitor toolbar on the post page.

modbar

Re-ordered and renamed a couple of buttons.
Now the main actions are on the very left: "Approve", "Delete", "Disapprove".

The biggest change is the "Duplicates" count – it checks IQDB for similar posts.
Obviously, the button would not be displayed if there are no duplicates.


Earlopain:
Doing IQBD almost quadruples the page load time from 0.15s to 0.55s which I find unacceptable.

Sindrake:
Yeah, it's quite unfortunate
But it's only approvers that are affected, not regular users.

faucetlol:
What would happen if an approver wanted to open multiple tabs at once, considering the hard limit of one IQDB request per second?

Earlopain:
The limit is imposed on the controller level and as such would not take effect here. There would definitly need to be error handling though, should IQDB be unavailable. Regardless, I'm not going to accept the increased page load for this, it's just too high.

The other thing looks fine, though the "Detailed Deletion/Disapproval" buttons are removed?

Edit: Nevermind, it's only just the big bold buttons that do this now.

Sindrake:
Okay, I'll remove the duplicates checker.

The "Detailed Deletion / Disapproval" buttons are still there. They are now named just "Delete" / "Disapprove"

DonovanDMC:
What about doing the duplicate check on the front end?  Just do that request after the page has finished loading
(If ratelimits are really an issue, add a route only Janitor+/approvers can use with higher/no limit besides the global limit (though it's 1/2, that shouldn't really be an issue)

Sindrake:
I don't think a special route is needed, could just run the lookup when the page comes into focus.
... and if an approver is done with the post in less than two seconds, then it's their own fault.

But I think that's something that would be more appropriate for the ModTools userscript for now.

PR #591, opened 2024-01-16
[Posts] Show file type in the post sidebar

Sindrake:
By popular demand, add the file type to the post's sidebar.

filetype

### DIFF
--- a/app/views/posts/partials/show/_information.html.erb
+++ b/app/views/posts/partials/show/_information.html.erb
@@ -28,6 +28,9 @@
     Size: <span itemprop="width"><%= post.image_width %></span>x<span itemprop="height"><%= post.image_height %></span> (<%= number_to_human_size(post.file_size) %>)
   </li>
   <li>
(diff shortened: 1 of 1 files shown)


faucetlol:
I was thinking it would be nice if the format could be the same as the IQDB results and save adding an entire new line to the sidebar

image

...but now I look at it, the size line is already one of the longest lines on the sidebar and it doesn't need to be any longer.

Sindrake:
Yeah, that line is already too long, so it's not an option.
It's a miracle that it does not overflow normally.

It will definitely overflow if a post maxes out all the limits.
overflow
But I don't believe that happened yet.
Or, at the very least, it's infrequent enough that I don't normally notice it.

faucetlol:
> It will definitely overflow if a post maxes out all the limits.

Luckily, a file that maxes out the filesize limit displays as "100 MB" rather than "100.0 MB", just barely scraping by overflowing the line. But that does show how close to the edge it's being pushed already.

PR #592, opened 2024-01-18
[Forums] Overhaul handling of AIBUR requests

DonovanDMC:
This pr does the following:
* Makes AIBUR forum posts impossible to hide by non-moderators
  * This results in a plain 403 for json hiding forum posts due to the fucky way permissions are handled. I don't have the motivation to tackle those just yet, though they definitely need some TLC.
* Disallows editing of AIBUR forum posts (and by extension, topics)
* Moves the AIBUR text outside of the forum post (fixer included to remove current text)
* Adds a polymorphic column to forum posts storing the tag change request they reference, if any (fixer included)
  * This lowers the queries per any given forum post for AIBUR from almost certainly 3 (no AIBUR relation), to 0 (no AIBUR relation), or 1 if it has one. Previously, forum posts would query each of alias, implication, and bulk update request in sequence to check if any exists that references the forum post. This results in more queries than needed.


faucetlol:
I'm sure this has already had internal staff discussion, but I'm really not a fan of people having their ability to hide posts taken away, for privacy reasons.

I do understand the importance of being able to refer to previous discussions. Would some form of anonymization (obviously where staff can still see who posted it) be an acceptable alternative?

DonovanDMC:
anonymization isn't really possible with the way e621 works, especially not in a way still visible to staff
The only other alternative we had was specifically disallowing hiding pending, approved, and rejected with replies

ex:
* pending - can't hide
* approved - can't hide
* rejected, no replies - can hide
* rejected, with replies - can't hide

Pending, rejected with replies, and approved AIBUR will very likely be unhidden if any Moderator+ sees them currently, so that'd be practically no different, though just disallowing it outright is simpler

What privacy concerns can an AIBUR have anyways? They can just edit anything sensitive out of the post if they're dumb enough to put something there, or report it and have a mod hide it

Further reiterating not getting the privacy bit:
* Forum posts can always be edited, unless marked with a warning
* BURs can be edited until rejected/approved, then it's set in stone either way (always visible on https://e621.net/bulk_update_requests)
* AI can't be edited by normal users period, and it's once more set in stone (always visible in the respective list)

Sindrake:
I do agree with the PR in general, as having people hide AIBUR threads is pretty irritating.

However, I am worried that some users will simply remove the contents of the message when they discover that they can't hide it.
One way to combat that would be to prevent editing the post too, perhaps after a certain period of time has elapsed.

DonovanDMC:
> I do agree with the PR in general, as having people hide AIBUR threads is pretty irritating.
> 
> However, I am worried that some users will simply remove the contents of the message when they discover that they can't hide it. One way to combat that would be to prevent editing the post too, perhaps after a certain period of time has elapsed.

I had also planned on trying to prevent the removal of the initial line that includes the actual alias, ideally we can check if the text is there on edit, and if not prepend it to the body

I don't think prevent editing after any time will be productive, as AIBUR can last for, and change over months

DonovanDMC:
> I had also planned on trying to prevent the removal of the initial line that includes the actual alias, ideally we can check if the text is there on edit, and if not prepend it to the body

I've done just that. With this, it should be impossible to remove an AIBUR from the forum post it's linked to

Sindrake:
> I've done just that. With this, it should be impossible to remove an AIBUR from the forum post it's linked to

Yeah, that works.

Although, if I am being completely honest, I think that the whole idea of putting the AIBUR contents into the message via a DText tag is somewhat flawed. If a forum post has an AIBUR associated with it, can we just output the actual AIBUR contents above the message body? This way, the message would contain just the AIBUR reason – and the user can edit it as much as they want to.

DonovanDMC:
> > I've done just that. With this, it should be impossible to remove an AIBUR from the forum post it's linked to
> 
> Yeah, that works.
> 
> Although, if I am being completely honest, I think that the whole idea of putting the AIBUR contents into the message via a DText tag is somewhat flawed. If a forum post has an AIBUR associated with it, can we just output the actual AIBUR contents above the message body? This way, the message would contain just the AIBUR reason – and the user can edit it as much as they want to.

I had also thought about that, but wasn't really fond of trying to mess with that. Though from looking at it, it should be pretty easy

DonovanDMC:
Done that, that was a hell of a lot easier than I thought it would be. They still could be combined into one text container, though I don't think it matters that much

Now all that's left to do is create a fixer, _and fix rubocop for the 47th time_

DonovanDMC:
There's a fixer, tested it and it seems to have worked fine

it should strip the command string and any trailing newlines

DonovanDMC:
That's (hopefully) one last force push for the road

DonovanDMC:
I lied, one more force push

I've now added a polymorphic column to forum posts that stores the AIBUR related to them, so they don't need to make up to 3 queries every single time a forum post is loaded, they'll now make at most one

PR #594, opened 2024-01-20
[Misc] Add favicons to some external links

Sindrake:
This pull request allows favicons to be displayed next to external links.
Right now, that is primarily used to decorate post source links and artist page external links.

<br clear="left" />
<br />

Note that not all favicons would be automatically displayed.
As requested, this functionality does not rely on any external services, and all favicons are stored locally.
That allows us to combine them into one sprite sheet. The resulting file is fairly lightweight, and would get cached client-side.

favicons


faucetlol:
I like the idea, but maybe all that color on the sidebar of every post could be a little too distracting? Monochrome icons might be a little easier on the eyes.

Also, as a bonus, here's a larger Toyhouse icon I made in the past for Reverser, for the lack of any official icon on Toyhouse actually having a reasonable resolution.
image

Sindrake:
I don't think having colored icons in the sidebar would be distracting at all. It looks fine.

sidebar

I've looked at the reverser icons before. Some are okay, but some look terrible in 16x16.
Ones like Furaffinity or Newgrouns are just an abstract mess of pixels.
The Toyhouse icon is useful, though. Thanks.

DonovanDMC:
The alias section being index based seems like an absolute mess to maintain, could it not just map to a string, which would then replace the hostname we received? Then you should be able to do the lookup as normal

Sindrake:
> The alias section being index based seems like an absolute mess to maintain, could it not just map to a string, which would then replace the hostname we received? Then you should be able to do the lookup as normal

You are correct, but I think it's still a good idea to eliminate the extra step of looking up the index.
What do you think about having it look it up once and then store the value, like this?
```
DECORATABLE_ALIASES = {
  "e926.net" => DECORATABLE_DOMAINS.find_index("e621.net"),
  "discord.gg" => DECORATABLE_DOMAINS.find_index("discord.com"),
  "pixiv.me" => DECORATABLE_DOMAINS.find_index("pixiv.net"),
}.freeze
  ```
It's a little ugly, though.

Sindrake:
Okie-dokie, that was a lot of faffing.

- Tweaked the icons a bit for more visual consistency. With mixed results.
- Fixed a bug that caused an error if the link wasn't valid.
- Swapped the URL aliases to looking up their values instead of hard-coding them.
- Added `artconomy.com`, `steamcommunity.com`, and `furrystation.com` icons, and an alias for `mastodon.art`.

Let's hope that this is it.

Sindrake:
Spoiler: this wasn't it.

PR #595, opened 2024-01-22
[PostVersions] Allow searching for only uploads 

DonovanDMC:
This pr add an option for searching post versions by their version number, allowing for example, searching exclusively for uploads.

### DIFF
--- a/app/models/post_version.rb
+++ b/app/models/post_version.rb
@@ -41,7 +41,7 @@ class PostVersion < ApplicationRecord
 
       if params[:updater_name].present?
         user_id = User.name_to_id(params[:updater_name])
-        must << {term: {updater_id: user_id}} if user_id
+        must << { term: { updater_id: user_id } } if user_id
       end
 
       if params[:updater_id].present?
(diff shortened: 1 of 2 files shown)


Earlopain:
Is there actually a use-case for anything other than `version == 1` and `version != 1`? If not, I would prefer this to simply be combined into a select menu.

DonovanDMC:
> Is there actually a use-case for anything other than `version == 1` and `version != 1`? If not, I would prefer this to simply be combined into a select menu.

None I can think of, no. I can do that.

DonovanDMC:
Changed to a dropdown of `Included`, `Excluded`, `Only`, with included being the default

Should exclude uploads be kept in the background for backwards compatibility?

Earlopain:
Eh, seems fine to just drop. I changed this to include a blank since it would otherwise always be part of the url when searching unrelated things. Already happened for the checkbox but now it can be avoided very easily.

DonovanDMC:
> Eh, seems fine to just drop. I changed this to include a blank since it would otherwise always be part of the url when searching unrelated things. Already happened for the checkbox but now it can be avoided very easily.

Right, hadn't thought of that.

faucetlol:
The "Included" option is now redundant to the blank line, is it not?

Earlopain:
Sure, but that's how most things do it. Especially the order fields often contain their default values as an explicit entry.

DonovanDMC:
> The "Included" option is now redundant to the blank line, is it not?

Not including it also could make it seem like it isn't an option, and it's hard to actually click that blank line

PR #596, opened 2024-01-24
[ForumPosts] Fix voting buttons

DonovanDMC:
This pr fixes two issues with voting on forum posts.
* All voting buttons currently on the page disappear when voting on anything
* [Clicking outside of the `i` element makes voting fail](https://discord.com/channels/431908090883997698/1099361575116222656/1190249521771450398)

This pr also doesn't fully remove the vote buttons, they're kept around, so they can be restored when unvoting.

### DIFF
--- a/app/javascript/src/javascripts/forum_posts.js
+++ b/app/javascript/src/javascripts/forum_posts.js
@@ -10,7 +10,7 @@ ForumPost.initialize_all = function() {
       $("#edit_forum_post_" + forum_post_id).fadeToggle("fast");
       e.preventDefault();
     });
-  
+
     $(".edit_forum_topic_link").on("click.danbooru", function(e) {
       var link_id = $(this).attr("id");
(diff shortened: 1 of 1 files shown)


app/javascript/src/javascripts/forum_posts.js
```diff
@@ -78,14 +78,15 @@ ForumPost.vote = function(evt, score) {
...
-    type: 'DELETE',
-    dataType: 'json',
-    accept: 'text/javascript',
+    type: "DELETE",
+    dataType: "json",
+    accept: "text/javascript",
   }).done(function(data) {
     $(evt.target).parents(".own-forum-vote").remove();
+    $(`#forum-post-votes-for-${id} .forum-post-vote-block`).show();
     Utility.notice("Vote removed.");
```
Sindrake, line 90: Perhaps, you should consider removing the "Vote removed" notification?
It seems fairly redundant – it should be relatively obvious already that the vote was removed.

PR #597, opened 2024-01-24
[ForumTopics] Refactor subscriptions and add muting

DonovanDMC:
This pr adds muting forum topics. Muted topics will never show as `NEW`, and will not cause the "Forum" text to have the updated class. This can be useful for watchers of the forum that continuously see topics they have no interest in popping up.

ForumSubscription has been merged into ForumTopicStatus, to both avoid having multiple tables, and having muted and subscribed topics
I've currently added a fixer and left the ForumSubscription model alone, pending deletion. I am unsure if this could be done in one swift sweep via a migration, so I've gone the route of migrating now, cleaning up later.

The subscribe & mute actions each ensure their affected entity is either subscribed or muted, never both.

I made the mute/unmute actions PUT. Part of me wants to make subscribe/unsubscribe also PUT, but I don't want to do that without any input.


DonovanDMC:
I found out today through testing that as it was it would almost never show topics as unread due to it discarding any that don't have a `ForumTopicStatus`, I fixed that via using an outer left join, and added a test for the unread indicator so something like this doesn't get missed in the future

PR #598, opened 2024-01-24
[Comments] Add "Post Uploader" to search options

DonovanDMC:
Adds A "Post Uploader" input to the comments search (only shows when parameter is present), retaining the search when coming from the "comments on" links
image

Also added explicit `comments` table reference where the score column is used, due to the index page failing to load when grouping by comments, and joined to the post when no comments exist (ambiguous column `posts.score`/`comments.score`)
In theory this is an issue that will show up next to never, but there's also no real harm in explicitly referencing the table

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -124,7 +124,7 @@ private
   end
 
   def search_params
-    permitted_params = %i[body_matches post_id post_tags_match creator_name creator_id post_note_updater_name post_note_updater_id poster_id is_sticky do_not_bump_post order]
+    permitted_params = %i[body_matches post_id post_tags_match creator_name creator_id post_note_updater_name post_note_updater_id poster_id poster_name is_sticky do_not_bump_post order]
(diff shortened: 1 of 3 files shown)


Earlopain:
Please rebase. The column references were definitely a problem, the query was just flat out broken. I already fixed this in https://github.com/e621ng/e621ng/commit/0fad5d56598431ac2cd88def8470f188203f93c3

Also, I feel like the presence check before `where_user` is something I'd delegate to that method. Can you check if that explicit check is needed?

DonovanDMC:
> Please rebase. The column references were definitely a problem, the query was just flat out broken. I already fixed this in [0fad5d5](https://github.com/e621ng/e621ng/commit/0fad5d56598431ac2cd88def8470f188203f93c3)
> 
> Also, I feel like the presence check before `where_user` is something I'd delegate to that method. Can you check if that explicit check is needed?

Rebased

I left the explicit presence check in due to reordering, without that it shouldn't be needed

PR #601, opened 2024-01-26
[AIBUR] Link to forum post & show votes in `show` action

DonovanDMC:
This pr centralizes the **Reference** displayed on AIBUR, and changes the link to the forum post, if present. The link goes to the topic with an anchor, so they actually see the request in context, rather than needing to click through the timestamp.
This could probably be extended through most if not all of the AIBUR show actions (making them all into one template which picks and chooses based on the model), but that just sounds messy.

This pr also adds the votes list to AIBURs. Voting does not work, both intentionally (they should see the context the request was submitted in), and due to the way e621 works.

image
image
image


DonovanDMC:
I pray that one day, that test will be fixed

Earlopain:
Made some changes and made use of [strict locals](https://guides.rubyonrails.org/action_view_overview.html#strict-locals) (new in Rails 7.1, they're pretty cool)

PR #604, opened 2024-01-29
[Blacklist] Increase the size of the editor window

Sindrake:
This is a spinoff from PR #586.

quickedit

Increased the size of the blacklist quick edit window, and also tweaked the contents a little to make it less janky.
Also, increased the size of the blacklist editor on the settings page, that thing could fit like 5 lines at most by default.


Sindrake:
I knew I should have deleted the branches too.

PR #610, opened 2024-01-31
[Misc] External link favicons (now without a spritesheet)

Sindrake:
This is a reworked version of #594, but without merging the icons into a single spritesheet.

This pull request allows favicons to be displayed next to external links, based on [this request](https://github.com/orgs/e621ng/projects/2/views/1?pane=issue&itemId=18707641).
Right now, that is primarily used to decorate post source links and artist page external links.

<br clear="left" />
<br />

Note that not all favicons would be automatically displayed.
As requested, this functionality does not rely on any external services, and all favicons are stored locally.

favicons


Earlopain:
Thank you, this is great. I've made some changes, particularly to how the images are served:

https://github.com/e621ng/e621ng/pull/610/commits/9e3ce9c6c12e42a8bf4b6263178807e0804dd7b9 serves the images from the packs folder. This makes the url change when the image content changes (which I don't think will happen particularly often but cloudflare is extremly agressive with caching them). More importantly, that path isn't part of rate limiting, which could actually be important here since an artist can have 25 urls.

`https:example.com` actually parses to a uri just fine but would error if trying to modify the host (nil), fixed that up, added a few tests.

PR #611, opened 2024-01-31
[WikiPages] Implement basic redirects

Sindrake:
This PR allows one wiki page to seamlessly redirect users to another wiki page.
That is necessary to avoid awkward pages like [this](https://e621.net/wiki_pages/6) – users can just be shown what they were looking for instead.

redirect

Users are not technically redirected between pages, as the URL remains the same.
Instead, the contents of the parent wiki page get displayed instead of the original one.

redirect2

As far as permissions are concerned, I think it's okay to let Privileged+ to handle this.
Opening it up to the public is fraught with peril, and only letting staff members handle this seems unnecessary. Privileged users, on the other hand, are both trustworthy enough to be allowed to set up redirects, and also more likely to be creating wiki pages in the first place.

### DIFF
--- a/app/controllers/wiki_pages_controller.rb
+++ b/app/controllers/wiki_pages_controller.rb
@@ -53,6 +53,9 @@ class WikiPagesController < ApplicationController
     end
 
     if @wiki_page.present?
+      unless @wiki_page.parent.nil?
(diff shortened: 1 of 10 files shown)


Earlopain:
This needs versioning.

PR #617, opened 2024-02-03
[UploadPresenter] Normalize tags & strip metatags

DonovanDMC:
This pr normalizes tags given to the TagSetPresenter, so capitalization and prefixes won't result in tags not getting their category color. It also strips metatags (like `set:`), since those shouldn't really be shown for viewing.

I also added `co`, `ch` & `oc` to `TagCategory::SHORT_NAME_MAPPING` so they would be included in the regex.

And lastly, my editor was complaining at me that the rbdoc on line 10 didn't have the correct parameter name, so I added it.
image

### Before
image
### After
image

### DIFF
--- a/app/logical/tag_category.rb
+++ b/app/logical/tag_category.rb
@@ -105,4 +105,5 @@ class TagCategory
   CATEGORIZED_LIST = %w[invalid artist copyright character species meta general lore].freeze
 
   SHORT_NAME_REGEX = SHORT_NAME_LIST.join("|").freeze
+  ALL_NAMES_REGEX = MAPPING.keys.join("|").freeze
 end
--- a/app/presenters/upload_presenter.rb
+++ b/app/presenters/upload_presenter.rb
@@ -7,6 +7,16 @@ class UploadPresenter < Presenter
   end
 
   def tag_set_presenter
-    @tag_set_presenter ||= TagSetPresenter.new(upload.tag_string.split)
+    @tag_set_presenter ||= TagSetPresenter.new(normalize_tags(upload.tag_string.split))
+  end
+
+  def strip_metatags(tags)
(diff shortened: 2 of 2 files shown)


Sindrake:
Where did `co` come from as a short form of copyright?
I agree with `oc`, and it does seem like `ch` had some usage, but that's not the case for `co`.

DonovanDMC:
> Where did `co` come from as a short form of copyright? I agree with `oc`, and it does seem like `ch` had some usage, but that's not the case for `co`.

It's in the code, I added what we already had to the short name mapping
https://github.com/e621ng/e621ng/blob/f5fcd8a53b382a2bfec4a0854d3a72083a235b05/app/logical/tag_category.rb#L2-L21

Earlopain:
Is this a concern on any page except the uploads page? Feels like that is the only place this would occur. I would prefer this logic to live in the UploadPresenter instead, or adding a keyword arg to only normalize when explicitly requested.

Also, its not safe to add to `SHORT_NAME_MAPPING`. While it may work benefically in this case, other place can behave unexpectedly by these added fields, like new metatags that suddenly exist or trying to filter by a column thats not there.

DonovanDMC:
> Is this a concern on any page except the uploads page? Feels like that is the only place this would occur. I would prefer this logic to live in the UploadPresenter instead, or adding a keyword arg to only normalize when explicitly requested.

Not that I can think of. I've moved it into UploadPresenter

> Also, its not safe to add to `SHORT_NAME_MAPPING`. While it may work benefically in this case, other place can behave unexpectedly by these added fields, like new metatags that suddenly exist or trying to filter by a column thats not there.

I hadn't really thought of that. I'm unsure if we currently have a regex for a combination of full & short metatag names, so I added a new `ALL_NAMES_REGEX` constant to the TagCategory module.

PR #620, opened 2024-02-22
[DMails] Add marking as unread

DonovanDMC:
Adds a button to individual dmails to mark them as unread. This can be useful when you read a dmail now, but can't/don't want to reply to it until later.

Includes relevant controller test.
Also adds "Delete" button to individual dmails.

### DIFF
--- a/app/controllers/dmails_controller.rb
+++ b/app/controllers/dmails_controller.rb
@@ -1,13 +1,13 @@
 class DmailsController < ApplicationController
   respond_to :html
-  respond_to :json, only: %i[index show destroy mark_as_read mark_all_as_read]
+  respond_to :json, except: %i[new create]
   before_action :member_only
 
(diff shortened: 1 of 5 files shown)


app/controllers/dmails_controller.rb
```diff
@@ -1,13 +1,14 @@
 class DmailsController < ApplicationController
   respond_to :html
-  respond_to :json, only: %i[index show destroy mark_as_read mark_all_as_read]
+  respond_to :json, except: %i[new create]
   before_action :member_only
+  before_action :load_dmail, except: %i[index new create mark_all_as_read]
```
Earlopain: Really not a fan of this pattern. Just stick to how it was previously.

DonovanDMC: Done

PR #622, opened 2024-03-01
Convert Tag links in PostVersions & Uploads to wiki links

DonovanDMC:
Does just what it says on the tin, converts links for tags on /post_versions & /uploads to wiki page links.
image
OnPaste 20240301-043007

Wiki page links are typically much more helpful here, since you're more likely looking at wiki pages to see if those tags are applicable, and not trying to find new tags to search for.

### DIFF
--- a/app/presenters/tag_set_presenter.rb
+++ b/app/presenters/tag_set_presenter.rb
@@ -7,6 +7,8 @@
 =end
 
 class TagSetPresenter < Presenter
+  include Rails.application.routes.url_helpers
+
   attr_reader :tag_names
 
   # @param [Array<String>] a list of tags to present. Tags will be presented in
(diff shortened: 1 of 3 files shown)


app/presenters/tag_set_presenter.rb
```diff
@@ -166,8 +166,9 @@ def build_list_item(tag, current_query: "", highlight: false)
     html
   end
 
-  def tag_link(tag, link_text = tag.name)
+  def tag_link(tag, link_text = tag.name, link_type = :tag)
+    link_base = link_type == :wiki_page ? "/wiki_pages/show_or_new?title=" : "/posts?tags="
```
Earlopain: Do `include Rails.application.routes.url_helpers` up top and then you should be able to use the helpers like usual.

PR #625, opened 2024-03-16
New search syntax (tag grouping)

Tarrgon:
This PR aims to provide a new tag syntax to provide users with proper tag grouping to allow for more advanced search queries.

Big disclaimer right off the bat, **I'm not a ruby dev, I have over 10 years of experience in development in general, but I have never touched ruby**. This code is currently **not optimized and should needs a lot of love before being production ready**, which is why I've marked it as a draft. I just want to get the discussion around this topic going and if anyone wants to contribute to it, they are more than welcome to.

Current new syntax I plan is to mimic my advanced search syntax, just since I've already made it and tested it in javascript, so I know that it works.

This allows grouping tags mainly for different "OR" groups such as: `( female ~ male ) ( solo ~ duo )`. This will only match posts containing a female or male, and are either solo or duo, which is not currently possible as `~female ~male ~solo ~duo` would match all 4 of those tags separately. Meaning any of them have to match and you can get `female group` or `solo intersex` as that still passes the query.

### Why?
E6's search syntax rides on extremely old code. Not being able to make "OR" tag groups limits the possibilities to just basic searching. And while I've resorted to making a database mirror with a more powerful syntax, I think it's time for e6 to get an upgrade, so I want to bring it into the site directly.

#### Why `tag ~ tag`?
OR logic is a relationship between two or more tags. It makes 0 sense to apply "OR" logic to just 1 tag. `~tag` means nothing and that makes `~tag ~tag` even more confusing when `tag ~ tag` is obvious in its intentions.

Furthermore, the new syntax supports an "or not" capability that wouldn't make sense with the current way of doing things. `~-tag` just doesn't make a whole lot of sense since it's all together. Where as the separation `tag ~ -tag2` makes it more readable as "tag or not tag2"

### Syntax descriptor

New syntax is as follows:
- `(` opens a new group. Must be preceded and succeeded by a space (unless it's at the start, or is being negated), this is because we have tags like `character_(artist)` which use parentheses.
- `)` closes an open group. Must be preceded and succeeded by a space. Groups can exist within groups without issue if necessary. Such as: `( female ~ male ( solo ~ duo ) )` is valid, albeit maybe a little pointless.
- `-` negates the next tag or group. Must directly precede a tag or `(`. Entire groups can be negated (ex `-( female ~ male )` would match posts not containing `female` or `male`. But `-( female male )`  would match posts not containing BOTH `female` and `male` meaning that either tag can be present if the other isn't)
- `~` makes the preceding and succeeding tag or group (separated by a space) have `OR` logic together. Both tags and groups can be OR'd. `female ~ ( male solo )` is valid, for example, or `female ~ male ~ intersex` is also valid. OR logic requires at least 2 tags or groups to operate. It can be slightly confusing for example `solo female ~ male` matches posts with `solo` AND `female` OR `male`, which is why I prefer to just wrap ORs in groups by default like `solo ( female ~ male )`, but it isn't necessary. These match the same posts.

Other than that you search like normal, however `order` tags are always considered top-level. You can't order a specific group because that isn't really viable.

### Testing/Enabling
To test, enable the option to "Use New Search Syntax" in advanced user settings.

**This is a radically breaking change to the search functionality and should not be the default for existing users and api usage as that will break a lot of things.**

### TODO
- [x] Writing tests
- [x] Max group depth to prevent funny infinite loops essentially DOSing the server.
- [x] Error on unclosed group
- [x] Proper tag count resolution (don't include operators or group openers)
- [x] Add "Use new search syntax" option to user settings.
- [x] Tag searching
- [x] Meta searching (basically everything in the form of `key:value`, like `status:pending`)
  - [x] `id:`
  - [x] `mpixels:`
  - [x] `ratio:`
  - [x] `width:`
  - [x] `height:`
  - [x] `duration:`
  - [x] `score:`
  - [x] `favcount:`
  - [x] `filesize:`
  - [x] `change:`
  - [x] `date:`
  - [x] `age:`
  - [x] `tag_count:`
    - [x] Tag count categories
  - [x] `comment_count:`
  - [x] `md5:`
  - [x] `status:`
    - [x] Remove forced `-status:deleted` to let them show. This basically would mean the `-status:deleted` needs to move to all the groups that don't have any mention of status.
  - [x] `user:`
  - [x] `approver:`
  - [x] `commenter:`
  - [x] `noter:`
  - [x] `note:`
  - [x] `pool:`
  - [x] `set:`
  - [x] `fav:`
  - [x] `parent:`
  - [x] `child:`
  - [x] `rating:`
  - [x] `filetype:`
  - [x] `delreason:`
    - [x] Remove forced `-status:deleted` to let them show
  - [x] `deletedby:`
    - [x] Remove forced `-status:deleted` to let them show 
  - [x] `description:`
  - [x] `source:`
  - [x] `voted:`
  - [x] `upvote:`
  - [x] `downvote:`
  - [x] `inpool:`
  - [x] `locked:`
  - [x] `hassource:`
  - [x] `ischild:`
  - [x] `isparent:`
  - [x] `pending_replacements:`
- [x] Ordering
  - [x] `id`
  - [x] `change`
  - [x] `md5`
  - [x] `score`
  - [x] `duration`
  - [x] `favcount`
  - [x] `created_at`
  - [x] `updated`
  - [x] `comment`
  - [x] `comment_bumped`
  - [x] `note`
  - [x] `mpixels`
  - [x] `portrait`
  - [x] `landscape`
  - [x] `filesize`
  - [x] `tagcount`
    - [x] Tag count categories
  - [x] `rank`
  - [x] `random`


Tarrgon:
This code is still not production ready, just trying to get all the pieces in place for testing, then I should be able to move it around. Ideally the building should happen in the actual query builder... But right now I'm doing it in the tag parser just since everything was already there. I'll figure it out.

Tarrgon:
Ok, so the main thing holding back the meta tags right now is that since `-status:deleted` is essentially assumed on all queries, I need to intelligently back it down the chain when a status tag is present. This means that when `status:x` is present in a group, the global `-status:deleted` must be moved to all non-related groups until it no longer targets the current group.

Not entirely sure how I'm gonna do that yet, but I'll figure something out.

Tarrgon:
I've realized an issue in my code and apparently my testing wasn't working properly lol, will fix

Tarrgon:
Ok, all of that should be fixed now. Smaller testing seemed to work.

Tarrgon:
I believe I've got statuses working, will need more testing, though to ensure parity.

Tarrgon:
Ok, ordering was basically just a copy paste from the old system, just had to add the support for multiple orders. With that the syntax is "technically" feature-complete as it has all the meta tag capabilities and ordering capabilities of the normal syntax.

Now I'm going to be doing testing, reorganizing, more testing, trying to add the option to the user settings, more testing, and finally testing.

Tarrgon:
I've now moved the build logic to the `ElasticPostQueryBuilderNew`, this comes with the slight optimization of in-place meta-tag resolution rather than temporary placeholders.

Tarrgon:
Alright, now to test, just go to your user settings -> advanced and set "Use New Search Syntax" to "Yes" it defaults to "No" for new and existing users.

Tarrgon:
I'm not sure what other tests to write since it works pretty differently from the old tag query class.

Other than that, though, I think this is ready,

Tarrgon:
Still looking for optimizations, preferably to reduce the recursive nature of it. Right now there's technically 4 different recursive calls. One to build the query itself, and 3 to handle each of the types `must`, `must_not`, and `should`. It should technically be possible to do them all in the single recursive method, but it'll take some thought that I will need time to think about lol

Tarrgon:
It will now prioritize the url parameter over user settings if it's present to ensure that it always works as intended

Tarrgon:
Tomorrow since I have the day off, I'm going to be testing basically everything, just for sanity:

- [x] Tag searching
  - [x] Negated tags
  - [x] Wildcard tags
  - [x] Negated wildcard tags (I actually don't know about this one just yet, forgot it was possible)
  - [x] OR negated tags
  - [x] Groups 
    - [x] Groups in groups
    - [x] Negated groups
    - [x] OR negated groups
- [x] Metatag searching 
- [x] Ordering (including multi-ordering)
- [x] And all of these things at the same time.

Tarrgon:
After a few fixes, this all seems to work

Tarrgon:
Just tested aliases as well, all working there.

Tarrgon:
So now I suppose it's time to run the linter and have it yell at me. Forgot about that.

Tarrgon:
With all that out of the way, time to write more of those pesky tests to ensure it works at all times.

Tarrgon:
I actually can't think of any further tests to apply to it.

Tarrgon:
~~Merge is untested at the moment. Will ensure it still works as intended~~

Seems to still work

app/models/user.rb
```diff
@@ -48,6 +48,7 @@ module Levels
     _no_feedback
     disable_user_dmails
     enable_compact_uploader
+    use_new_search_syntax
```
DonovanDMC: You can't insert new preferences anywhere, they're positional. It *has* to be at the end, else the preferences after it will be shifted around

app/logical/tag_query_new.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+class TagQueryNew < TagQuery
+  class UnclosedGroupError < StandardError; end
+  class MaxGroupDepthExceededError < StandardError; end
+  class MaxTokensAchievedInGroupError < StandardError; end
+
+  attr_reader :q, :resolve_aliases
```
DonovanDMC: you don't need to redefine the readers since you're extending the base

app/models/user.rb
```diff
@@ -884,6 +885,10 @@ def compact_uploader?
     post_upload_count >= 10 && enable_compact_uploader?
   end
 
+  def use_new_search_syntax?
+    use_new_search_syntax
+  end
```
DonovanDMC: You don't need to define this, the HasBitFlags class already defines it
https://github.com/e621ng/e621ng/blob/3229106431dbf2636a4fb7e2586cc3ef7da5bf57/app/logical/danbooru/has_bit_flags.rb#L26-L28

app/views/users/edit.html.erb
```diff
@@ -95,6 +95,10 @@
           hint: "Enables a more compact and less guided post uploader." %>
         <% end %>
 
+        <%= f.input :use_new_search_syntax, :as => :select, label: "Use New Search Syntax",
+                    hint: raw("Enables the use of the new search syntax."),
```
DonovanDMC: you don't need this raw() here

DonovanDMC (commented):
Decided to give this a onceover and point out any simple issues

app/models/user.rb
Tarrgon: I was unaware the order was important. I'll move it.

app/logical/tag_query_new.rb
Tarrgon: Fixed 0c2b9d9156a63ecd5deeafe1d76162a270144367

app/models/user.rb
Tarrgon: Fixed 276b298b64e86871fbab7b131b865f9700a3d8a0

app/views/users/edit.html.erb
Tarrgon: Fixed c5d336ccffd0730223b2a6576963a76bde89f614

app/models/user.rb
Tarrgon: Fixed 9eff693ec34e2c879b4b2b78e82781af49330f99

Tarrgon:
Whoops rubocop didn't like that

app/models/user.rb
DonovanDMC: > I was unaware the order was important. I'll move it.

Ideally they wouldn't be, but instead of explicitly defining the values, we use the index value. The flag value is `1 << index`. 
Ex:
`1 << 0` = 1
`1 << 1` = 2
`1 << 2` = 4

These combine into a final number to say which are active, ex `3` (0 & 1), `5` (0 & 2), etc

Inserting a flag in the middle changes the index of later flags, so their value changes, making previous preferences now incorrect.
Ideally we'd explicitly define the value for each preference so the order doesn't matter, and so we can actually remove unused flags. The position being important is the reason those unused flags are still in there

Tarrgon: Got it. I'll keep that in mind for future PRs, though I doubt I'd be doing anything else preference related. I didn't even know it was a mask until you linked the `HasBitFlags` class with the

```rb
define_method("#{attribute}?") do 
  send(field) & bit_flag > 0 
end 
```

Part. I can't say I did an extremely deep dive to understand the code, which is why you saw a bit of copy pasting existing things and changing the content, like with the actual preference in the settings using the "raw" method, or the attributes, or the attribute accessors.

Tarrgon:
Also just gonna leave this question here: is raising the tag limit out of the question? Not by much, but with a more powerful syntax there will be actual reason to use more tags. Especially because, let's be honest here, since this isn't default not comparatively few people will use it and could probably have a higher budget of tags, it doesn't even really need to be a "marketed" feature of it, just something silently there if we think that it can cause issues on query times across the entire site if it was widely used.

Tarrgon:
I've found a flaw in the parsing of using both operators. Something that I didn't consider since the current syntax doesn't support it, either. Using both OR and NOT on the same tag. As in `tag1 ~ -tag2` meaning a post must have `tag1`, or it must not have `tag2`. Which currently parses to the current syntax's equivalent of `~tag1 -tag2`. Which is unexpected and I'll need to fix it.

Tarrgon:
I think this fixes it

mm12:
I still hate syntax changes.. I think it should be backwards compatible, at the least

Tarrgon:
> I still hate syntax changes.. I think it should be backwards compatible, at the least

Which is why this is an optional change. If I made it backwards compatible, there'd be no reason to do that, but the issue is that the current system would need a full rewrite to be able to even come close to supporting this while also being backwards compatible.

The simple solution if you dislike the syntax changes would be to not use the new syntax. Though I doubt this will get accepted regardless.

mm12:
> > I still hate syntax changes.. I think it should be backwards compatible, at the least
> 
> Which is why this is an optional change. If I made it backwards compatible, there'd be no reason to do that, but the issue is that the current system would need a full rewrite to be able to even come close to supporting this while also being backwards compatible.
> 
> The simple solution if you dislike the syntax changes would be to not use the new syntax. Though I doubt this will get accepted regardless.

Fair enough, but r34 is the only booru that i know that uses this syntax vs the backwards compatible version. Does r34 use Elastic and opensearch?

Tarrgon:
> Does r34 use Elastic and opensearch?

Couldn't tell you, I designed it off of what I personally believe to be the most readable syntax. Which to me includes separating multi-handed operators from tokens. And have given my reason for doing so in the original PR comment as to why I changed it.

> Why tag ~ tag?
OR logic is a relationship between two or more tags. It makes 0 sense to apply "OR" logic to just 1 tag. ~tag means nothing and that makes ~tag ~tag even more confusing when tag ~ tag is obvious in its intentions.

> Furthermore, the new syntax supports an "or not" capability that wouldn't make sense with the current way of doing things. ~-tag just doesn't make a whole lot of sense since it's all together. Where as the separation tag ~ -tag2 makes it more readable as "tag or not tag2"

Sindrake:
I'll be honest with you, while the idea is definitely solid, I am also not a huge fan of changing the format.

I would personally prefer something more compatible with the existing search patterns.
Having an operator not attached to a tag is not something we currently do.

May I suggest a more conventional format? Something like `~(tagA tagB)` would return results with either tagA OR tagB.
It would match your existing negation pattern.

Tarrgon:
> May I suggest a more conventional format? Something like `~(tagA tagB)` would return results with either tagA OR tagB.
> It would match your existing negation pattern.

There's 2 main issues with this imo.
First, parentheses directly attached to tags can cause confusion for tags that end with parentheses, which is why I've avoided it.

Second, this is less of an issue, more of a question, but how would this work with tag grouping? Since not every group is necessarily an OR group, it's important to be able to group without the OR operation being applied to every tag inside that group as that really matters with the order of operations.

Take the following query: `( tagA tagB ) ~ ( tagC tagD )` this equates to posts with both `(tagA AND tagB) OR (tagC AND tagD)`. This is mainly what people are after with tag grouping and ORs as with the current syntax, it's not possible to separate these two OR groups, meaning it's impossible to replicate the aforementioned query as all we really would be able to offer is `~tagA ~tagB ~tagC ~tagD` which isn't the same.

With the syntax you provided, how would you recreate the query I provided since the OR is between two groups?

I wouldn't be entirely opposed to something like `[( tagA tagB ) ( tagC tagD )]` where the square brackets denote an OR group, and parentheses within them group ANDs together. There are 16 tags with less than 100 posts that start or end with a square bracket, so adding "tags can't end with square brackets" to the tag validator wouldn't even really be something people would notice. But this also has the issue of not being similar to e621's current syntax, which is the main complaint for this new syntax.

With that, a simple OR query can still be `[tagA tagB]` which would be the same as `~tagA ~tagB`, since there's no parentheses, there are no AND groups to resolve out, so everything is OR'd together.

Tarrgon:
I've converted this back to draft for the time being as before anything happens to this, it needs major code revision and review, regardless of what the syntax may or may not be. This was one of my first times using ruby and it's a relatively complex PR, so there are definitely issues in there.

Tarrgon:
Closing this because I don't plan on continuing working on this feature anymore, and both the code itself and the format is unacceptable.

PR #632, opened 2024-04-06
[HelpPages] Fix showing related only to Janitor+

DonovanDMC:
Fixes the related help pages only being show to Janitor+ due to condition nesting. This happened in 069a009dde5a0f95aae280b68de663d5c46318c7


Earlopain:
What a silly mistake to make. Thanks

PR #636, opened 2024-04-10
[Artists] Normalize name in show & show_or_new

DonovanDMC:
This pr normalizes the name used for lookups (similar to #631). It also allows using the id or name for all actions that use `load_artist`.

I've also quietly removed the `source` parameter. It does nothing anymore, and using it throws an error.


app/controllers/artists_controller.rb
```diff
@@ -71,19 +71,18 @@ def destroy
...
   end
 
   def show_or_new
-    @artist = Artist.find_by(name: params[:name])
+    @artist = Artist.named(params[:name])
     if @artist
       redirect_to artist_path(@artist)
     else
-      @artist = Artist.new(name: params[:name] || "")
+      @artist = Artist.new(name: params[:name].present? ? Artist.normalize_name(params[:name]) : "")
```
Earlopain:
```suggestion
      @artist = Artist.new(name: Artist.normalize_name(params[:name] || ""))
```

Sindrake (approved):

PR #638, opened 2024-04-18
[Notes] Remove note previews, replace with dtext previews

DonovanDMC:
This PR returns the deferred posts from the previewed dtext. Currently these are ignored, which results in `thumb #` not working in the preview, but working after saving the note, then reloading the page.

This route is also mildly different from the other dtext response routes, instead of `{ html, posts }`, it now returns `{ body, posts }`, it's an easy fix to be consistent, but also a breaking change I'm not sure if we're comfortable with.


Earlopain:
Just have this use the dtext previews controller, there's barely any difference. This one can be removed, datadog tells me this gets one request every 15 minutes so chances are nobody will even notice during deploy when one can still be routed to it from non-deployed apps.

DonovanDMC:
> Just have this use the dtext previews controller, there's barely any difference. This one can be removed, datadog tells me this gets one request every 15 minutes so chances are nobody will even notice during deploy when one can still be routed to it from non-deployed apps.

Even better, will do.

DonovanDMC:
I assume we're just not caring that the dtext preview controller allows color while the note previews (and notes themselves) don't? We could "fix" that via a parameter that forces color to be disabled

Earlopain:
It's fine this way

.rubocop_todo.yml
```diff
@@ -510,7 +510,6 @@ Layout/SpaceInsideHashLiteralBraces:
     - 'app/controllers/comments_controller.rb'
     - 'app/controllers/help_controller.rb'
     - 'app/controllers/moderator/ip_addrs_controller.rb'
-    - 'app/controllers/note_previews_controller.rb'
```
Earlopain: I update this file occasionally, no need to do it for every change

PR #639, opened 2024-04-20
[WikiPages] Restrict skip_secondary_validations to Janitor+

DonovanDMC:
This pr restricts `skip_secondary_validations` to Janitor+.

In addition:
* The checkbox only shows to Janitor+
* The checkbox only shows if a title error includes "Move the posts and update any wikis linking to this page first."

Currently this option will show to anyone for any title error, which results in confusion. Due to the option being called "Force rename", this looks like an option that will allow creating wiki pages with invalid titles. This option does nothing for normal users since they cannot rename wiki pages.

I also think this option's checkbox could use a rename, since "Force rename" implies more than it actually does, but that's for another day.


app/views/wiki_pages/_form.html.erb
```diff
@@ -22,7 +22,7 @@
 
     <%= f.input :edit_reason, label: "Edit Reason" %>
 
-    <% if @wiki_page.errors[:title].present?  %>
+    <% if CurrentUser.is_janitor? && @wiki_page.errors[:title]&.any? { |error| error.include?("Move the posts and update any wikis linking to this page first.") }  %>
```
Earlopain, line 27: Are there even any other errors for titles? Relying this specifically on the error message is as bad as relying on the fact that there is only one. The label for this is good enough

DonovanDMC: Yes, providing an invalid title, which is the whole crux of this pr

image

I had originally only tested for "Move the posts", but threw in the rest for extra credit

Sindrake (approved):

PR #640, opened 2024-04-27
Add period to comment for consistency

SeesawEffectt:
Other comment sentences all ended with a period in the file.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -57,6 +57,6 @@ Rails.application.configure do
   # Annotate rendered view with file names.
   # config.action_view.annotate_rendered_view_with_filenames = true
 
-  # Raise error when a before_action's only/except options reference missing actions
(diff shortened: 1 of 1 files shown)


Earlopain:
These files are generated from a template in rails, I invite you to make a PR there instead. I usually only update these during a rails update since changes here will be overwritten during that and will need to be applied again. 

https://github.com/rails/rails/blob/0a9e39d14e960c8db447e21aa02f685b74da1839/railties/lib/rails/generators/rails/app/templates/config/environments/development.rb.tt#L88-L89
https://github.com/rails/rails/blob/0a9e39d14e960c8db447e21aa02f685b74da1839/railties/lib/rails/generators/rails/app/templates/config/environments/test.rb.tt#L67-L68

SeesawEffectt:
Ah, wasn't aware, will do there instead.

PR #641, opened 2024-04-30
[Flags] Deleting a post should resolve all flags.

Tarrgon:
Deleting a post right now neither resolves the flags on the post, nor does it resolve even its own deletion flag. This makes a pileup of useless unresolved flags. This will fix that problem in the future, but all deleted + flagged posts will need to have their flags updated to resolved in order to clear out the old mess.

I am unsure if this was an intentional feature, but it feels like all it does is create issues for resolving flags properly as the list becomes too hard to understand where we're at with flags

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1207,10 +1207,12 @@ class Post < ApplicationRecord
         reason = pending_flag.reason
       end
 
+      flags.each(&:resolve!)
+
       force_flag = options.fetch(:force, false)
       Post.with_timeout(30_000) do
         transaction do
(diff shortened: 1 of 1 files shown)


Sindrake:
We discussed this in Discord, but I'll re-iterate it here for posterity.

The site currently treats "resolved" flags as "rejected".
I do agree that it is confusing. However, I would argue that a proper solution to this requires a rewrite of the system to give flags distinct states: `pending`, `rejected`, and `accepted` – with the latter being treated as a deletion.

However, in the current system, accepting this would color all the correct flags red on the flags list page.
Reversing the colors would also not be very helpful, as that will instead highlight all pending flags in red, which is also not helpful.

All in all, I think it's best to look for other solutions.

PR #644, opened 2024-05-04
[AIBUR] Add the ability for admins to undo AIBURs

Tarrgon:
Still WIP, needs tests, but this adds the ability for admins to click an "Undo" button on AIBURs to undo all their changes.

I made a new sidekiq queue because for some reason on the `tags` queue, the undos would sometimes fight each other for which one got applied. So I had some cases where a post got hit by an alias and an implication in the same BUR, and when undoing it would only undo one of them, despite the alias/implication being properly removed and logs showing it did remove it from the post. My only assumption was that the tags were being accessed at the same time and it wasn't resolving all the changes? I'm not entirely sure, but making this queue fixed that, at least on small scale, for whatever reason.

I should note: for non alias/implications in BURs, the undo info is generated before the job is queued, this means that any posts that add the tags before the BUR job is executed will not be included in the undo info. I did this because I couldn't find any reasonable way to give the BUR info from the job to add to its undo info record. This isn't the same for alias/implications because they create their own undo information. I tried doing this for nukes and mass updates, but couldn't figure it out so I opted to just generate the undo info immediately, as the number of posts missed will more often then not be negligible.


Tarrgon:
I also need to ensure it fixes blacklists

Earlopain:
Thank you for this, however, I'm not going to accept it. AIBURs can already be very unexpected, and I don't want something that looks like it might help, only to make things worse.

This probably comes from the forced/vore thing and I can confidently say something like this would have made it worse. The implication for that was only renamed, the undo information was from years ago and clicking on the button to undo would have removed all the valid tags that were added on initial approve, from whatever that implication was back then.

There are just so many things that can go wrong, and for something that will be used maybe once a year its just not worth it. It's nice to have the info available to reference when someone happens but I will never do something based on that alone without double-checking that it will do the right thing. What I have done in the past is get some db exports and compare tags there since I'm very confident in their integrity.

What I will instead do is fix the bugs that lead to these situations in the first place so it doesn't happen again. There are some things that AIBURs do automatically that are unexpected and potentially disruptive. Since they themselves are used somewhat infrequently it takes time to unearth that behaviour.

Tarrgon:
> The implication for that was only renamed, the undo information was from years ago and clicking on the button to undo would have removed all the valid tags that were added on initial approve, from whatever that implication was back then.

I'm sure this could be fixed if that is the case, but if this has no chance of being accepted then I won't put any further time into it.

PR #650, opened 2024-06-06
Allow janitor+ to auto approve submitted replacements

Tarrgon:
Makes replacing as a janitor+ less of a hassle by allowing them to check a box to immediately approve the replacement on submit.

This is set to always penalize the current uploader as that's the default.


Tarrgon:
There may be better ways to pass the param to the backend, but this seemed like the easiest without messing with creation params.

Tarrgon:
Alright that should be good now

Sindrake (dismissed):

Sindrake (approved):

Sindrake (approved):

PR #652, opened 2024-06-17
Allow post approvers to search by verified uploaders

Tarrgon:
I'm willing to put in more time to optimize this, but due to the fact that (currently) there are fewer than 20 active approvers and 90k artist entries to search (which is nothing), I don't believe this is really intensive enough for the guaranteed value approvers will obtain from it.


Tarrgon:
See https://github.com/e621ng/e621ng/pull/663

Issue #383, opened 2022-03-21
Unify artist is_locked/is_active


Earlopain:

Sindrake:
Addressed in #660.

PR #660, opened 2024-06-30
[Artists] Convert is_active into proper deletion

DonovanDMC:
This pr makes the `is_active` field on artists obsolete and makes the destroy action actually delete the artist. As such I've changed the level requirement from Janitor to Admin.

I've removed `is_active` for artists in most places around the codebase (such as searching), I've left artist versions alone since that's history and should be left alone. I've also left alone it logging is_active into artist versions.

It also adds the same `(locked)` text wikipages have next to their title to artists.
image

It also adds the `ensure_can_edit` check to the `edit` route, so it doesn't bring up the edit field when the user can't actually edit the artist.
image

Keeping with the way wiki pages are, the "Edit" button remains visible regardless.


Sindrake:
To be perfectly honest here, I am sure if the "deleted" status is even something that we want to have.
I would get rid of it entirely, and add a way to permanently destroy an artist page.

DonovanDMC:
Seems like merging it with is_locked was on the agenda at some point with with issue #383, I'd be perfectly happy to just toss the column out the window

I'm not sure if we can actually drop the column, but we can for sure make it not do anything

DonovanDMC:
rubocop can kiss my ass

Sindrake:
Thank you kindly.
I don't want to approve too many things at once, so I'll get back to this one next week.

PR #663, opened 2024-07-05
[Posts] Add verified artist search

DonovanDMC:
Based on #652, but modified to use an opensearch property for better performance.

This pr adds an `artverified` metatag which is a boolean if the uploader is linked to any artist on the post.
I didn't notice any performance issues when testing locally with ~10,000 posts, so this should be usable for the public.


DonovanDMC:
I thought I already fixed rubocop but I ran with `-a` instead of `-A`, pain

Sindrake (approved):

PR #669, opened 2024-07-06
[PostSets/Pools] Handle metatag additions when set/pool is full

### LINKED ISSUE #438: Trying to add a post to a full set using a tag edit causes an unexpected error
I often add posts to sets using the `set:` metatag while editing the tags, which is a little faster than navigating the usual "Add to set" UI. 

One of my sets recently reached the limit of 10,000 posts and attempting to add a post to that set using tag edits causes the page to load for a long time, before producing an unexpected error. Adding it the normal way results in the expected `Error: Sets can have up to 10,000 posts each` so it just seems like the metatag way just isn't expecting the error.

`2dd35052-97ba-4023-8fc5-16f379ecd994` is a log ID of one of the errors if that's helpful in any way.

DonovanDMC:
This pr makes attempting to add to a set/pool when it's full gracefully fail rather than generating a 10,000+ line stacktrace.

Fixes #438


Sindrake:
I am overjoyed by the idea of making a pool/set big enough to test this.

PR #670, opened 2024-07-06
[UserFeedbacks] Add ability to soft delete feedbacks

### LINKED ISSUE #661: Soft delete for Records
I like to see the records of people so I can get an impression what sort of person they are.
When records are deleted, they get turned into staff notes and those are terribly far away from the record screen 
and thats really inconvenient.

Records should be soft deleted (hidden) at which point they are only visibile to staff and have the usual red background.
The staff note conversion would be removed.

DonovanDMC:
This pr adds the ability to soft delete feedback, allowing them to still exist while being hidden from view. Destroying has been kept as a separate action and still creates a staff note.

When viewing a user's records, if any exist that were deleted, a "View All" link will appear.
image

Only Moderator+ can view these soft deleted records. Nothing is changed about this besides adding a deleted indicator to the left on the index action, and adding a status field under date on the show action.

Resolves #661


Sindrake:
Okay, this is nifty. Although, I envisioned a different approach.
What do you think about simply making an "Deleted" feedback category – one that is invisible to regular users – and just put the deleted feedbacks in there?

The obvious downside is that the original category would be lost. But that's something we can live with, I think.

Sindrake:
Also, design-wise, I kinda hate how there is this massive offset on the left side of the table when the deleted records are hidden.
Screenshot 2024-07-11 082051
It's still not great even when those _are_ shown. Just looks awkward.
Screenshot 2024-07-11 082156
Any chance you can put the "(deleted)" indicator somewhere else?

DonovanDMC:
Losing the original category isn't amazing in my opinion, but that also isn't important information we need to hold on to
And if it were a different category entirely, we _could_ color it differently and forgo the text indicator

Sindrake (dismissed):
Looks good to me, just a couple of minor nitpicks from the Discord discussion.
- [x] The "View All" link would work better at the bottom.
    It should be a button that spans the whole width, with larger font size.
 - [x] Move the `(deleted)` label somewhere else, possibly under the date.
 - [x] Consider changing the feedback color on deleted posts, to stand out more.

Sindrake (approved):

PR #672, opened 2024-07-10
[Posts] Overhaul destroyed posts

DonovanDMC:
This pr:
* Adds a ui to posts: `/admin/destroyed_posts`, currently only visible to Admins
   image
* Adds the ability to supply a reason when destroying a post
   image
* Makes reuploads create an actual ticket
   image
* Adds the ability to toggle reuploading a destroyed post creating these tickets
  * currently only togglable by bd staff users
  * creates a `StaffAuditLog` entry when toggled


DonovanDMC:
I couldn't for the life of me get creating a ticket to work in the `no_pending_duplicates` post replacment validation, so currently they're using a workaround by setting an accessor then creating the ticket after the fact
https://github.com/e621ng/e621ng/blob/f34d35caeb9f2c69ffceb3c2f8bb3fcde4d4015d/app/models/post_replacement.rb#L55-L59
https://github.com/e621ng/e621ng/blob/f34d35caeb9f2c69ffceb3c2f8bb3fcde4d4015d/app/models/post_replacement.rb#L36-L41
https://github.com/e621ng/e621ng/blob/f34d35caeb9f2c69ffceb3c2f8bb3fcde4d4015d/app/controllers/post_replacements_controller.rb#L26

I don't like it, but I already spent a day banging my head against a wall trying to fix it other ways

Sindrake:
Could you add a better error message for when a user tries to upload / replace a destroyed post?
This nebulous message likely causes them to think that this is a bug, and try again.
Which would in turn spam us with tickets.

Screenshot 2024-07-11 151335

Maybe something like "This image had been deleted from the site, and cannot be re-uploaded."

Sindrake:
The link in the DMail does not work.
Screenshot 2024-07-11 152547

Sindrake:
Would it be possible to have a different ticket message for replacements?
Screenshot 2024-07-11 153928

This just looks like the report is about a normal upload.
I would also like to see which post they were trying to replace.

Additionally, that is a link to e621 in that ticket, for some reason.
Doesn't work if the domain is different (on e926, e6ai, or locally).
Maybe, you could make that a relative link instead? Like:
```
"destroyed post #69":/admin/destroyed_posts/69
```

db/structure.sql
```diff
@@ -4493,266 +4495,266 @@ ALTER TABLE ONLY public.favorites
 SET search_path TO "$user", public;
 
 INSERT INTO "schema_migrations" (version) VALUES
-('20100204211522'),
```
Sindrake: Please, un-fuck this =/
It should not matter, but it annoys me.

app/javascript/src/javascripts/posts.js
```diff
@@ -354,9 +354,9 @@ Post.initialize_links = function() {
   });
   $("#destroy-post-link").on('click', e => {
     e.preventDefault();
-    if(!confirm("This will permanently delete this post (meaning the file will be deleted). Are you sure you want to delete this post?"))
-      return;
-    Post.destroy($(e.target).data('pid'));
+    const reason = prompt("This will permanently delete this post (meaning the file will be deleted). What is the reason for destroying the post?")
+    if(reason === null) return;
```
Sindrake, line 358: Reason would never be null here.
If the user leaves the prompt empty, it'll simply return an empty string.

DonovanDMC: Canceling returns null, if we don't check for null it's impossible to cancel
This is set up where an empty reason is valid

DonovanDMC:
> Could you add a better error message for when a user tries to upload / replace a destroyed post? This nebulous message likely causes them to think that this is a bug, and try again. Which would in turn spam us with tickets.
> 
> Screenshot 2024-07-11 151335
> 
> Maybe something like "This image had been deleted from the site, and cannot be re-uploaded."

I left the error vague on purpose due to it having previously been vague, it was definitely vague for a reason (likely to give out as little information as possible)

> Which would in turn spam us with tickets.

Which is a great reason to disable notifications for that destroyed post, it wouldn't matter how many times they try to reupload it
It could also be changed so that only one ticket is made per combination of user & post within a timeframe of say, 5 minutes

I don't really have a horse in the race on what the error is, so yeah I'll change it

Sindrake:
> I don't really have a horse in the race on what the error is, so yeah I'll change it

I asked NMNY already, and he said that he's okay with making the error more straightforward.

Also, please stop force-pushing stuff, and make commits instead.
I'll squash them into a single commit when I merge the PR anyways. Force-pushing causes some wacky merge conflicts for me, and also makes it harder to review changes you made after the review.

app/javascript/src/javascripts/posts.js
Sindrake: Right, okay.

DonovanDMC:
> Also, please stop force-pushing stuff, and make commits instead. I'll squash them into a single commit when I merge the PR anyways. Force-pushing causes some wacky merge conflicts for me, and also makes it harder to review changes you made after the review.

I'm happy to do individual commits rather than force pushing, I prefer that in fact
I'm used to earlopain almost never squashing

Sindrake:
Tested it all again, looks good to me.
We'll see how it works with the real deal.

Thanks for your hard work.

PR #683, opened 2024-07-26
[WikiPages] Implement basic redirects

Sindrake:
This PR allows one wiki page to seamlessly redirect users to another wiki page.
That is necessary to avoid awkward pages like [this](https://e621.net/wiki_pages/6) – users can just be shown what they were looking for instead.

redirect

Users are not technically redirected between pages, as the URL remains the same.
Instead, the contents of the parent wiki page get displayed instead of the original one.

redirect2

As far as permissions are concerned, I think it's okay to let Privileged+ to handle this.
Opening it up to the public is fraught with peril, and only letting staff members handle this seems unnecessary. Privileged users, on the other hand, are both trustworthy enough to be allowed to set up redirects, and also more likely to be creating wiki pages in the first place.

### DIFF
--- a/app/models/wiki_page.rb
+++ b/app/models/wiki_page.rb
@@ -10,7 +10,7 @@ class WikiPage < ApplicationRecord
   validates :title, uniqueness: { :case_sensitive => false }
   validates :title, presence: true
   validates :title, tag_name: true, if: :title_changed?
-  validates :body, presence: { unless: -> { is_deleted? || other_names.present? } }
+  validates :body, presence: { unless: -> { is_deleted? || other_names.present? || parent.present? } }
   validates :title, length: { minimum: 1, maximum: 100 }
   validates :body, length: { maximum: Danbooru.config.wiki_page_max_size }
   validate :user_not_limited
--- a/app/models/wiki_page.rb
+++ b/app/models/wiki_page.rb
@@ -95,9 +95,7 @@ class WikiPage < ApplicationRecord
         q = q.where("is_deleted = false")
       end
 
-      if params[:parent].present?
-        q = q.where("parent LIKE ? ESCAPE E'\\\\'", params[:parent].downcase.strip.tr(" ", "_").to_escaped_for_sql_like)
-      end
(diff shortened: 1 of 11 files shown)


app/models/wiki_page.rb
```diff
@@ -94,6 +95,10 @@ def search(params)
         q = q.where("is_deleted = false")
       end
 
+      if params[:parent].present?
+        q = q.where("parent LIKE ? ESCAPE E'\\\\'", params[:parent].downcase.strip.tr(" ", "_").to_escaped_for_sql_like)
+      end
```
DonovanDMC:
```suggestion
      q = q.attribute_matches(:parent, params[:parent]&.tr(" ", "_"))
```

The `attribute_matches` does the same thing internally for text columns, additionally it will use the tsvector when wildcards aren't present
https://github.com/e621ng/e621ng/blob/8edb1b9226d63aeb405aedd596af2956317d3ae1/app/models/application_record.rb#L104-L108

It also has a check for nil (empty parameters end up nil, so this is functionally the same as a presence check)
https://github.com/e621ng/e621ng/blob/8edb1b9226d63aeb405aedd596af2956317d3ae1/app/models/application_record.rb#L35-L36

DonovanDMC:
Oh yeah, forgot to mention: help pages
There should probably either be a check to make sure help pages either can't use redirected wikipages (and wikipages for help pages can't be redirected) or help pages should follow those redirects

DonovanDMC (changes requested):
see above as well as discord

Sindrake:
> There should probably either be a check to make sure help pages either can't use redirected pages (and wiki pages for help pages can't be redirected) or help pages should follow those redirects

I feel like it should be the former.

DonovanDMC:
> > There should probably either be a check to make sure help pages either can't use redirected pages (and wiki pages for help pages can't be redirected) or help pages should follow those redirects
> 
> I feel like it should be the former.

Not allowing redirects makes more sense in my book, help pages can use _any_ wikipage so there's no need to allow their wikipages to be redirected

PR #696, opened 2024-08-01
Allow janitors to bypass dmail limits

Tarrgon:
Janitors are basically detectives that find artist user accounts. Some choose to dmail these users to ask them to verify. This limit makes that hard.


faucetlol:
Instead of adding a new line, you should be fine to just replace `is_moderator?` with `is_janitor?` since it will return true for anybody ranked higher than janitors - moderators included.

Sindrake (approved):

PR #697, opened 2024-08-03
[Tickets] Add artist/tag tickets

DonovanDMC:
This pr adds the ability to report both artists and tags, as well as a polymorphic `model` relation to tickets. Two separate migrations are included, one to add the new columns and one to remove the old columns. Obviously there's a required order: first migration, fixer, second migration. To prevent rails from yelling about migrations when a page is loaded in development, change this:
https://github.com/e621ng/e621ng/blob/fc7d84affdb2ff36b88f37588f4122c7f864e117/config/environments/development.rb#L51 to `false`.

A pr for the bot will follow suit.


Sindrake:
This should be in two separate PRs.

Sindrake:
Also, this will definitely cause an issue with deployment.

Remember, releases normally hit e926 first, then e621 a week later.
This PR will break all tickets on e621 for a week.

DonovanDMC:
I've forced pushed out the polymorphic changes, I'll recreate that once this is merged

PR #698, opened 2024-08-03
[Tags] Alias antecedent ignore

### LINKED ISSUE #329: Antecedent name restriction blocking valid alias
fucky_wucky 1
The BUR error messages are blocking aliases of invalid tags to valid tags.

Tarrgon:
Fixes #329.

Invalid tags that already exist should be aliased away to the valid tags where possible, however the tag validator prevents users from making alias requests with antecedent names that it considers invalid.

This PR allows antecedent names on aliases to skip most of the validation sequence, except for checking for blanks, and checking for tags starting with `-`, which would break the tag string diff.

This will also allow us to create aliases for non-english artists to the english representation of their name, which could prevent people not tagging them, or tagging them improperly as a helpful side effect.


app/models/tag_implication.rb
DontTalkToMeThx, line 1: Debating if we should even have this, but as it currently stands, implications disabled ascii check, so moving it here is to maintain that feature.

app/models/tag_implication.rb
```diff
@@ -5,6 +5,7 @@ class TagImplication < TagRelationship
 
   array_attribute :descendant_names
 
+  validates :antecedent_name, tag_name: { disable_ascii_check: true }, if: :antecedent_name_changed?
```
Tarrgon, line 9: Debating if we should even have this, but as it currently stands, implications disabled ascii check, so moving it here is to maintain that feature.

Tarrgon:
Now that I think about it, if we allow zero-width spaces in alias antecedents, we should highlight them in the forum post/alias view for admins.

Sindrake (changes requested):
I am not sure if I agree with how this is implemented.

I would reduce the number of rules that can be bypassed like this to the absolute minimum.
Just the ones that are _currently_ causing issues or are waiting to be aliased – like the colon `:` character from the original issue, the peso `$` sign, backslash `\`, and non-acii characters. 

If more issues come up in the future, we can add a bypass for those later.

Tarrgon:
recheck

Sindrake (approved):

Sindrake (approved):

PR #699, opened 2024-08-03
[UserInput] Add newline normalization

### LINKED ISSUE #363: Counting issues with `\n` in dtext box
See https://e621.net/forum_topics/31929

The textarea seems to only contain `\n`, but they are converted to `\r\n` before submitting, thus breaking validation when the text contains a few linebreaks and is close to the limit.

Tarrgon:
Adds newline normalization for user input areas that have a maximum length.

Fixes #363


Tarrgon:
There may be better placement for these lines, I tried to go with above the group of validations it resides in, though

PR #705, opened 2024-08-04
[WikiPages] Redirect back to wiki when editing/fixing tag

DonovanDMC:
This pr adds a simple hidden field, filled in from the `Referer` header which if the header contains `wiki_pages` will redirect to the wiki page rather than the tag.

### DIFF
--- a/app/controllers/tag_corrections_controller.rb
+++ b/app/controllers/tag_corrections_controller.rb
@@ -20,8 +20,8 @@ class TagCorrectionsController < ApplicationController
 
     if params[:commit] == "Fix"
       @correction.fix!
-      if params[:from_wiki].to_s.truthy? && @correction.tag.wiki_page.present?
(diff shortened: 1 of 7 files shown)


Sindrake (changes requested):
I don't really like how this works.

To me, there is pretty much no reason to ever be on the tag's show page. There isn't any information there that can't be accessed from the wiki or the tag index. Thus, redirecting the user there after the tag is edited does not make sense in pretty much all circumstances.

I would prefer if the redirects worked like this:
1. If the user arrived at the tag edit page from the wiki, they should be redirected back to the wiki, even if the tag does not actually have an actual wiki page. Just send them to `show_or_new`, it'll be fine.
2. In all other circumstances, the user should just get redirected to the tag search for that specific tag. The same way redirect currently works for the "fix" action.

Does that make sense to you?

Sindrake (approved):

PR #709, opened 2024-08-06
[PostReplacements] Prevent approving if post has notes and is note locked

DonovanDMC:
This pr makes approving a replacement skip attempting to resize notes if the post is note locked. As was reported [in Discord](https://discord.com/channels/431908090883997698/432280999293091872/1270320547825455105), this throws an error otherwise.

### DIFF
--- a/app/javascript/src/javascripts/post_replacement.js
+++ b/app/javascript/src/javascripts/post_replacement.js
@@ -35,7 +35,7 @@ PostReplacement.approve = function (id, penalize_current_uploader) {
   }).done(function () {
     set_status($row, "approved");
   }).fail(function (data) {
-    Utility.error(data.responseText);
+    Utility.error(`Failed to approve replacement: ${data.responseJSON.message}`);
     set_status($row, "replacement failed");
   });
 };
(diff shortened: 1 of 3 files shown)


Tarrgon:
Hm. I wonder if the lock should be ignored while replacing. As the lock is mainly just for regular users since janitor+ can unlock them. And janitor+ are the ones accepting replacements (in most cases... anyways).

DonovanDMC:
> Hm. I wonder if the lock should be ignored while replacing. As the lock is mainly just for regular users since janitor+ can unlock them. And janitor+ are the ones accepting replacements (in most cases... anyways).

The previous sentiment I've gotten when attempting to add bypassing to locks is that we apparently don't want that
In reality bypassing _just_ for replacing will be messy, and letting janitor+ bypass it entirely goes back to the previous sentence

faucetlol:
Just simply skipping the resizing seems like bad practice to me in this case, it's definitely going to cause problems when notes end up being _locked_ into the wrong place on the image, requiring staff intervention to fix it, which I assume would have to be done manually because the rescale method was already skipped.

Additionally, what would happen if an image was replaced with a lower resolution image (not uncommon with jpg->png replacements) and then the existing notes end up outside the boundary of the new image size?

I do agree that bypassing checks that are there for a purpose is also bad practice, but probably the least harmful option of the two.

DonovanDMC:
The approval could just be prevented with a proper error, as a compromise to skipping (which has issues as mentioned) and bypassing (which is messy)

PR #713, opened 2024-08-08
[AvoidPosting] Add caching, order search, and tweak modactions

DonovanDMC:
This pr adds caching to the static list, adds an order search option, and does some minor tweaking to the modaction text format (linking to the avoid posting entry)

### DIFF
--- a/app/decorators/mod_action_decorator.rb
+++ b/app/decorators/mod_action_decorator.rb
@@ -67,15 +67,15 @@ class ModActionDecorator < ApplicationDecorator
 
       ### Avoid Posting ###
     when "avoid_posting_create"
-      "Created \"avoid posting entry\":/avoid_postings/#{vals['id']} for [[#{vals['artist_name']}]]"
+      "Created \"avoid posting ##{vals['id']}\":/avoid_postings/#{vals['id']} for [[#{vals['artist_name']}]]"
     when "avoid_posting_update"
-      "Updated \"avoid posting entry\":/avoid_postings/#{vals['id']} for [[#{vals['artist_name']}]]"
+      "Updated \"avoid posting ##{vals['id']}\":/avoid_postings/#{vals['id']} for [[#{vals['artist_name']}]]"
     when "avoid_posting_destroy"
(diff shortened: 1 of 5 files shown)


app/decorators/mod_action_decorator.rb
```diff
@@ -67,15 +67,15 @@ def format_description
 
       ### Avoid Posting ###
     when "avoid_posting_create"
-      "Created avoid posting entry for artist \"#{vals['artist_name']}\":/artists/show_or_new?name=#{vals['artist_name']}"
+      "Created \"avoid posting entry\":/avoid_postings/#{vals['id']} for [[#{vals['artist_name']}]]"
```
Sindrake: I think it would be preferable to have the link list the avoid posting entry number, to match the format of other things.
So, ex. `\"avoid posting #{vals['id']}\":/avoid_postings/#{vals['id']}`

app/views/avoid_postings/_search.html.erb
```diff
@@ -13,4 +13,5 @@
     <%= f.input :staff_notes %>
   <% end %>
   <%= f.input :is_active, label: "Active?", collection: [%w[Yes true], %w[No false], %w[Any any]] %>
+  <%= f.input :order, collection: [["Artist Name", "artist_name"], %w[Created created_at], %w[Updated updated_at]] %>
```
Sindrake: Set a default value, or add a blank one to return to ordering by ID.

app/models/avoid_posting.rb
```diff
@@ -96,7 +97,19 @@ def search(params)
       q = q.attribute_matches(:staff_notes, params[:staff_notes])
       q = q.where_user(:creator_id, :creator, params)
       q = q.where("creator_ip_addr <<= ?", params[:ip_addr]) if params[:ip_addr].present?
-      q.apply_basic_order(params)
+      case params[:order]
+      when "artist_name", "artist_name_asc", nil
```
Sindrake: Why is the default order alphabetical?
That's not the case anywhere else on the site.

Sindrake (changes requested):

Sindrake (approved):

PR #718, opened 2024-08-11
[Flags] Flip inferior flag

Tarrgon:
Lets staff move an "inferior" type flag to the parent.

See https://github.com/orgs/e621ng/projects/1/views/1?pane=issue&itemId=71429263

### DIFF
--- a/app/javascript/src/javascripts/posts.js
+++ b/app/javascript/src/javascripts/posts.js
@@ -59,7 +59,7 @@ Post.initialize_moderation = function () {
   });
   $(".move-flag-to-parent-link").on("click", e => {
     e.preventDefault();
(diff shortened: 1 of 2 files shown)


Tarrgon:
Draft since I've yet to test other buttons and make sureit doesn't break anything

Tarrgon:
There is another eslint issue completely unrelated to this PR:
https://github.com/e621ng/e621ng/blob/master/app/javascript/src/javascripts/utility/storage_util.js#L16

Fixing it here feels out of scope, ~~but the eslint test will fail.~~ Ok idk why it worked lmao

Sindrake (approved):

PR #723, opened 2024-08-14
[PostView] Show verified uploaders to everyone

Tarrgon:
Implements https://github.com/orgs/e621ng/projects/1/views/1?filterQuery=-status%3ADone&pane=issue&itemId=73408921

Not sure if this is actually something that's wanted, but the code existing should make it easier to decide.

Examples (taken while logged out):
Of a post with an uploader linked to one of the artist tags:
r6exue_19464 1

Same uploader uploading content to a tag they are not verified for:
11z5vv_19465 1

And for completion, here's the above post while logged into a janitor+ account:
0rsy1i_19466 1


Sindrake (approved):

DonovanDMC:
This now shows the record number section to anyone if the artist is verified, which feels unwanted
image

While that information isn't private, it also isn't relevant to the average user

PR #726, opened 2024-08-14
Add a fully featured blocking system

DonovanDMC:
This pr adds a fully featured blocking system.

Hiding blips, comments, forum topics/posts, and blocking dmails can each be selected separately.
image
image

Each of these has a link to toggle blocking. This toggle is persistent in the current browser.
image

Admins can view the blocks of anyone, but they cannot edit them or block people on their behalf.
Staff members cannot disable messages, but they can use any other option.
image

### DIFF
--- a/app/controllers/user_blocks_controller.rb
+++ b/app/controllers/user_blocks_controller.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+class UserBlocksController < ApplicationController
+  before_action :member_only
+  before_action :load_user
+  before_action :check_privilege, except: %i[index]
+  respond_to :html, :json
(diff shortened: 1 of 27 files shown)


Tarrgon:
What happens if a user blocks messages and then eventually becomes staff? Does it remove their existing message blocks?

DonovanDMC:
Nothing is changed, just the same that the setting for disabling dmails isn't changed
The setting & blocks aren't checked at all for janitor users, it exits out beforehand
https://github.com/e621ng/e621ng/blob/aa55893e28a35504307b6a19a672d2346876c449/app/models/dmail.rb#L157-L163

PR #743, opened 2024-09-02
Add deleted record count to profiles

DonovanDMC:
This pr both adds a deleted count to the record numbers on profiles, and removes the words, leaving just the numbers. The record meta on posts does not include deleted records.
image
image


Sindrake (approved):
I am 58% sure that people will complain about the shortened feedback summary, but we'll get there when we get there.

PR #749, opened 2024-09-05
[Posts] Add back comment disabling

DonovanDMC:
This pr adds back the previous comment disabling feature, to work alongside locking comments, for when we want to completely torch a comment section for good. It also allows moderators to lock comments, as well as making the post events for both only not searchable, rather than hidden entirely.

To fix issues we had previously (disabled comments being visible in /comments), a join is used to only pick visible comments rather than only code checks.

https://discord.com/channels/431908090883997698/1099781331623358688/1280935337815965778


app/views/comments/partials/index/_list.html.erb
```diff
@@ -1,28 +1,40 @@
 <div class="comments-for-post" data-post-id="<%= post.id %>">
-  <% if post.is_comment_locked? %>
-    Comment section has been locked.
-  <% end %>
-  <div class="row notices">
-    <% if post.comments.hidden(CurrentUser.user).count > 0 || (params[:controller] == "comments" && post.comments.count > Comment::RECENT_COUNT) %>
+  <% if !CurrentUser.user.is_moderator? && post.is_comment_disabled? %>
```
Sindrake: This whole section seems messy, and it has some redundancies.
Are you sure that you can't restructure it to avoid those?

app/models/comment.rb
```diff
@@ -47,13 +47,16 @@ def hidden(user)
...
     def visible(user)
-      if user.is_moderator?
-        where("comments.score >= ? or comments.is_sticky = true", user.comment_threshold)
-      elsif user.is_janitor?
-        where("(comments.score >= ? or comments.is_sticky = true) and (comments.is_sticky = true or comments.is_hidden = false or comments.creator_id = ?)", user.comment_threshold, user.id)
-      else
-        where("(comments.score >= ? or comments.is_sticky = true) and (comments.is_hidden = false or comments.creator_id = ?)", user.comment_threshold, user.id)
+      q = where("comments.score >= ? or comments.is_sticky = true", user.comment_threshold)
+      unless user.is_moderator?
+        q = q.joins(:post).where("comments.is_sticky = true or posts.is_comment_disabled = false")
```
Sindrake: I would argue that the users should still see their own comments even if the post has the comment section disabled.
As it is, this may create a strange situation where the user may have the profile page may say that they have created comments, but clicking on the comment count may show no results.

Not showing users their own comments on the post page is fine with me, though.

app/views/posts/partials/show/_edit.html.erb
```diff
@@ -72,6 +74,7 @@
       <fieldset class="limits">
         <%= f.input :hide_from_anonymous, as: :boolean, label: "Hide from Anon" %>
         <%= f.input :hide_from_search_engines, as: :boolean, label: "Hide from search engines" %>
+        <%= f.input :is_comment_disabled, label: "Comments" %>
```
Sindrake: I would prefer to have this labeled as "Hide comments", for consistency.

Sindrake (changes requested):

app/views/comments/partials/index/_list.html.erb
DonovanDMC: It's a bit difficult to do how we need it without being redundant, since we need two different ways to display it:
* for normal posts and moderators, as normal
* for disabled comments, we need *only* the comments, no empty notice, and no "Show All" link

While these could be intertwined I think it'd just be messier with the amount of if statements needed

Sindrake (approved):

PR #760, opened 2024-09-18
[DangerZone] Allow admins to hide pending posts for a specified amount of hours

Tarrgon:
Implements https://github.com/orgs/e621ng/projects/2/views/1?pane=issue&itemId=18707651, but as hours instead of days, since that seemed more useful.

A security tool for mass post raids. It will hide pending posts for the amount of hours specified, afterwards they will be visible again.

Uses an opensearch query when active, so this does not reduce pages to a few entries per page, since they aren't returned in the query in the first place.

Made a small video to show what it does. There are cuts to reduce the video duration because local load speeds are awful for me and because github has a 10mb limit:

https://github.com/user-attachments/assets/ffa2d11e-1d43-444c-968d-f391d182e55b

### DIFF
--- a/app/logical/danger_zone.rb
+++ b/app/logical/danger_zone.rb
@@ -26,7 +26,7 @@ module DangerZone
   def self.hide_pending_posts_for
     Cache.redis.get("hide_pending_posts_for").to_f || 0
   rescue Redis::CannotConnectError
-    0
+    PostPruner::DELETION_WINDOW * 24
   end
 
   def self.hide_pending_posts_for=(duration)
--- a/app/logical/danger_zone.rb
+++ b/app/logical/danger_zone.rb
@@ -10,7 +10,7 @@ module DangerZone
       return true
     end
 
-    user.is_staff? || !post.is_pending? || post.created_at.before?(hide_pending_posts_for.hours.ago)
+    post.uploader_id == user.id || user.is_staff? || !post.is_pending? || post.created_at.before?(hide_pending_posts_for.hours.ago)
   end
 
   def self.min_upload_level
--- a/app/logical/danger_zone.rb
+++ b/app/logical/danger_zone.rb
@@ -10,7 +10,7 @@ module DangerZone
       return true
     end
 
-    (!user.nil? && user.is_approver?) || !post.is_pending? || post.created_at.before?(hide_pending_posts_for.hours.ago)
+    user.is_staff? || !post.is_pending? || post.created_at.before?(hide_pending_posts_for.hours.ago)
   end
 
   def self.min_upload_level
--- a/app/logical/danger_zone.rb
+++ b/app/logical/danger_zone.rb
@@ -5,6 +5,14 @@ module DangerZone
     user.level < min_upload_level
   end
 
+  def self.post_visible?(post, user)
+    if hide_pending_posts_for <= 0
+      return true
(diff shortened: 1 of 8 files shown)


Tarrgon:
I'll look into these tests tomorrow, not sure how it's causing it to fail considering it shouldn't even be active in any of the tests outside its own

app/logical/danger_zone.rb
```diff
@@ -5,6 +5,14 @@ def self.uploads_disabled?(user)
     user.level < min_upload_level
   end
 
+  def self.post_visible?(post, user)
+    if hide_pending_posts_for <= 0
+      return true
+    end
+
+    (!user.nil? && user.is_approver?) || !post.is_pending? || post.created_at.before?(hide_pending_posts_for.hours.ago)
```
DonovanDMC: all staff should bypass this, not just approvers
we specifically have an `is_staff?` function that can be used

Also in normal usage `CurrentUser.user` *should* never be nil, the only places it can possibly be nil are in the console or tests

https://github.com/e621ng/e621ng/blob/2ba1173d36326ef902790b12baea3241227ad732/app/controllers/application_controller.rb#L9
https://github.com/e621ng/e621ng/blob/2ba1173d36326ef902790b12baea3241227ad732/app/controllers/application_controller.rb#L166-L168
https://github.com/e621ng/e621ng/blob/2ba1173d36326ef902790b12baea3241227ad732/app/logical/session_loader.rb#L16-L17

The anonymous user is a full user model, so levels can be checked against it just fine
https://github.com/e621ng/e621ng/blob/2ba1173d36326ef902790b12baea3241227ad732/app/models/user.rb#L270-L275

app/logical/elastic_post_query_builder.rb
```diff
@@ -314,5 +314,27 @@ def build
     else
       order.push({id: :desc})
     end
+
+    if !CurrentUser.is_approver? && DangerZone.hide_pending_posts_for > 0
```
DonovanDMC: Same as above

DonovanDMC (changes requested):

DonovanDMC:
> I'll look into these tests tomorrow, not sure how it's causing it to fail considering it shouldn't even be active in any of the tests outside its own

Further tests are failing because you aren't resetting the value - redis isn't mocked, any changes *do* actually happen and will stick around
You need to add to the teardown function
https://github.com/e621ng/e621ng/blob/96f1e057d5a577b05d23d2a0c4e04e79bad0c6ab/test/functional/admin/danger_zone_controller_test.rb#L11-L13

app/logical/danger_zone.rb
Tarrgon: > Also in normal usage CurrentUser should never be nil, the only places it can possibly be nil are in the console or tests

~~Yeah it was my mistake, I switched away from it in a leter commit~~ Misread this

Tarrgon: > we specifically have an `is_staff?` function that can be used

I'll add it, but since not all approvers are staff because the permission can be granted separately, it'll need both

Tarrgon:
There is an issue of users not being able to see their _own_ pending posts, now that I think about it. Might have to think of a way to work around that to show them...

Tarrgon:
There we go

PR #763, opened 2024-09-21
Add support for folding displays

Psychpsyo:
## Summary
This adds support for folding displays in responsive mode by ensuring that there's an even number of columns (4) so that the search results do not get laid out onto the fold.

image

## Technical Details
This is done by having the results use `display: grid` (instead of `flex`) to be able to control the number of rows.
Since `31vw` can now not be used as a constant for how wide a result should be, this is now `100cqw` (100% of container with), with the post having been declared a container in the inline direction.
The four-column layout gets toggled with a media query for `horizontal-viewport-segments`, which is available by default in Samsung Internet and some versions of Edge, as well as behind an experimental feature flag in Chrome.

### DIFF
--- a/app/javascript/src/styles/common/z_responsive.scss
+++ b/app/javascript/src/styles/common/z_responsive.scss
@@ -199,8 +199,9 @@
 
     #posts #posts-container {
       width: 100%;
-      display: flex;
-      flex-wrap: wrap;
+      display: grid;
+      grid-template-columns: repeat(3, 1fr);
+      gap: 1vw;
       align-items: center;
       justify-content: space-evenly;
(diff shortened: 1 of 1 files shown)


Psychpsyo:

Psychpsyo:
Is there anything I need to do as an assignee or is this just for organization?

Psychpsyo:
@Sindrake Is there anything I need to do as an assignee or is this just for organization?

Sindrake:
It's for organization.
Mainly to have your icon show up next to the PR in the spreadsheet, so that I can tell at a glance who did what.

PR #769, opened 2024-10-22
[Posts] Fix pool metatags with capital letters

DonovanDMC:
A regression that happened in b3b16759c4901cbe3df7803b1afd85f9e7af5463

Fixes the pool metatags (`newpool`, `pool`, `-pool`) with capital letters (such as `A_Pool`)
Previously the pool would be created successfully but the post would not be added to the pool

This also removes the downcasing from `tag_string_diff`, which prevented using capital letters in the pool metatags (the ui uses tag_string & old_tag_string which doesn't do this, so this only affects api usage)

Additionally the default text when using newpool has been replaced with an empty string (if we leave it blank the description will be null due to the column being nullable)

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -696,7 +696,7 @@ class Post < ApplicationRecord
         when /^newpool:(.+)$/i
           pool = Pool.find_by_name($1)
           if pool.nil?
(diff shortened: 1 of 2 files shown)


DonovanDMC:
yeah that's what I get for not running rubocop before pushing

PR #783, opened 2024-11-06
[Autocomplete] Remove mp4 from file type completion

faucetlol:
image

e621 does not have mp4 posts, nor does it intend to, so having this here is nothing but confusing.


faucetlol:

PR #785, opened 2024-11-09
Adding Apple Silicon Mac support in Docker Container

eiffelwong1:
## Issue

image

launching docker on Apple silicon mac would fail due to nokogiri library not found.

## Solution

by adding gcompat, a glibc compatibility layer to the docker image, the docker can be launched and webpage is up for apple silicon macs

### source
https://stackoverflow.com/questions/70963924/unable-to-load-nokogiri-in-docker-container-on-m1-mac/71014474#71014474

## testing

rebuild the docker and relaunch the server
`docker compose down -v`
`docker compose build --no-cache`
`docker compose run --rm e621 /app/bin/setup`
`docker compose up`
`docker exec -it e621ng-e621-1 /app/bin/populate`
- then access localhost:3000 and verify basic functionality 

machine tested on: 
Chip: M2 Pro
MacOS: 14.6.1 (23G93)

## Other
Special thanks to @DonovanDMC for finding the StackOverflow solution

Issue #782, opened 2024-11-06
[Mobile View] Parent / Children image is too small


eiffelwong1:
<img width="483" alt="image" src="https://github.com/user-attachments/assets/88c28222-da1b-4b39-9e00-d619e1aa45b5">

mentioned by user SpuriousZabaione

example: https://e621.net/posts/5163793?q=isparent%3Atrue

PR #786, opened 2024-11-09
Fix parent children preview too small in mobile view

eiffelwong1:
# Issue: 
#782 

# Fix
Edited the CSS to reply on flex, flex-wrap to achieve better responsiveness in smaller screen
Docker file edits fixes Apple silicon launch issue see: https://github.com/e621ng/e621ng/pull/785

# Demo

https://github.com/user-attachments/assets/dff8e3ac-07d1-4fbb-b4b5-8e0e98543ff2


Sindrake:
I think I would also want to stack the elements in the `div.parent-children`  vertically below a certain resolution.
But that's nitpicking.

PR #788, opened 2024-11-12
[Comments] Show direct link comments to the creator regardless of settings

Tarrgon:
See https://discord.com/channels/431908090883997698/1099361575116222656/1305898089848242186

Just makes the comment show page always show the comment to the creator of the comment. Even if they have hidden comments disabled or the score is below their threshold.


Sindrake:
I agree, but why put this in the view and not in the model?
Could just add it to the `should_see` method.

Tarrgon:
Does should_see have access to `current_page?` I didn't want to make users always see their own hidden comments, as if I made the edit in should see, as far as I know it would make them see non-direct link comments.

PR #798, opened 2024-11-19
Automatically like posts when they are favorited, backend

eiffelwong1:
# Issue

https://github.com/orgs/e621ng/projects/1/views/7?pane=issue&itemId=69975628

# What's done

adding upvote action to the backend in favorite action.

upvote doesn't shown right after
view is correct after a refresh

# Demo

## works when haven't upvote yet
image

note: the upvote button didn't go green (view is correct after a refresh)

image

## favoriting works if post were upvoted before (no error)
image

### DIFF
--- a/app/javascript/src/javascripts/favorites.js
+++ b/app/javascript/src/javascripts/favorites.js
@@ -41,7 +41,7 @@ Favorite.create = function (post_id) {
     }).done(function () {
       Post.notice_update("dec");
       Favorite.after_action(post_id, 1);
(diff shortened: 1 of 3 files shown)


DonovanDMC (changes requested):
Upvoting really should be kept distinct from the favorite manager module, ideally that should be done in the controller
It also should be possible to favorite without trying to upvote (e.g. a query param), and you also need to handle the possible errors from voting so they can't prevent favoriting

PR #799, opened 2024-11-23
[UserFeedbacks] Add ban link

clragon:
Adds a Ban link to the user feedback page, for easily creating a ban when necessary.


DonovanDMC:
You can use `.dig(:search, :user_id)` so you don't need to check both individually

Sindrake (approved):

PR #808, opened 2024-11-26
Automatically like posts when they are favorited, backend (2nd attempt)

eiffelwong1:
# Issue

https://github.com/orgs/e621ng/projects/1/views/7?pane=issue&itemId=69975628

1st attempt:
https://github.com/e621ng/e621ng/pull/798

# What's done

adding upvote action to the backend in favorite action.

upvote doesn't shown right after
view is correct after a refresh

note: removing favourite doesn't remove upvote

# Demo

## works when haven't upvote yet
image
note: the upvote button didn't go green (view is correct after a refresh)

image

## favoriting works if post were upvoted before (no error)
image

### DIFF
--- a/app/controllers/favorites_controller.rb
+++ b/app/controllers/favorites_controller.rb
@@ -28,7 +28,8 @@ class FavoritesController < ApplicationController
   def create
     @post = Post.find(params[:post_id])
     FavoriteManager.add!(user: CurrentUser.user, post: @post)
-    flash.now[:notice] = "You have favorited this post"
+    VoteManager.vote!(user: CurrentUser.user, post: @post, score: 1)
+    flash.now[:notice] = "You have favorited and upvoted this post"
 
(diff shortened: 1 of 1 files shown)


Sindrake:
Second time I'm looking into this, and second time I'm realizing what a mess upvoting and favoriting actually is.

The vote buttons should be updated together with the favorites button.
Additionally, it might be a good idea to have a checkbox somewhere that would allow users to decide whether they want to link upvotes to favorites like this or not.

clragon:
I dont understand why this would be a good idea.
If there is no meaningful difference between upvoting and favoriting, would it not be better to just remove favoriting altogether?

But we have this split and it seems sensible to either keep it, or collapse it entirely, rather than make the UI unpredictable for users.

Sindrake:
Intriguing, but I've decided against this.

PR #823, opened 2024-12-10
[StaffNotes] Add updating and deleting

clragon:
This PR adds updating and deleting staff notes. 
It slightly changes the UI.
It also moves the endpoint to /staff_notes, which felt more appropriate.

Modifying staff notes always triggers a new mod action which contains the full details of the modification.
These mod actions are only visible to mods.

staff notes cannot be destroyed currently.

image


DonovanDMC:
~~This pr looks to make destroying possible via the api but without any associated modaction~~
^ I see now that you've excluded the route, I'm of the opinion that we don't need destroying staff notes period and hiding them should be good enough, hence why my implementation never had full destroying

I'm also not a fan of the modactions being visible to people at all, they ideally shouldn't know we're creating notes in the first place rather than just hiding details, which is why I took the staff audit log (which I gave a ui that was only visible to staff) and put it there

Also if you're going to relocate the route you should similarly relocate the controller and views

clragon:
got it, I'll move the files and rip out the code for destroying all together. 

I guess I can look into just excluding these mod actions from the query for non staff users, although this seems, well, unorthodox to me, considering even IP bans are part of the publicly visible mod actions. 
to be honest I dont think any mod actions need to be visible whatsoever, but I know thats a more niche opinion.

DonovanDMC:
> considering even IP bans are part of the publicly visible mod actions

(FWIW I also moved those out of the modactions in my own code)
I also don't think those should be in the mod actions but that's an argument for another day

app/models/staff_note.rb
```diff
@@ -12,28 +37,45 @@ def search(params)
...
+    return false unless user.is_staff?
+    user.id == creator_id || user.is_admin?
   end
 
-  def unresolve!
-    self.resolved = false
-    save
+  def can_delete?(user)
+    return false unless user.is_staff?
+    return true if creator_id == user.id || user.is_admin
```
Sindrake: This should be `user.is_admin?`

app/views/staff_notes/_search.html.erb
```diff
@@ -1,6 +1,8 @@
-<%= form_search path: admin_staff_notes_path, always_display: true do |f| %>
+<%= form_search path: staff_notes_path do |f| %>
   <%= f.user :creator %>
+  <%= f.user :updater %>
```
Sindrake, line 3: This does not work, since `updater_name` is not a permitted search param.

Sindrake (changes requested):

Sindrake (approved):

PR #837, opened 2025-01-13
[Comments] Fix comments show page bug

Tarrgon:
Fixes a bug that arises when a hidden comment would be shown in the `/comments` page listing while not being a direct link to that comment.


Tarrgon:
Perhaps we should have it not on just show? So mods can link several comments at once?

PR #901, opened 2025-02-09
[UserFeedback] Fix filtering nil during restoration

clragon:
Fixes insertion by correctly filtering `nil` when we skip an entry.


Sindrake:
Third time's the charm.

PR #930, opened 2025-02-21
[AIBUR] Add vote column and ordering option to AIRBUR listings

mm12:
#### Adds a column showing a vote overview to the AUBUR listings
Shows how many 👍, 😐, 👎 there are for each AIBUR. A new `_overview` partial is added, which shows the icon and count only. 
image
image
If there is no forum topic, they do not appear:
image

#### Add a score for ordering
The score follows the formula (positive - negative)/total. This allows neutral (or any other type) of vote to impact the score. For example, an item with 5 positive and 1 neutral is ranked higher than  5 positive and 0 neutral. If there are no votes, the score is 0. 
image

#### Migrations/fixes
Add migration that adds a `vote_score` column to the `forum_post` table, and a fix that populates it. 

### Discussions:
- should the updated in `forum_post_vote` model be private?
- should other changes be added now, to avoid more migrations?
  - Adding other vote options, like 'flip' - scored the same as neutral 
  - Allow votes as 'reactions' on all forum posts - would need new moderation, so it wouldn't be added this time, but the migration can be edited so that it doesn't need to be done if and when it does get added.


mm12:
The failing test has nothing to do with me

app/models/forum_post.rb
```diff
@@ -279,6 +284,10 @@ def readd_tag_change_request_label
     self.body = normalized_body.present? ? "#{label}\n\n#{normalized_body}" : label
   end
 
+  def update_vote_score
+    update_column(:vote_score, vote_score)
+  end
```
Sindrake, line 290: Ditch the `vote_score`.

First of all, it's not helpful. Sure, it conveys the difference between up and downvotes – but that's not really useful information. Any number of downvotes or "meh" votes means that the proposed tag relationship could be questionable.

Second, it would make adding new vote options slightly more complicated.
For example, I would like to add a "reverse" vote option. How would that factor into the calculation?

Finally, I'm not convinced that your implementation even works.
The `update_vote_score` method seems to set the `vote_score` column to its own current value.
The `vote_score_calculation` isn't called from anywhere.

db/fixes/128_add_forum_vote_scores.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require File.expand_path(File.join(File.dirname(__FILE__), "..", "..", "config", "environment"))
+
+ForumPost.in_batches(of: 100) do |batch|
+  updates = []
+
+  batch.each do |fp|
+    next unless fp.votable? # Skip if the post is not votable
+    new_score = fp.send(:vote_score) # Use the calculation method, NOT update_vote_score
+    next if fp.vote_score == new_score # Skip if no change
```
Sindrake: I don't think this works?
`vote_score` just reads the column value. Then, it's compared to itself.

app/models/forum_post_vote.rb
```diff
@@ -10,6 +10,7 @@ class ForumPostVote < ApplicationRecord
   scope :down, -> { where(score: -1) }
   scope :by, ->(user_id) { where(creator_id: user_id) }
   scope :excluding_user, ->(user_id) { where.not(creator_id: user_id) }
+  after_save :update_vote_score
```
Sindrake: This only fires when a vote is created, but not on destroy.
Score does not get recalculated when a vote is removed.

app/controllers/tag_aliases_controller.rb
```diff
@@ -28,7 +28,7 @@ def update
   end
 
   def index
-    @tag_aliases = TagAlias.includes(:antecedent_tag, :consequent_tag, :creator, :approver).search(search_params).paginate(params[:page], :limit => params[:limit])
+    @tag_aliases = TagAlias.includes(:antecedent_tag, :consequent_tag, :creator, :creator, :approver, forum_post: :votes).search(search_params).paginate(params[:page], :limit => params[:limit])
```
Sindrake:
```suggestion
    @tag_aliases = TagAlias.includes(:antecedent_tag, :consequent_tag, :creator, :approver, forum_post: :votes).search(search_params).paginate(params[:page], :limit => params[:limit])
```

Duplicate.

app/controllers/tag_implications_controller.rb
```diff
@@ -24,7 +24,7 @@ def update
   end
 
   def index
-    @tag_implications = TagImplication.includes(:antecedent_tag, :consequent_tag, :creator, :approver).search(search_params).paginate(params[:page], :limit => params[:limit])
+    @tag_implications = TagImplication.includes(:antecedent_tag, :consequent_tag, :creator, :creator, :approver, forum_post: :votes).search(search_params).paginate(params[:page], :limit => params[:limit])
```
Sindrake:
```suggestion
    @tag_implications = TagImplication.includes(:antecedent_tag, :consequent_tag, :creator, :approver, forum_post: :votes).search(search_params).paginate(params[:page], :limit => params[:limit])
```

Duplicate.

app/models/tag_relationship.rb
```diff
@@ -165,6 +165,10 @@ def search(params)
         q = q.order("antecedent_name asc, consequent_name asc")
       when "tag_count"
         q = q.join_consequent.order("consequent_tag.post_count desc, antecedent_name asc, consequent_name asc")
+      when "score_desc"
+        q = q.joins(:forum_post).order("forum_posts.vote_score DESC")
+      when "score_asc"
+        q = q.joins(:forum_post).order("forum_posts.vote_score ASC")
```
Sindrake: This join will drop rows that don't have a corresponding forum post.

Sindrake (changes requested):
Displaying vote counts for AIBURs seems viable.
The `vote_score` column is not, in my opinion.

app/models/forum_post.rb
mm12: I did a refactor recently to make it up-to-date. before that, it worked. `vote_score` is not shown, it is only used for sorting, iirc. As previously discussed in discord, additional items will count the same as `meh`

edit: And it is the diff OUT OF the total votes - so non up/down are accounted for.

app/models/tag_relationship.rb
mm12: Confused here - `forum_post_vote` belongs to `forum_post`, therefore sorting by a vote score when there cannot be a score wouldn't make sense. Is there something I am missing?

db/fixes/128_add_forum_vote_scores.rb
mm12: I _think_ i tested it, but it would have been 17 months ago. Will look into if it still works.

app/models/forum_post_vote.rb
mm12: ah i see, its on [create and update](https://guides.rubyonrails.org/active_record_callbacks.html), but not `destroy`.

app/models/forum_post.rb
mm12: If you want to look into how donovan re-implemented it (see near top), feel free, but life is getting busy so I probably won't have time to do that myself.

mm12: ok, current thought is i changed something to test a function and forgot to undo it. the `calculate` method probably was called `vote_score` at some point.

mm12:
~~TODO: Add test for `vote_score`~~

Sindrake:
Is this a "no" to ditching the `vote_score` then?

mm12:
> Is this a "no" to ditching the `vote_score` then?

If you want to get rid of it I can, but I won't be implementing a new sorting to take its place, meaning that the information on the index will be entirely visual. As described in previous comments, it can handle new vote types just fine. I just don't see a good reason to remove it. I'll try to address the concerns you have: 

> First of all, it's not helpful. Sure, it conveys the difference between up and downvotes – but that's not really useful information. Any number of downvotes or "meh" votes means that the proposed tag relationship could be questionable.

The number is `(up-down)/total` - so `meh` votes bring the score closer to `0`. 

> Second, it would make adding new vote options slightly more complicated. For example, I would like to add a "reverse" vote option. How would that factor into the calculation?

It doesn't, as long as it updates `total`, it will bring it closer to 0. This can be tweaked as needed for future types fairly easily. 

> Finally, I'm not convinced that your implementation even works. The `update_vote_score` method seems to set the `vote_score` column to its own current value. The `vote_score_calculation` isn't called from anywhere.

It didn't, but it does now. `vote_score_calculation` was renamed and not updated in most places, resulting in the issue. 


edit: btw, i see this being useful in the future too. I am adding the column knowing that there is potential for `forumPostVote` to be used in more places in the forum, rather than only tag relation voting.

Sindrake:
Good job on the fixes, but I'm just going to close this.

If we are keeping the `vote_score` column, the migration and fixer script need to be brought back up to the front.
I cannot do this myself, since I am unable to force-push into this branch.

Even with that handled, the fixer script is broken.
```
score_values = updates.map { |_, score| score }.join(", ")   # "0.5, -0.5"
id_list      = updates.map(&:first).join(", ")               # "1, 2"
sql = "UPDATE forum_posts SET vote_score = #{score_values} WHERE id IN (#{id_list})"
# => UPDATE forum_posts SET vote_score = 0.5, -0.5 WHERE id IN (1, 2)
```
The `SET vote_score = 0.5, -0.5` is a syntax error, Postgres reads `, -0.5` as a second assignment with no column. Any batch with more than one row will raise.

Tests are severely lacking.
No test that `update_vote_score` actually writes the column, and none for the unvote-recalculates path, for example.

All in all, unacceptable.

mm12:
>I cannot do this myself, since I am unable to force-push into this branch.

Do you need to rebase, or can you just merge? I do not allow force pushes on my fork because they rewrite history

Feedback noted. It worked when i tested it, but it is what it is. The changes to the tests were planned but i forgot

mm12:
Keeping as draft so I can go off of e621ng/e621ng#2301. If I get time to update it to be compatible, that is, since things are going to be busy for me starting next week. See [discord](https://discord.com/channels/431908090883997698/1526659225919422534) for more info on the usefulness of this feature.

-----

Edit 26.07.17: After looking over 2301, I do not think it's changes have an impact on this PR: 
1. "SWAP" will not impact the score
2. "AIBUR_WARDEN" votes are the same as normal votes. 
  - Warden votes check in the *view* if the vote is from a warden. 
  - In the future, we can change how votes are weighed based on roles. However, it is likely better to wait for `granular scoped permissions` on this, rather than needing to rework it later. 

Therefore, only simple merge conflicts will remain.

PR #987, opened 2025-03-23
[Uploads] Allow mp4s with av1

clragon:
This PR allows uploading mp4s as long as their video codec is av1 and their audio codec is opus.
The mp4 container format is an open container and av1 is an open format.
Only files using opus audio are allowed, as aac is non-free.

I have double-checked this with mairo.


faucetlol:
This PR should probably also revert #783

PR #1034, opened 2025-04-05
[Tags] Add an option to destroy tags

Sindrake:
Somewhat ill-advised, so there are restrictions on this.

* [x] BD staff only
* [x] No posts, deleted or otherwise, can have this tag
* [x] No active aliases or implications

Better info output:
* [x] Add tag deletion to the mod actions
* [x] Display anything that may block the deletion on the corrections page:
    * [x] Total number of posts, including deleted ones
    * [x] Aliases and implications, active or otherwise


Sindrake:
Moving this to next week.  
Gonna need to check whether leaving deleted aliases / implications behind with destroyed tags is going to cause issues.

PR #1036, opened 2025-04-08
[Uploads] Increase the preview image size

Absolute-Arthur:
This is a suggestion based on experience, I'd understand if you do not approve.
Increases the preview image size by removing the flex attribute of the div, making the left (tags and description) and right (preview) evenly sized, allowing to better see the image and tag the details.

Before
image
After
image

### DIFF
--- a/app/javascript/src/javascripts/uploader/uploader.vue.erb
+++ b/app/javascript/src/javascripts/uploader/uploader.vue.erb
@@ -1,6 +1,6 @@
 <template>
     <div class="flex-grid-outer">
-        <div class="col box-section" style="flex: 2 0 0;">
+        <div class="col box-section">
             <div class="flex-grid border-bottom">
                 <div class="col">
                     <label class="section-label" for="post_file">File</label>


Absolute-Arthur:

Sindrake:
Interesting. This is probably a good idea, but I'm not sure if everyone will agree.

faucetlol:
I like this, the form doesn't really need to take up that much space. It's hard to see the finer details on the image when I'm on my laptop and don't have a spare monitor to view the image in another window, and a PITA to keep swapping tabs while tagging. Honestly it would be pretty nice if it was click-and-draggable to resize the panels to users' desired sizes, but that may be overengineering it a little.

On occasion I've actually scaled down my browser window so it displays the mobile mode, because the image then displays larger even if my browser window is only using 50% of my screen width.

On an unrelated tangent, I did once experiment with hover-to-zoom on the image preview and it was pretty good for details, especially stuff like seeing eye colors. I never got it to a quality that I thought was good enough to make a PR, though.

PR #1047, opened 2025-04-10
[Post] Fix delreason breaking on paging

mm12:
Bug Summary: Using post navigation (prev/next) leads back to the same post. 
Bug Circumstances: Post is first/last on page (clicking the button goes to a new page), and `delreason` is in query

Cause: Searches automatically allow `status:deleted` when `delreason` is in the query. `PostSearchContext` didn't get the memo, and only checked for the `status` metatag. 

Fix: add `delreason` to the exceptions for automatically adding `-status:deleted` in `PostSearchContext`

### DIFF
--- a/app/logical/post_search_context.rb
+++ b/app/logical/post_search_context.rb
@@ -6,7 +6,7 @@ class PostSearchContext
   def initialize(params)
     tags = params[:q].presence || params[:tags].presence || ""
     tags += " rating:s" if CurrentUser.safe_mode?
-    tags += " -status:deleted" unless TagQuery.has_metatag?(tags, "status", "-status")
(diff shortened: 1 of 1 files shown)


jmortiger:
The same problem exists with `deletedby`. That being said, I'm 70% sure both this & the preceding `rating:s` are redundant. The same code caused a similar problem in `app/logical/post_sets/post_sets.rb` and removing that line fixed it. `ElasticPostQueryBuilder` already checks for whether to auto-hide deleted posts through `TagQuery`, & it accepts a `enable_safe_mode` argument that defaults to the value of `CurrentUser.safe_mode?` in the post model. I'm planning on checking tomorrow if these can just be removed.

mm12:
Closing in favor of #1050

PR #1077, opened 2025-04-25
[Post Search] `order:rank` alternative/replacement (`order:hot`)

jmortiger:
Fixes [forum #58375](https://e621.net/forum_topics/58375).
* Adds an ordering method (`order:hot`)
* Removes the 2 day restriction from `order:rank`
* Replaces all current uses of `order:rank` w/ `order:hot` (e.g. `Hot` page)
* Added `hot_from` metatag to define starting range for `order:hot`

Similarly to `order:rank`, `order:hot` seeks to order posts uploaded in the past 2 days based on both their score & how recently they were posted; however, while the `rank` algorithm is arguably inconsistent and is viewed as confusing, this replacement algorithm used in `hot` is far more consistent, self-explanatory, & predictable. As shown [here](https://www.desmos.com/calculator/ammxws6vsl), this algorithm simply scales a post's score by how deep into the 2 day range it is, with posts uploaded 2 days ago having their resultant score being halved. This way, an extremely high-scoring post won't be placed after a post that has less than half it's score. As shown in the graph, the curve of the scaling can be adjusted by changing `ElasticPostQueryBuilder::RANK_EXPONENT`. The current scale has been chosen to have posts' effective scores decay at a fairly steady rate while still being smooth.
* An exponent of 0 removes the impact of score altogether
* An exponent of 0<x<1 makes the rate of change increase over time
* An exponent of 1 results in a linear decay
* An exponent of >1 makes the rate of change decrease over time

This PR also replaces the prior usage of `order:rank` & both removes the time restrictions from `order:rank` & adds the `hot_from` metatag to enable direct comparisons of their output at any point in time (use `order:rank date:<x+2>_days_ago..<x>_days_ago` to restrict to the same time frame).

### DIFF
--- a/app/helpers/posts_helper.rb
+++ b/app/helpers/posts_helper.rb
@@ -2,8 +2,7 @@
 
 module PostsHelper
   def discover_mode?
-    # params[:tags] =~ /order:rank/
-    params[:tags] =~ /order:(?>rank|hot)/
+    params[:tags] =~ /order:hot/
   end
 
   def next_page_url
--- a/app/helpers/posts_helper.rb
+++ b/app/helpers/posts_helper.rb
@@ -2,7 +2,8 @@
 
 module PostsHelper
   def discover_mode?
-    params[:tags] =~ /order:rank/
+    # params[:tags] =~ /order:rank/
+    params[:tags] =~ /order:(?>rank|hot)/
   end
 
   def next_page_url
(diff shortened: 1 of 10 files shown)


jmortiger:
Hasn't been thoroughly retested (my container's acting up), but I'm fairly confident this is good to go

PR #1088, opened 2025-05-03
[Posts] Revamped video conversion

Sindrake:
Reworked the video conversion from the ground up.
Some key points
* only generating h264 samples now
* reworked ffmpeg parameters for better results
* reduced the frame rate in samples, but only if the original was over 30 fps
* the fit vertically and fit horizontally modes now display a sample that is capped at 1080p


Sindrake:
Going to have to look into replacing the current ffmpeg wrapper with something that had seen an update in the last decade.

PR #1091, opened 2025-05-04
[AIBUR] Show information on relation type.

mm12:
ui tweak: Add a warning about what type of tag relationship request is being made. 

This warning is made using Dtext. Changing the implementation might be a good idea, so this will be a draft for now.


mm12:
Thoughts: 
- Should be backed by a wiki page instead. 
- ~~Perhaps use the tag hint snippets that show in search?~~ not a view component

Sindrake:
The A/I pages need a larger design overhaul.

PR #1111, opened 2025-05-18
[TagPreview] Add new tag preview

clragon:
Adds a new tag preview that updates automatically, 
in functionality identical to re6's live tag preview but in visuals more similar to e6's vanilla design. 

This removes the old "Preview final tags" button and repurposes it's endpoint.

Below is a preview of the new tag preview:

image

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -13,18 +13,15 @@ class TagsController < ApplicationController
   end
 
   def index
-    @tags = Tag.search(search_params).paginate(params[:page], :limit => params[:limit], :search_count => params[:search])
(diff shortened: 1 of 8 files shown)


Sindrake:
This is looking promising, but it's not ready for release yet.

The preview crashes catastrophically on posts with a lot of tags.
I would also look into how this error message gets mangled like this, since this definitely isn't normal.
Screenshot 2025-05-24 165311

The upload page also spams the console with errors:
Screenshot 2025-05-24 165145

This definitely needs more client-side caching.
As of right now, the entire tag list gets sent over to the back-end every time the user adds a tag. Instead, you should store the tag data client-side, and only fetch new info for tags that are not present there.

Also, the tag output should definitely be limited in height – could be something like `50vh`, with scrollbars added after that.

PR #1157, opened 2025-06-22
[PostCache.js] Fixed pool parsing bug

jmortiger:
Fixed pool parsing bug that caused `pool` blacklist filters to fail.

See [here](https://discordapp.com/channels/431908090883997698/460517895420772353/1385794598806360136) & [here](https://discordapp.com/channels/431908090883997698/460517895420772353/1386317981801381949)

### Note
This was created through the web editor, and is thus untested (though I don't see what could fail). This also leaves a potential future source of problems marked by a `NOTE` comment.

### DIFF
--- a/app/javascript/src/javascripts/models/PostCache.js
+++ b/app/javascript/src/javascripts/models/PostCache.js
@@ -7,6 +7,67 @@ export default class PostCache {
 
   static _elements = {};
 
+  // IDEA: Use the generic builder below instead of 2 separate instances of the same function.
+  /* static fromGeneric (data, id = undefined) {
+    // Likely won't happen, but won't hurt to check
+    if (!id) {
+      id = parseInt(data.id);
+      if (!id) return null;
+    }
+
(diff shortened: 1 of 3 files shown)


Sindrake:
This was a bit of a rabbit hole for me.
Some of the decisions I made when I wrote the new blacklist handler were somewhat questionable.

I think that it's better to return the list of pools as a space-separated string.
Parsing JSON is not terribly performant, and while the alternative you proposed [is viable](https://jsbench.me/y0mclclmf4/1) I think we might as well bypass the issue altogether. What complicates this is the fact that setting an array as an attribute [here](https://github.com/e621ng/e621ng/blob/master/app/javascript/src/javascripts/thumbnails.js#L63) automatically converts it to a space-separated string anyways, and writing a special case to have it remain as stringified array would be pointlessly complicated.

I've also implemented your idea of a generic builder, albeit somewhat differently.
I'm certain that there was a reason I had two separate functions doing pretty much the same thing, but it's clearly not needed anymore.

PR #1158, opened 2025-06-24
[Files] Add Webp support

### LINKED ISSUE #1337: Add webp support
The Webp file format is [widely supported](https://github.com/e621ng/e621ng/blob/master/app/javascript/src/javascripts/autocompletable_input.vue) and used on some of our sources as original file. 

Because of this, it would be useful to allow uploading webps directly to our site, increasing coverage of original files, with good quality and file size.

The server already uses webp internally for thumbnails. The change to support this is adjusting the allowed file list.

mm12:
Adds Webp support to uploader, file validator, tests. 

Fixes #1337.


config/danbooru_default_config.rb
mm12, line 1: In reality, it is unlikely for webp to hit this, except maybe on long animations.

app/javascript/src/javascripts/uploader/file_input.vue
mm12, line 1: I am considering cleaning up the string, since it is a bit of a mess. Preferable, formats would be together instead of being separated by file and mime type

mm12 (commented):
I actually considered adding AVIF as well, but VIPS doesn't support it without some modification to the module itself, seemingly. As for JPEG XL, it has basically no browser support, so it will be a while before then.

test/unit/post_test.rb
```diff
@@ -1000,12 +1000,15 @@ def assert_tag_match(posts, query)
       end
 
       context "tagged with animated_gif or animated_png" do
```
clragon:
```suggestion
      context "tagged with animated_gif, animated_png or animated_webp" do
```

clragon (changes requested):

clragon (dismissed):

PR #1184, opened 2025-07-12
[Post] Rework the favorite transfer process

mm12:
* Do favorite/set moving AFTER the rest of the moves. This _might_ fix the deadlock on deletions.
 * Consistency: Manual favorite moves also should move sets, like deletion does.

My theory is the favorite/set moving job is taking too long, and it deadlocks any other moves (sources/tags). If this is true, then having them take place last should significantly reduce the chances of it happening. Hard to tell if this will work on a small scale.

Edit: Going to add some more tooling here that should make transferring items easier and clearer, even if the fix doesn't work

### DIFF
--- a/app/controllers/favorites_controller.rb
+++ b/app/controllers/favorites_controller.rb
@@ -31,6 +31,12 @@ class FavoritesController < ApplicationController
 
   def create
     @post = Post.find(params[:post_id])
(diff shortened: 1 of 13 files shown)


mm12:
mb, thought this was done

PR #1193, opened 2025-07-23
[Gems] Update nokogiri

Sindrake:


faucetlol:
Maybe I'm blind, but I don't see anywhere that Nokogiri is actually being utilized anywhere in e621ng?

Sindrake:
> Maybe I'm blind, but I don't see anywhere that Nokogiri is actually being utilized anywhere in e621ng?

We are not using it directly, but it's a dependency for [actionpack](https://rubygems.org/gems/actionpack/versions/8.0.2).
I've updated it because of a [security advisory](https://github.com/advisories/GHSA-353f-x4gh-cqq8).

PR #1196, opened 2025-07-24
[Tags] Uniformly add abbreviated millions to tag counts (#1189)

jmortiger:
I took a look at e926, & found the tag counts are still not displaying millions.
After looking into it, it looks like the tag set presenter modified by #1189 isn't actually used to render tag sets, so the change isn't reflected by the UI; a partial view is used instead.
I'll look at removing one of these redundant files some other time, but I've found the actual file that renders these elements & confirmed this one actually does work.

#### Edit
After some more digging, it turns out `TagSetPresenter` *is* used to render the list of tags in the sidebar... but *only for favorites*. If you're logged in, [this page](https://e926.net/favorites) _will_ correctly show the updated labels. I'm working on making them all use the presenter, but until then, this will do fine.


Sindrake:
I see what happened here.

A presenter _used_ to be how the tag list in the sidebar was assembled.
I reworked it at one point, to make it work better on mobile and to allow for the quick blacklist buttons. I used a partial instead of the presenter, since I was not happy with how complicated and messy that function was becoming. It seems like I forgot to use this new system on the favorites page, and didn't remove the old functions from the presenter.

I believe that the best way to go here would be to get rid of the old code in the presenter altogether.

app/views/posts/partials/common/sidebar/_tag_list_item.html.erb
```diff
@@ -17,8 +17,16 @@
...
+  <% if post_count < 1_000 %>
+    <% post_count = post_count.to_s %>
+  <% elsif post_count < 10_000 %>
+    <%post_count = format("%.1fk", (count / 1_000.0)) %>
+  <% elsif post_count < 1_000_000 %>
+    <% post_count = "#{post_count / 1_000}k" %>
+  <% elsif post_count < 10_000_000 %>
+    <% post_count = format("%.1fm", (count / 1_000_000.0)) %>
+  <% else %>
+    <% post_count = "#{post_count / 1_000_000}m" %>
```
Sindrake, line 29: At this point, it might be best to move this whole block into a helper.

jmortiger:
Yeah, I'm currently in the process of pushing everything to partials & merging all the sidebar stuff into a common partial on my most recent PR; this was a bandaid to get it out the door while I untangle everything.

app/views/posts/partials/common/sidebar/_tag_list_item.html.erb
jmortiger: I can pull the helper from my more thorough branch back into this.

jmortiger:
Also, this is still used for inline tag rendering, so completely ditching the tag presenter would require pulling that out too.

Sindrake:
Well, there isn't really a pressing need for an urgent fix.
So might as well do it properly.

jmortiger:
Closing in favor of #1200

jmortiger:
Closing in favor of #1200

PR #1197, opened 2025-07-24
[DMails]  Unified response format for dmails, for empty and filled data

alphyron:
The json response were doesen't have the same schema, depending if the list is empty or not.

### DIFF
--- a/app/controllers/dmails_controller.rb
+++ b/app/controllers/dmails_controller.rb
@@ -12,7 +12,7 @@ class DmailsController < ApplicationController
     @query = Dmail.active.visible.search(search_params).includes(:to, :from)
     @dmails = @query.paginate(params[:page], limit: params[:limit])
 
-    respond_with dmails: @dmails
+    respond_with @dmails.to_json
   end
 
   def show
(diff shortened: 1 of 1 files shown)


alphyron:

alphyron:
@dragon Is there a reason for your downvote? I am always open for suggestions.

alphyron:
@clragon Is there a reason for your downvote? I am always open for suggestions.

clragon:
The way rails uses json (by adding a unnecessary root key) is both very annoying and extremely inconsistent across all of the code.
However, the preferred result is to _omit_ this root key. This is much easier and the expected standard for any API consumer.

Because of that, I'd rather have a PR that ensures even empty results do not produce a root key. Perhaps that could be modifying that helper function, to fix this problem across many endpoints all at once.

alphyron:
@clragon I definitely see where you are coming from and would agree with you. There could be a way to fix it like your suggestion. The only question is if you want to do that. Currently the most API Endpoints have an root key and so the most users of the API are configured to use it with a root key. In my opinion, that would break too much. 

If that doesn't play matter and e621 want to change it nevertheless of course I can change it to that. Or at least I can try my best to.

alphyron:
I removed the root key on both cases. Only on DMails currently, don't want to do it on all other endpoints without further consultation.

PR #1199, opened 2025-07-28
[Users] Added `:is_verified?` to the user API response

jmortiger:


Sindrake:
I thought that this was referring to the artist verification, and was quite confused for a bit.
Perhaps, we can consider renaming this attribute to something like `email_verified` in the future.

PR #1200, opened 2025-07-29
[Posts] [Favs] [Tags] Unified sidebar tag list rendering

jmortiger:
Removes most of `TagSetPresenter` in favor of using partials for sidebar tag list rendering.
* Unify tag list sidebar partials
* Unify favorites & post search partials

This currently leaves `TagSetPresenter` for usage w/ inline tag display (specifically with uploads)


Sindrake:
I removed the rubocop todo list changes from the PR.
From now on, I'll probably just have that list regenerated on a schedule, once a month or so. I saw a few github actions that might help automate that.

Also, I fixed a small issue that caused the plus and minus icons not to show up next to the tags on the favorites page.
I'm not _entirely_ sure whether this is a bug, since the current implementation on e6 also lacks those. I guess we'll see if people will complain.

PR #1260, opened 2025-08-30
[ApiKeys] Add API key management

### LINKED ISSUE #736: API key management
Currently its only possible to have one API key. Thats a bit encumbering when we use many different third party applications.

Implement proper API key management. 

This includes:
- api key creation and revokation
- api key names
- api key expiration date (optional)

these extra features would make API keys much more safe. 
if we wanna go overboard, perhaps we could display which user agents have accessed the key, but I am not sure if that data is available.

clragon:
Adds the ability to have multiple Api keys.

- Api keys now have names, as well as an optional expiration date.
- Each Api key now tracks its last used IP and user agent.
- A usage tracking job that is deduplicated has been added*.
- Api keys are disallowed from accessing Api key routes.
- An index page for Api keys has been added.
- An Api key controller with tests has been implemented.
- The settings page has been updated to point to the new Api key page.
- A migration script to convert the table has been included.

The code in this PR is based on work by @DonovanDMC specifically in https://github.com/FemboyFans/FemboyFans/commit/018eca40e7d549c2e728862d1508be21da4845df.

<sub>\* The usage job can only be triggered once every minute. Any new jobs with the same key during that minute are _rejected_, this means should IPs or user agents differ, they will not be updated within that minute. This seemed like a sensible trade-off to me, rather than continously updating or replacing the job. It would have been possible to track the new parameters in Redis on replacement. Replacing the job on a new request would also pose issues with continues requests not updating anything at all, since the job is always replaced.</sub>

Fixes #736.

image

image

image


clragon:
Addtional ideas:

1. Disallow editing API keys?
    Being able to modify the expiration date after creation is strange.
2. Default to an expiration date of one year
    It might be better to limit the keys by default, unless a user chooses no expiration

clragon:
GitHub for example has predefined ranges for expiration for their keys:

image

image

image

Which seems like a sensible approach.

clragon:
Lasty, it might also be worth considering adding a search page in case we want administrators or BD staff to have insight into the api keys that users create.

clragon:
I have implemented better selection UI and disabled editing.
I will also remove usage count and implement some kind of debouncing of the writing of which IP last accessed the key.
We might be able to do that with Sidekiq and Redis. I have investigated it, but the exact amount of jobs or when to write at scale isnt clear to me.

Sindrake (changes requested):
The mobile UI looks terrible.

Image

The rest is excellent. Very well done.

app/jobs/update_api_key_usage_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def self.lock_args(args)
+    [args[0]] # ignore different ip and user agent
+  end
+
+  def perform(api_key_id, ip_address, user_agent = nil)
+    api_key = ApiKey.find_by(id: api_key_id)
+    return unless api_key
+
+    api_key.update_usage!(ip_address, user_agent)
```
Sindrake: This will fail for legacy API keys, since they do not have names.
```
ActiveRecord::RecordInvalid: Validation failed: Name can't be blank
```

Sindrake (changes requested):

app/jobs/update_api_key_usage_job.rb
clragon: The UI has been mobilized.

app/logical/session_loader.rb
```diff
@@ -118,7 +118,7 @@ def authenticate_api_key(name, api_key)
     user, api_key_record = auth_result
     CurrentUser.user = user
     CurrentUser.api_key = api_key_record
-    api_key_record&.update_usage!(@request.remote_ip)
+    UpdateApiKeyUsageJob.set(wait: 1.minute).perform_later(api_key_record.id, @request.remote_ip, @request.user_agent) if api_key_record
```
kiranoot: I believe that this is likely to overload the sidekiq system with jobs. Jobs still have to be distributed to workers and there is overhead related to this.

A better approach would be to use a system like the throttle code uses(Or the throttle code itself). With Redis or some other memory store.

DonovanDMC:
Forcing regeneration of api keys that have been in use for years and kept perfectly hidden breaking several workflows that might still just work while the creator doesn't even remember they exist isn't cool

Don't be like NPM

clragon:
I understand that concern, @DonovanDMC, but it was previously impossible to create expiring API keys. For that reason, this seems like a good opportunity to start again, nudging users to explicitly choose an expiration date and a descriptive name for their API key(s), as well as invalidating all the keys which might have been forgotten about by now. 

I do think making all old API keys end eventually would be very useful, more so than not breaking existing workflows. Do you feel this is not the case?

DonovanDMC:
Could there possibly be a compromise somewhere in the middle? Allow users to somehow confirm or convert their existing api key while keeping the actual key part in tact, and then expiring all unconfirmed/unconverted keys after 6 months? It would likely require a temporary column or some other mechanism but I see that  as a small price to pay over needlessly invalidating all existing keys with no choice in the matter

(also fun fact: apikeys predating ng are 32 characters while newly generated keys are 24 characters, making older keys theoretically more secure)

kiranoot:
I would make it so that if old keys expire, they don't all expire on the same day. Otherwise the support overhead of dealing with it will be much higher than if it's staggered. Say over a month. It could also use a news entry if it's going to be done.

clragon:
We could add a temporary column that marks old API keys as legacy, which then allows removing their expiration date.  Assuming we would wanna allow an API key rescue, we probably would want users to pick a name and inform them that API keys with expiration dates are what we recommend. Maybe we could simply let them "edit" their old API key. I have removed editing keys generally though, since I would push for immutable, time-gated secrets as the understanding of API keys.

I do find that this somewhat enables users to pick the "easy" route of not renewing their API key. From previous experiences, the average user seems to miss out on our announcements. Likely, they wont be aware that their keys will expire, they expire and they have to replace them anyways. For this scenario, it doesnt seem worthwhile to include an option to "rescue" a legacy API key. 

I suppose we could send an email to every user who has an API key to inform them that they will expire, whether or not we support rescuing the keys. That would address the problem of being uninformed to some degree, and might prevent support cases.

From reading the code, new keys also dont appear to be "less" secure than old keys. The underlying implementation is actually has better randomness.

clragon:
I have come the conclusion that, while expiring old api keys seems attractive from a security standpoint, the orchestration of making such a transition a passable experience for most users is more complex than the gained security is worth. 

For that reason, I have removed the addition of expiration dates for old API keys from the code.

Sindrake:
Would you mind rebasing this branch?

Sindrake (dismissed):

PR #1279, opened 2025-09-03
[Posts] Alias `order:rank` to `order:hot`

jmortiger:
Not yet tested.


jmortiger:
*sigh*... I'll deal with it later...

PR #1284, opened 2025-09-06
[Icons] Added icon creator script `gen_icons`

jmortiger:
This allows e6 & downstream projects to easily overwrite almost all of their icons by creating the main icon (`public/images/logo/default.svg`) & then running this script.

### DIFF
--- a/gen_icons
+++ b/gen_icons
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+# cspell:ignore Inkscape mstile
+
+declare sourceFile="public/images/logo/default.svg"
+while [[ -n "$1" ]]; do
+	case "$1" in
+		-s | --source-file-path)
+			shift
+			sourceFile="$1"
+			shift
+			;;
+		--)
+			shift
+			;;
+		--help | -h)
+			cat <<-EOF
+				Uses Inkscape to generate the icons from the source SVG (defaults to "$sourceFile"). The only ones that must be manually created are "safari-pinned-tab.svg" (requires manual design) & "apple-touch-icon.png" must be manually edited to add the background.
+				Usage:	$0 --source-file-path SOURCE_FILE
+								$0 -s SOURCE_FILE
+								$0 --help
+			EOF
+			exit
+			;;
+	esac
+done
+inkscape -D --export-type=png --export-png-color-mode=RGBA_8 -h 256 -o public/images/main-logo.png "$sourceFile"
+# #region android-chrome
+inkscape -D --export-type=png --export-png-color-mode=RGBA_8 -h 192 -o public/android-chrome-192x192.png "$sourceFile"
+inkscape -D --export-type=png --export-png-color-mode=RGBA_8 -h 512 -o public/android-chrome-512x512.png "$sourceFile"
+# #endregion android-chrome
+# #region Apple
+# TODO: Add blue background
+inkscape -D --export-type=png --export-png-color-mode=RGBA_8 -h 180 -o public/apple-touch-icon.png "$sourceFile"
+# TODO: Figure out safari-pinned-tab.svg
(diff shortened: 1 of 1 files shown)


pupgray:
This might be easier to use without the dependency on inkscape. We could reuse libvips since it needs to be installed anyway, and run it in a rake task. Something like this maybe? (just barely working with some tweaked dimensions, still needs .ico export/seperate SVG for the small favicons that use a version of the logo without 621)

```ruby
task gen_icon: :environment do
  filename = Rails.public_path.join "images/logo/default.svg"
  source = IO.binread filename

  versions = {
    "images/main-logo.png" => proc { _1.thumbnail_image(256) },
    "android-chrome-192x192.png" => proc { _1.thumbnail_image(192) },
    "android-chrome-512x512.png" => proc { _1.thumbnail_image(512) },
    "apple-touch-icon.png" => proc { _1.thumbnail_image(150).gravity(:centre, 180, 180, extend: :background, background: [0, 0, 0, 0]).flatten(background: [19, 47, 84]) },
    "mstile-144x144.png" => proc { _1.thumbnail_image(144) },
    "mstile-70x70.png" => proc { _1.thumbnail_image(70).gravity(:centre, 128, 128, extend: :background, background: [0, 0, 0, 0]) },
    "mstile-150x150.png" => proc { _1.thumbnail_image(150).gravity(:centre, 270, 270, extend: :background, background: [0, 0, 0, 0]) },
    "mstile-310x310.png" => proc { _1.thumbnail_image(310).gravity(:centre, 558, 558, extend: :background, background: [0, 0, 0, 0]) },
    "mstile-310x150.png" => proc { _1.thumbnail_image(310, height: 150, crop: :none).gravity(:centre, 558, 270, extend: :background, background: [0, 0, 0, 0]) },
    "favicon-32x32.png" => proc { _1.thumbnail_image(32) },
    "favicon-16x16.png" => proc { _1.thumbnail_image(16) },
  }

  versions.each_pair do |name, proc|
    image = Vips::Image.svgload_buffer(source, unlimited: true)
    proc.call(image).pngsave(Rails.public_path.join(name).to_s, compression: 5)
  end
end
```

jmortiger:
That would probably work, I just threw this together in 10 minutes for personal use to fix e6ai still using the e621 icon in various places, realized they/e621/other forks might need to change the icon in the future, and made a PR for it. I didn't want to deal with docker or running it locally for quick script and already had Inkscape installed (I was originally going to do it manually with Inkscape anyways), plus I'm unfamiliar with vips and don't want to go through the trouble of learning it for this one thing. Only potential problem I can see is that, at least currently, the various icons aren't actually square (at least in some cases), and I think the original currently used icons were generated using Inkscape, or at least the same settings (they seem to come out to the same output resolution and file size), and I don't want to change them unilaterally.

Sindrake:
This is interesting, but in the grand scheme of things it seems unnecessary.

Both e621 and e6ai required tweaks to the favicons at lower resolutions to remain legible, which must be done manually anyways.
And since this task only needs to be done once per project, automating the process isn't really a requirement.

jmortiger:
I'd agree it is unnecessary, but I would note that
* this does create a potentially useful template for most output resolutions (excluding the lowest ones of course)
* this is helpful for anyone else who might want to make their own fork down the line
* there was discussion of changing the e6ai name (& presumably icon) at some point
* [e926 is also using e621's icon assets](<https://e926.net/apple-touch-icon.png>) for everything (including the SVGs rendered on the page) & it'd be nice to not have to manually export every icon if/when that's resolved

Tweaking it to map different SVG input files to different output resolutions (w/ potentially different settings) would also likely help that.

I'll close it, at least for now; maybe I'll revisit it way down the line & make it more configurable if we ever decide on a redesign.

PR #1289, opened 2025-09-09
[Tickets] Allow staff to see an accused user's record preview

mm12:
> [!IMPORTANT]
> Requires e621ng/e621ng#1881 
> e621ng/e621ng#1890 is recommended as well, but strictly speaking, isn't required for functionality.

Add the record preview that is visible on the posts page to user tickets. I am considering adding them to other user fields (submitter, accused), thought will need to do some consulting since there is already links there for moderators.


mm12:
Ok, well, I know the current code does not work. However, in looking into this, I have made some findings. 
First, we actually have 2 implementations of what is pretty much the same thing. We have `PostsHelper.user_record_meta`, and `UsersHelper.user_feedback_badge`. Of course, one of these is the "classic" text version, and the other is the badges. 
Second, these both call `feedback_pieces`, which loops through items and iterates counts. 

~~Ideally, both helpers would be in the `UserHelper` or perhaps in a new `UserFeedbacksHelper`. However that seems a bit out of scope for this PR...~~

Ideally, both would be in a `component` so they could be used across the application.

Sindrake (approved):

PR #1291, opened 2025-09-10
[Admin] Editing a user's profile automatically adds a staff note with the prior text

jmortiger:
Haven't added accompanying test yet, still figuring out how best to format such.


Sindrake:
Let's come back to it later.

PR #1320, opened 2025-09-21
[PostFlags] Allow some flag reasons to require additonal context

### LINKED ISSUE #1322: Force adding proof for certain types of flags
> Can we force users to add their evidence instead of allowing flags like this. Tired of flags that require DMail follow-up or we let the suspect tracer walk.

Image

Done in e621ng/e621ng#1320

mm12:
Disallows empty 'note' for flag reasons that should always include more context. Closes #1322


app/views/post_flags/_new.html.erb
```diff
@@ -55,10 +55,62 @@
 
 <%= javascript_tag nonce: true do -%>
   const flagReasonLabels = document.querySelectorAll(".flag-reason-label");
+  const noteField = document.getElementById("flag_note_field");
```
Sindrake: At this point, move this into a js file.
I believe the old script was inlined because it was pretty small, but that is no longer the case.

Sindrake (changes requested):

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+  document.addEventListener("DOMContentLoaded", () => PostFlags.init());
+} else {
+  // Document already loaded
   PostFlags.init();
-});
+}
+
+// Support Turbo (Rails 7) or Turbolinks navigation
+document.addEventListener("turbo:load", () => PostFlags.init());
+document.addEventListener("turbolinks:load", () => PostFlags.init());
```
clragon: We dont have those (and likely never will).
```suggestion
```

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
 
-$(() => {
+// Auto-initialize on standard DOM load
+if (document.readyState === "loading") {
+  document.addEventListener("DOMContentLoaded", () => PostFlags.init());
+} else {
+  // Document already loaded
   PostFlags.init();
-});
+}
```
clragon: Jquery `$` already runs on document load. This change only makes the code longer.
```suggestion
$(() => {
  PostFlags.init();
});
```

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+      console.debug("PostFlags: no radio buttons found for reason_name");
+    }
+    if (flagReasonLabels && flagReasonLabels.length) {
+      flagReasonLabels.forEach(label => {
+        label.addEventListener("click", () => {
+          flagReasonLabels.forEach(l => l.classList.remove("selected"));
+          label.classList.add("selected");
+        });
+      });
+    }
```
clragon: This "fallback" introduces loads of code for an impossible scenario. 
The container cant not be present, since we wrote it into the HTML ourselves,
so this never happens. If this does happen somehow, its likely just that we broke something in a terrible way, and this code should stop working, rather than try to recover.

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+          if (noteField) noteField.focus();
+        } else {
+          clearError();
+        }
+      } catch (err) {
+        console.error("PostFlags: submit handler error", err);
+      }
+    });
+  } else {
+    console.debug("PostFlags: form not found; submit handler not attached");
```
clragon: This, together with the other debug messages, introduces loads of noisy messages into the console on each page, for little to no benefit. 

Image

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+  }
+  let errorMessage = null;
+
+  // Parse dataset boolean values robustly. Treat presence or values like "", "true", "1", "yes" as true.
+  function parseDatasetBoolean (val) {
+    if (val === undefined || val === null) return false;
+    const v = String(val).trim().toLowerCase();
+    if (v === "" || v === "true" || v === "1" || v === "yes") return true;
+    return false;
+  }
```
clragon: The config file is written in ruby, and only `true` and `false` are valid booleans. Empty string, `1`, and `yes` are therefore not candidates for values we would receive from there. For that reason I believe this function is unnecessary.

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+
+  function updateNoteRequired () {
+    try {
+      if (!noteField) return;
+      const selected = document.querySelector("input[name=\"post_flag[reason_name]\"]:checked");
+      const label = noteField.closest(".flag-notes")?.querySelector("label");
+      if (selected && parseDatasetBoolean(selected.dataset.requireExplanation)) {
+        noteField.required = true;
+        if (label) {
+          // Add a small indicator span instead of replacing the label content
```
clragon: This is a meaningless LLM comment that doesnt explain anything about the current code, only about its previous version, which is not present in either the history nor this commit.

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+    }
+  }
+
+  if (form) {
+    console.debug("PostFlags: attaching submit handler to form", form);
+    form.addEventListener("submit", function (e) {
+      try {
+        updateNoteRequired();
+        if (isNoteRequiredAndEmpty()) {
+          e.preventDefault();
```
clragon: This code is buggy, because it will prevent the submit event, but Rails UJS will disable the button on submit, to prevent double-submits. This means, we can no longer submit after receiving this error. 
To fix this, we can prevent Rails from doing this 
```ruby
      <%= f.submit "Submit", class: "st-button submit", data: { disable_with: false } %>
```
and do it ourselves:
```js
      } else {
        clearError();
        e.submitter.disabled = true;
      }
```

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,176 @@
...
+
+  function isNoteRequiredAndEmpty () {
+    try {
+      if (!noteField) return false;
+      const selected = document.querySelector("input[name=\"post_flag[reason_name]\"]:checked");
+      return selected && parseDatasetBoolean(selected.dataset.requireExplanation) && noteField.value.trim() === "";
+    } catch (err) {
+      console.error("PostFlags.isNoteRequiredAndEmpty error", err);
+      return false;
+    }
```
clragon: The code contains a lot of try-catch statements like this one that catch nothing. None of the functions inside the try body can actually throw, so this just bloats the code.

clragon (changes requested):

app/javascript/src/javascripts/post_flags.js
mm12: Sorry, added those when I moved it to the JS file, and the submit button stopped working

app/javascript/src/javascripts/post_flags.js
clragon: We can also use the very convenient inbuilt [Boolean function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) instead:

```suggestion
  function parseDatasetBoolean (val) {
    return Boolean(val);
  }
```

But we should inline that.

app/javascript/src/javascripts/post_flags.js
```diff
@@ -1,17 +1,121 @@
...
+    form.addEventListener("submit", function (e) {
+      updateNoteRequired();
+      if (isNoteRequiredAndEmpty()) {
+        e.preventDefault();
+        showError("Note is required for the selected reason.");
+        if (noteField) noteField.focus();
+      } else {
+        clearError();
+        if (e.submitter) {
+          e.submitter.disabled = true;
```
mm12: Was warned that this isn't supported by some browsers. I am not familiar with JS, will need info on it

clragon: We can see the docs for this field [here](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter).
According to canIuse, this is available [basically everywhere](https://caniuse.com/mdn-api_submitevent_submitter).

clragon (commented):
This code looks okay to me now. 
I am not a fan of the LLM comments, but thats whatever.

Right now, this also uses a lot of JS to spawn HTML elements that dont need to be dynamically created; They could be part of the HTML and edited/shown/hidden when relevant. This would make things a lot less complex. I don't ask for that change though, since its a bunch of work, and this works.

mm12:
> This code looks okay to me now. I am not a fan of the LLM comments, but thats whatever.
> 
> Right now, this also uses a lot of JS to spawn HTML elements that dont need to be dynamically created; They could be part of the HTML and edited/shown/hidden when relevant. This would make things a lot less complex. I don't ask for that change though, since its a bunch of work, and this works.

I would really rather do that, honestly. 

edit: I messed around with it a bit, and the problem is that the label for the field is auto-generated, so you can't add text stating the field is required in the desired place without doing essentially the same thing. 
image

Sindrake (approved):

PR #1342, opened 2025-09-25
[Posts] Fix MP4 videos not pausing when the blacklist is toggled

### LINKED ISSUE #1341: [Bug] MP4 uploads aren't paused when toggling on blacklist
**Describe the bug**
When enabling your blacklist on a post page, playing blacklisted videos should be paused. WebMs are paused, but MP4s aren't.

**To Reproduce**
1. Open an MP4 post with tags on your blacklist
2. Disable your blacklist
3. Play the video
4. Enable your blacklist

**Expected behavior**
The video pauses.

**Additional context**
This happens because the code to pause blacklisted videos [identifies videos by the property `data-file-ext="webm"`](https://github.com/e621ng/e621ng/blob/872707931797d2eefa7ddc38001ef97f2bbd9ff2/app/javascript/src/javascripts/blacklists.js#L249), which *was* always true for videos until MP4 support was added.

ghost:
Closes #1341

### DIFF
--- a/app/javascript/src/javascripts/blacklists.js
+++ b/app/javascript/src/javascripts/blacklists.js
@@ -246,11 +246,13 @@ $(() => {
   // This seems extraordinarily uncommon, so it's here
   // just for feature parity with the old blacklist.
   if (!Page.matches("posts", "show")) return;
-  let container = $("#image-container[data-file-ext='webm']").on("blk:hide", () => {
-    const video = container.find("video");
-    if (!video.length) return;
-    video[0].pause();
-  });
+  let container = $("#image-container[data-file-ext='mp4'], \
+                    #image-container[data-file-ext='webm']")
+    .on("blk:hide", () => {
+      const video = container.find("video");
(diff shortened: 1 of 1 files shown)


ghost:

ghost:
Surprised I manged to break tests with a single line change. Fixed that and squashed it into a commit fitting your naming scheme.

clragon:
Thank you for your PR!

Perhaps we should find a more reliable way to detect whether the Post is a video rather than checking individual file types, in case we add additional ones in the future.

ghost:
Right; there's a fair few instances of "mp4" / "webm" scattered around; I thought there would've been a column on Post for the media's type (image, video, or swf). Unfortunately, I’m not confident enough in Rails to add that myself 😅

Sindrake:
I kind of went down a rabbit hole when after this PR.
This is an okay stopgap measure, but in the long run we should go with the steps in #1345.

Sindrake (approved):

PR #1348, opened 2025-09-27
[PostEvents] Allow staff to see flag notes on Post Events

### LINKED ISSUE #1347: [PostEvent] Post events should show staff extra into
Post events should show staff extra info. Specifically, the `note` on post flags, like the `post_flags` index has.

While I'm in there, might as well move it to a `grid`.

mm12:
Closes #1347

Changes: 
- [x] Make `Flag Notes` into a partial
- [x] PostEvent index
  - [x] Make PostEvent table `grid`-based instead (Partially done, needs refinement)
  - [x] Add FlagNote partial to index
  - [x] Add fallback in view
    - [x] for if `reason` exists but the flag doesn't
    - [x] for if neither exist 
- [x] Change event's `extra_data`
  - [x] Post flag add
  - [x] Post deletion
  - [x] Remove old `reason`
- [x] DB Fix
  - [x] Add `flag_id` to `extra_data`, and remove `reason`. 
  - [x] Go by `PostFlag` to avoid issues where a `PostEvent` exists but the flag was expunged. 
- [x] Tests 
  - [x] Fix tests that rely on the view

Considerations: 
- `PostEvent` does not allow searching `extra_data`, so it *should* be fine to remove the reason. Since we are already loading `PostFlag` to get the `note`, we can take the reason from there instead.


app/javascript/src/styles/views/post_events/_post_events.scss
```diff
@@ -0,0 +1,121 @@
+:root {
```
Sindrake, line 1: You shouldn't be putting styles in the `specific` directory anymore.
I've been (semi-successfully) trying to phase it out in favor of the `views` directory, which allows more specificity rather than just dumping all styles into one folder.

mm12:
Since I thought it was a good idea to also move things to a partial when i did this, things are now out of sync and need real conflict resolving.

Sindrake:
Using grid on the post events page is viable, and is a good idea in the long run.
It would be moving the UI in the right direction.

Back-filling post flag data into the post events is only theoretically useful at this point.
We could come back to that in the future, if doing so actually solves some kind of problem we are having.

As is, this is not going to work.

PR #1351, opened 2025-09-29
[TagPreview] Add implication tooltips

### LINKED ISSUE #1350: Tag Preview does not show information about implications
The previous "final tags" button showed exactly which tags implied which (even though, in a very convoluted way).
While this information is present in the API response, it is not displayed in the UI. 

We should also consider how this change might need to adapt to mobile devices.

clragon:
Adds:
- Hover title showing which tags cause implications

This allows desktop users to access this information, but won't work on mobile. 
Fixes #1350 for now.


Sindrake (approved):
lgtm

PR #1352, opened 2025-09-29
[UI] Replace Jquery UI autocomplete

### LINKED ISSUE #1315: Replace Jquery UI autocomplete
Currently, our autocomplete component is provided by Jquery UI. 
Jquery UI is deprecated and must be removed. We should therefore pick a new way to implement autocomplete.

Since we do not want to increase the part of our codebase that is made up of Vue, and we would not like to reinvent the wheel, the most likely solution is to pick a library that provides us with autocomplete with pure web tech.

An example of such a library is [autocompleteJS](https://tarekraafat.github.io/autoComplete.js/).

clragon:
Replaces the Jquery UI autocomplete with a vanilla JS version. Fixes #1315.


app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+
+    if (!match) {
+      return {};
+    }
+
+    let term = match[0];
+    let metatag = "";
+
+    const prefixMatch = term.match(TAG_PREFIXES_REGEX);
+    if (prefixMatch) {
```
Sindrake: This seems to break autocomplete for negated metatags.
For example: `-rating:safe`.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+    const scrollTop = window.pageYOffset || document.documentElement.scrollTop;
+    const scrollLeft = window.pageXOffset || document.documentElement.scrollLeft;
+
+    this.dropdown.style.position = "absolute";
+    this.dropdown.style.left = (rect.left + scrollLeft) + "px";
+    this.dropdown.style.top = (rect.bottom + scrollTop) + "px";
+    this.dropdown.style.minWidth = rect.width + "px";
+  }
+
+  bindEvents () {
```
Sindrake, line 522: The fact that this is all vanilla JS makes me a little sad.
Non-namespaced events seem a little dangerous, especially since the autocompleted inputs oftentimes have _other_ scripts binding listeners to them.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+        name: `${metatag}:${option}`,
+        label: `${metatag}:${option}`,
+        category: "metatag",
+        type: "metatag",
+      }))
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .slice(0, 10);
+  },
+
+  async getMetatags (metatag, term) {
```
Sindrake, line 128: You seem to treat category prefixes (`artist:`, `character:`, etc) as metatags.
But there does not seem to be any handling of these here, which causes the autocomplete to return no results for them.

Few people are likely to use category prefixes in search, but they should still be handled for feature parity with the old autocomplete implementation.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+      case "commenter":
+      case "comm":
+      case "noter":
+      case "noteupdater":
+      case "fav":
+      case "favoritedby":
+      case "flagger":
+      case "upvote":
+      case "downvote":
+        return this.getUsers(term).then(results => results.map(user => ({
```
Sindrake: Previous implementation colored usernames based on the user's level.
This one, sadly, does not.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+    return data.map(wiki => ({
+      name: wiki.title,
+      label: this.formatLabel(wiki.title),
+      category: wiki.category_id,
+      type: "wiki_page",
+      id: wiki.id,
+    }));
+  },
+
+  async searchItems (query, dataFetcher, { minLength = 1, maxResults = 15 } = {}) {
```
Sindrake: Old minimum input length was 3.
I am not outright opposed to changing that, but attempting to autocomplete things based on only one or two letters seems silly.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+    if (!query.trim()) {
+      this.results = [];
+      this.selectedIndex = -1;
+      this.render();
+      this.close();
+      return;
+    }
+
+    try {
+      this.results = await this.searchFn(query, this.input);
```
Sindrake: It might make sense to check whether the input hasn't changed while we were waiting for the server to get back to us with autocomplete result. That way, we won't be serving outdated data.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+    this.input.addEventListener("keydown", this.handleKeydown.bind(this));
+    this.input.addEventListener("blur", this.handleBlur.bind(this));
+    this.input.addEventListener("focus", this.handleFocus.bind(this));
+
+    this.dropdown.addEventListener("mousedown", (e) => e.preventDefault());
+    this.dropdown.addEventListener("click", this.handleDropdownClick.bind(this));
+
+    this.repositionHandler = this.positionDropdown.bind(this);
+    window.addEventListener("scroll", this.repositionHandler);
+    window.addEventListener("resize", this.repositionHandler);
```
Sindrake: Every autocomplete instance adds two new global scroll and resize listeners. This seems unwise, especially on the upload page.

I would either create a shared listener (ex. see the dialog implementation) or only attach these listeners when the autocomplete dropdown is open.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+      if (index >= 0 && this.results[index]) {
+        this.selectItem(this.results[index]);
+      }
+    }
+  }
+
+  async search () {
+    const query = this.input.value;
+
+    if (!query.trim()) {
```
Sindrake: It might make sense to prevent sending duplicate requests by checking whether the current input value isn't exactly the same as the one searched last time around.

app/javascript/src/javascripts/autocomplete.js
Sindrake: Also, it may make sense to add some kind of minimum length for these kinds of metatag parameters.
Autocomplete based on only one or two (or zero!) letters isn't really useful for usernames.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+  },
+
+  formatLabel (text) {
+    return text.replace(/_/g, " ");
+  },
+
+  getHref (item) {
+    switch (item.type) {
+      case "user":
+        return "/users/" + item.id;
```
Sindrake: The user links always end up as `/users/undefined`.
Presumably, the ID gets lost somewhere.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,667 @@
...
+      case "ArrowDown":
+        event.preventDefault();
+        this.navigateDown();
+        break;
+      case "ArrowUp":
+        event.preventDefault();
+        this.navigateUp();
+        break;
+      case "Enter":
+        event.preventDefault();
```
Sindrake, line 566: I would also stop event propagation here, just in case.

Sindrake (changes requested):
Overall, pretty solid.
Found some minor issues.

app/javascript/src/javascripts/autocomplete.js
clragon: I investigated this and found out that `artist:some_artist` isn't a valid tag search for the backend, so by correctly replacing the suggestion after the tag category, we set up the user for a broken search. 

I have instead opted to just ignore the tag category and replacing it on enter. Since the autocomplete doesnt support searching by category, this was how it should have worked.

app/javascript/src/javascripts/autocomplete.js
clragon: I think writing this in vanilla js is actually quite nice, although I assume we might never get rid of Jquery, since it is much too useful. For that reason, this could have also used Jquery. If you want me to, I can update the appropriate spots, so that we are consistent everywhere.

app/javascript/src/javascripts/autocomplete.js
```diff
@@ -0,0 +1,751 @@
...
+    if (!currentQuery.trim()) {
+      this.results = [];
+      this.selectedIndex = -1;
+      this.query = "";
+      this.render();
+      this.close();
+      return;
+    }
+
+    if (currentQuery === this.query) return;
```
clragon:
```suggestion
    if (currentQuery.trim() === this.query.trim()) return;
```

Sindrake (approved):
lgtm

PR #1353, opened 2025-09-30
[Dialog] Minor dialog enhancements

jmortiger:
* Initial position definition is more flexible
* Improves handling of fullscreen toggles
* Stops dialog from going offscreen from resizing/fullscreen toggles
   * Stops dialog from sliding around when resizing unless it would exceed the containers bounds; can be toggled off for prior behavior, but the UI to do so is not there.

### DIFF
--- a/app/javascript/src/javascripts/utility/dialog.js
+++ b/app/javascript/src/javascripts/utility/dialog.js
@@ -1,11 +1,12 @@
 export default class Dialog {
 
-  // Container to which all dialogs are appended
+  /** Container to which all dialogs are appended */
   static _container = null;
   static containerWidth = 0;
   static containerHeight = 0;
-  // Used to store the current timeout
+  /** Used to store the current timeout */
   static _currentTimeout = null;
+  /** Container to which all dialogs are appended */
   static get container () {
     if (this._container !== null) return this._container;
 
@@ -13,12 +14,12 @@ export default class Dialog {
     this.updateContainerDimensions();
 
     // Window dimension changes
-    $(window).on("resize orientationchange", () => this.updateContainerDimensions());
+    $(window).on("resize orientationchange", this.onUpdateContainerDimensions);
 
     // Fullscreen changes
     $(document).on("fullscreenchange webkitfullscreenchange mozfullscreenchange MSFullscreenChange", () => {
       if (this._currentTimeout) clearTimeout(this._currentTimeout); // Stops quickly toggling fullscreen causing trouble
-      this._currentTimeout = setTimeout(() => this.updateContainerDimensions(), 100); // Small delay to ensure layout has settled
+      this._currentTimeout = setTimeout(this.onUpdateContainerDimensions, 100); // Small delay to ensure layout has settled
     });
 
     return this._container;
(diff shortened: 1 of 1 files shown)


app/javascript/src/javascripts/utility/dialog.js
```diff
@@ -246,6 +289,37 @@ export default class Dialog {
...
+      height: height,
+      left: xMin,
+      top: yMin,
+    });
+  }
+
+  _isPinned = true;
+  /** True if the dialog is currently pinned */
+  get isPinned () { return this._isPinned; }
+  /** If changed to true, will trigger an update */
```
Sindrake: The function does the opposite of what the comment says.
An update is triggered if the value is set to false.

app/javascript/src/javascripts/utility/dialog.js
```diff
@@ -160,33 +194,42 @@ export default class Dialog {
...
+      && this.xMin >= 0 && this.xMax <= _max.x
+      && this.yMin >= 0 && this.yMax <= _max.y)
+      return;
+
+    const positionDef = this.currentNormalizedPosition;
+
+    const projectedPosition = {
+      x: (_max.x) * positionDef[0],
+      y: (_max.y) * positionDef[1],
+    };
```
Sindrake: Dead variable?

app/javascript/src/javascripts/utility/dialog.js
```diff
@@ -246,6 +289,37 @@ export default class Dialog {
     else this.open();
   }
 
+  _xMin = null;
+  _yMin = null;
+  get xMin () { return this._xMin == 0 ? this._xMin : this._xMin ||= Number(this.$dialog.css("left")); }
+  get yMin () { return this._yMin == 0 ? this._yMin : this._yMin ||= Number(this.$dialog.css("top")); }
```
Sindrake: Slightly confusing ternary. I would recommend adding parentheses to clarify the order of operations.

More importantly, `this.$dialog.css("left")` and `this.$dialog.css("top")` don't return numbers.
They return strings like `128px`.

Unless I am missing something, `Number("128px")` will return NaN, which is not what we want.

app/javascript/src/javascripts/utility/dialog.js
```diff
@@ -1,21 +1,27 @@
...
+  /** Container to which all dialogs are appended */
   static get container () {
     if (this._container !== null) return this._container;
 
     this._container = $("<div id='dialog-container'>").appendTo("body");
     this.updateContainerDimensions();
 
     // Window dimension changes
-    $(window).on("resize orientationchange", () => this.updateContainerDimensions());
+    $(window).on("resize orientationchange", this.onUpdateContainerDimensions);
```
Sindrake: We should probably namespace these events, just in case.

Sindrake (changes requested):

app/javascript/src/javascripts/utility/dialog.js
jmortiger: I'm not sure if you're referring to the events listened to or the callback. For the former, this was already like this & these aren't custom events. For the latter, I'll update it.

app/javascript/src/javascripts/utility/dialog.js
jmortiger: I was thinking of parseInt; replaced. The only reason for this ternary is that valid value of `0` is falsy; in all other cases, if the current value is falsy, assign it from the CSS value; otherwise, return its current value.

app/javascript/src/javascripts/utility/dialog.js
Sindrake: For the events listened.
It's my mistake, I forgot to do it in the original implementation.

```JS
$(window).on("resize.dialog orientationchange.dialog", () => this.onUpdateContainerDimensions());
```

Sindrake (approved):
lgtm

PR #1362, opened 2025-10-01
[Misc] Fix devcontainers and codespaces

### LINKED ISSUE #1360: Fix the development container
The development container does not work properly in many scenarios: 
- Codespaces: Fails to set up
- Desktop: Sometimes works for a while, but then becomes irrecoverable. 

Severely gets in the way of development in VSCode-based environments without having workarounds.

mm12:
Closes #1360. 

I recommend using another terminal window with WSL anything that involves git (fetch, pull, commit, push, etc.), as there is an incompatibility between the recommended WSL image (Ubuntu 22) and the image for e621ng (Alpine). This causes issues with SSH and GPG, preventing them from working properly.

These changes fix devcontainers so they do not crash constantly. The primary issue was seemingly the `dev-container` profile, which told it not to restart the services. I believe this prevented the container from attaching to the ports it needed to. However, because they were removed, starting and stopping the container is linked with docker. If you exit the container and would like to keep the container running, you will need to run `docker compose up` again. 

In addition to the fix, I have added some tooling that developers may find helpful. 
The first is tasks. These do not always work, but they allow a quick way to run common commands when not using the dev container. 
The second is the addition of extensions that may be useful in development to the devcontainer. Those that are not already included are commented out. One of these extensions is StyleLint, which is a linting tool for CSS and SCSS. Usage of this extension should eventually become redundant, since ESLint has a plugin for it.


mm12:
Current investigation:

- [x] while `/packs/static/` and `images/counter` contents work correctly, `/data/*` does not. mounted to devcontainer in it's `public/data` ... hm [REVERSION]
- [-] rubocop isn't using project config https://github.com/Shopify/ruby-lsp/issues/1738 (WONTFIX)
- [x] Fix login: `ActionController::InvalidAuthenticityToken. Did you properly authorize your request?` (CORS)

Sindrake (dismissed):
Looks great to me.

One small point of contention: one of the tests fails when running through VSCode.
```
Failure:
PostTest#test_: URLs: should generate the correct urls for animated gifs.  [test/unit/post_test.rb:2324]:
--- expected
+++ actual
@@ -1 +1 @@
-"www.example.com/data/preview/deadbeef.jpg"
+"https://77723c80fbaf/data/preview/deadbeef.jpg"
```

mm12:
That is very strange, since the `default_base_url` should only do that under `Rails.env.development?`, not `Rails.env.test?`. I'll look into it.

.vscode/tasks.json
```diff
@@ -0,0 +1,120 @@
...
+        "isDefault": true
+      },
+    },
+    {
+      "label": "e621: stop",
+      "type": "shell",
+      "command": "docker compose down",
+      "detail": "Stop the application",
+      "group": "build",
+    },
```
Sindrake, line 50: Both of these commands fail for me.

```
Executing task: docker compose up -d 

unknown shorthand flag: 'd' in -d
See 'docker --help'.
```
```
Executing task: docker compose down 

docker: 'compose' is not a docker command.
See 'docker --help'
```

.devcontainer/devcontainer.json
```diff
@@ -5,27 +5,45 @@
...
-    "git-safe-dir": "git config --global --add safe.directory /app"
-  },
+  "mounts": [
+  //   "source=/home/${localEnv:USER}/.gnupg,target=/home/vscode/.gnupg,type=bind,consistency=cached",
+    "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind"
+  ],
+  "postAttachCommand": ["/bin/sh", "/app/.devcontainer/post-attach.sh", "bash"],
   "postCreateCommand": {
-    "bash-completion": "sudo apk --no-cache add bash-completion && echo 'source /etc/bash/bash_completion.sh' >> ~/.bash_profile"
+    "bash-completion": "apk update && apk --no-cache add bash-completion docker && echo 'source /etc/bash/bash_completion.sh' >> ~/.bash_profile"
```
Sindrake: Do we need to have the full docker engine in the devcontainer?
It seems like `docker-cli` would work just as well, since it's only used to talk to the host's socket.

Sindrake (changes requested):

.devcontainer/devcontainer.json
mm12: Currently trying to figure out why using `docker-cli-compose` makes it so that the main service can't find the `Procfile` even if I directly mount it.

.vscode/tasks.json
clragon: You likely have the old `docker-compose` rather than the new `docker compose`.

mm12: If `docker-cli-compose` is installed, even restarting the service using Docker Desktop fails.

Sindrake (approved):

PR #1378, opened 2025-10-03
[Posts] Update the score in the sidebar when voting

### LINKED ISSUE #1377: Score under Information in the sidebar isn't updated when upvoting/downvoting
When upvoting/downvoting, the post's current score is updated in the voting block, but the Score in the sidebar under Information is not. It'd be nice if it was.

ghost:
Closes #1377.

Also, in adding this I noticed ESLint isn't linting all of the JS in tree. For whatever reason the glob `./app/javascript/src/**/*.js` doesn't end up including files in sub directories; you can see this when looking at test results when ESLint dumps all of the filenames its linting. `PostsShowToolbar.js` actually doesn't pass linting because of the unused `LStorage` import, but I thought I'd leave it to be fixed elsewhere.

### DIFF
--- a/app/javascript/src/javascripts/views/PostsShowToolbar.js
+++ b/app/javascript/src/javascripts/views/PostsShowToolbar.js
@@ -85,6 +85,13 @@ export default class PostsShowToolbar {
 
   static async vote (direction) {
     return PostVote.vote(PostsShowToolbar.currentPost.id, direction).then((data) => {
+      // Update Score in Information
+      $(".post-score").text(data.score)
+        .removeClass("score-negative score-neutral score-positive")
+        .addClass(data.score > 0
+          ? "score-positive"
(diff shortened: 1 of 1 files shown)


ghost:
Commit naming, right, sorry sorry :)

Sindrake:
Interesting.

I don't know if having the score duplicated in the sidebar is even necessary. It does not really serve any purpose.
I left that section alone because we needed to have a link to the post votes somewhere in there.

The code you've got looks and works okay, I am just undecided on whether we need this.

ghost:
Sure. Yeah, the Information section certainly isn't perfect; it shows the post ID, which is always visible in the URL, but doesn't show the MD5, which is almost impossible to find on deleted posts (from what I can find, it's only visible as part of the URL of the Inkbunny reverse search link). That irks me too.

Sindrake (approved):

PR #1379, opened 2025-10-03
[Misc] Fix ESLint not linting JS files in subdirectories

ghost:
As mentioned in #1378, ESLint currently isn't linting any JS files in subdirectories of `javascript/src`. This is because the provided glob argument is unquoted, which means Bash is expanding it instead of Node, and Bash, by default, does not recurse `**` into subdirectories. The fix is then to quote the argument so that Node expands it instead.

Additionally, I fixed all of the newly reported warnings. This involved changing the ESLint config, adding `{ "caughtErrorsIgnorePattern": "^_" }` to the `no-unused-vars` rule because there are two places in `app/javascript/src/javascripts/models/Favorite.js` where `catch` is used without using the error, but it nonetheless _must_ be captured (`try { } catch { }` is invalid JS).

### DIFF
--- a/app/javascript/src/javascripts/models/Favorite.js
+++ b/app/javascript/src/javascripts/models/Favorite.js
@@ -10,9 +10,9 @@ export default class Favorite {
    */
   static async create (post_id) {
     if (!post_id) return Promise.reject(new Error("Post ID is required"));
-    
+
     return TaskQueue.add(async () => {
-      return fetch(`/favorites.json`, {
+      return fetch("/favorites.json", {
         method: "POST",
         headers: {
           "Content-Type": "application/json",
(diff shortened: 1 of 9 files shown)


ghost:
Forgot to push the ESLint config changes, oops

Sindrake:
Well, this was a silly mistake.
Thanks for catching it.

Coincidentally, do you have a e621 account?
I would like to credit you properly in the changelogs, if possible.

ghost:
> Well, this was a silly mistake. Thanks for catching it.
> 
> Coincidentally, do you have a e621 account? I would like to credit you properly in the changelogs, if possible.

Sure, <https://e621.net/users/2248328>. Thanks!

PR #1382, opened 2025-10-03
[Posts] Move comment blacklist and edit buttons

### LINKED ISSUE #1381: [Post] Move UI elements for a less confusing UX
1. Move the edit button from being next to comments to being in the toolbar. 
2. Move comments blacklist to being near the threshold item

Image

mm12:
Closes #1381 

Edit button in toolbar: 
Toolbar

Edit button moves to dropdown on small screens:
Toolbar dropdown

Blacklist location with no hidden comments:
image

Blacklist location with hidden comments:
image


Sindrake (changes requested):
The "Edit" button does not work.

It should also be less prominent.
Putting it between "Favorite" and "Change Size" will result in people accidentally editing tags. I would put it after "View":

Image

Keep in mind, people will send us death threats for adding another step to editing tags on mobile.
But that's just something we may have to live with.

mm12:
pretty sure it broke when the JS changed, this should fix it

Sindrake (approved):

PR #1383, opened 2025-10-04
Rename hide to delete

### LINKED ISSUE #1358: Rename "Hide" to "Delete"
Some of our UI elements feature the text "hide" for what is essentially a soft delete operation, and "delete" for true delete.
Others correctly use "delete" and use "destroy" for true delete.

We would like to normalize the latter, as it is more intuitive for both staff and users.

clragon:
Renames hide to delete, and true delete to destroy. everywhere. 

Fixes #1358


Sindrake:
This is unacceptable and unreviewable for reasons already discussed.

PR #1390, opened 2025-10-08
[DMails] Change default response text

jmortiger:
Adds link to original message & wrapped quoted message in a section tag


Sindrake:
Not a bad idea in principle, but looks extremely ugly.

Screenshot 2025-10-08 084950

jmortiger:
Would it be ok w/o the original message link? I'd assume that'd be at least on par with the current implementation.

Sindrake:
Yes, I think so.

PR #1393, opened 2025-10-08
[ESLint] Enable ESLint formatting in VS Code

jmortiger:
The setting is off by default, so it only uses the JS/TS language support for formatting by default. https://discordapp.com/channels/431908090883997698/1393442162305138758/1425476601508003910

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -21,4 +21,10 @@
     "javascript"
   ],
   "eslint.format.enable": true,
+  "[javascript]": {
+  	"editor.defaultFormatter": "dbaeumer.vscode-eslint"
+  },
+  "[typescript]": {
+  	"editor.defaultFormatter": "dbaeumer.vscode-eslint"
+  },
 }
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -20,4 +20,5 @@
   "eslint.validate": [
     "javascript"
   ],
+  "eslint.format.enable": true,
```
Sindrake, line 23: I believe that you should also set the default formatter to ESLint, to avoid getting a popup regarding having multiple active formatters.
```
"editor.defaultFormatter": "dbaeumer.vscode-eslint",
```

Sindrake (changes requested):

.vscode/settings.json
jmortiger: I've scoped it for JS & TS; don't want to find out if it interferes w/ anything else.

Sindrake (approved):
lgtm

PR #1400, opened 2025-10-10
[HelpPages] Display a default help page in place of an index

jmortiger:
1st part of the [Wiki Help Index Rebuild](<https://discordapp.com/channels/431908090883997698/1420143660833443840>).


Sindrake:
Slightly revamped approach.

The `/help` route now always tries to fetch the default landing page for everyone.
The old help index now lives at `/help/list`.

Sindrake (approved):

PR #1404, opened 2025-10-12
[Post] Use post_path for md5 redirect

mm12:
Use `post_path` wrap on the post object when using the `md5` parameter for posts.


Sindrake (commented):
Any particular reason for this?
It seems to work fine as is.

PR #1405, opened 2025-10-12
[Wiki] Use wiki path to link to artist page 

mm12:
Wrap the link to the artist page with the `show_or_new_artists_path` rather than linking directly to the value of the artist name.


Sindrake (approved):
Looks good to me.
Consistency with the [other "View artist" link](https://github.com/e621ng/e621ng/blob/master/app/views/wiki_pages/show_or_new.html.erb#L14).

PR #1408, opened 2025-10-14
[Search] Reveal all fields auto-hidden when empty to staff

jmortiger:
For staff work, hiding this stuff is maddening. I don't remember where, but spe asked if we could stop doing this, at least for ids.
image


Sindrake (approved):
I feel like seeing the massive forms everywhere might get annoying.
Let's see if staff members complain, I suppose.

PR #1417, opened 2025-10-16
[DText] Allow for the usage of a local dtext_rb gem

Sindrake:


docker-entrypoint.sh
mm12, line 1: Doesn't appear to have the executable flag

PR #1436, opened 2025-10-23
[Utility] Added generic collapsible elements

jmortiger:
Adding a class of `st-collapsible-header` makes it behave like the collapsible tag lists in the sidebar of the post search/view pages. Can have a default arrow indicator that animates by adding an `indicator` attribute of value `arrow`, & can animate the collapse with an `animated` attribute. The accompanying callback triggers the state change by adding a `collapsed` attribute.

### DIFF
--- a/app/javascript/src/javascripts/utility.js
+++ b/app/javascript/src/javascripts/utility.js
@@ -73,6 +73,33 @@ Utility.regexp_escape = function (string) {
   return string.replace(/([.?*+^$[\]\\(){}|-])/g, "\\$1");
 };
 
+/**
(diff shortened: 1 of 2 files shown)


Sindrake (commented):
Do you have plans to use this for something, or is it just a nice thing to have?

jmortiger:
Oh yeah, I was gonna put this on other elements in the sidebar as a consolation for putting the Edit button back; letting it be collapsed to shrink it back down. Although it's not persistent (at least not right now).

...I just forgot to actually do that.

PR #1444, opened 2025-10-27
[Users] Fix overflow issues on the mobile version of the upload page

KernelEquinox:
This change fixes an issue with the upload limit explanation rendering on mobile devices.

Additionally, it now groups each section of the upload limit formula by operation, that way any wrapping will occur such that the mathematical operator is always at the beginning of the new line.

Prior to operation grouping, it would be possible for the formula to wrap as shown:
```
10 + (0 / 10) - (0 / 4) -
0 = 10
```
Note that the wrapped line starts with the number and not the mathematical operator.

With operation grouping, that same wrapping formula will now result in the operator leading the new line:
```
10 + (0 / 10) - (0 / 4)
- 0 = 10
```


KernelEquinox:

PR #1207, opened 2025-08-03
[Replacements] Mobile Replacements 

mm12:
Add a mobile-friendly replacement UI, along with additional functionality for a better user experience. This is in preparation for exiting beta. Due to the sheer number changes, there is likely several issues that have slipped by despite testing. Please report them. Fixes #1389.

> [!IMPORTANT]
> This PR is to have individual changes added through smaller PRs before the "big merge". A PR reference will be added to items as this happens. These PRs are **BLOCKING** for this PR, and should all be merged beforehand. Internal refs: [[1]](https://discord.com/channels/431908090883997698/1393442162305138758/1428766298464194560) [[2]](https://discord.com/channels/431908090883997698/1393442162305138758/1428827278682361856)

### New functionality
- Add ability to transfer replacements to another post -- #1451 
  - Only works if post if pending or rejected.
  - Fails if the new replacement has an existing md5 match
  - Adds a new PostEvent to both the old and new posts
- Add "Silent Approval" (subject to change) -- #1449
  - If an approver submits a replacement, they have the option to approve it without taking upload credit
  - The replacement creator remains the approver
- Add ability for any staff to add notes to replacements -- #1450 
  - Notes currently only visible to the replacement's creator and staff
  - Only one note on a post at a time
  - Notes are a new table, and model, belonging to replacements, to avoid mostly-empty columns
  - Notes create a mod actions log
- Replacements can now be reported, allowing tickets against them -- #1447 
- Misc -- #1448 
  - Secondary links added to replacements (just "Listing")
  - Replacements show if the creator is the artist
### Changed
- Replacements upload page now shows errors that don't overflow on small screens
- Maximum reason length increased to 300 characters
- Mobile UI 
  - Replacements now use a `grid` instead of `table`. Each replacement has an expandable card. 
  - Cards can be expanded/collapsed, and have a compact style for mobile screens. 
  - These are automatically transitioned between. On mobile, cards start collapsed. 
  - Larger buttons and clear status indicators make mobile interaction easier
- The mobile breakpoint was moved to a global constant, and acts as the default site-wide
### Fixes
- Better error handling
  - More controllers now check for errors before returning html, and return JSON if an error occurs
  - If an HTML error page is returned, JS will attempt to get the error information, instead of displaying the HTML
  - source list being `nil` no longer causes view to fail
- Broken `test-large` jpeg replaced with non-corrupted jpeg

PR #1449, opened 2025-10-28
[Replacements] Add ability to approve replacments 'silently'

mm12:
> [!IMPORTANT]
> This is PR is a subset of #1207. This PR should be merged _before_ it.
> Degradation info: 
> ☑️ Functions: Appears to function correctly, but some jank may occur. 
> ✅ UX: UI appears as it should. 

image

- Add "Silent Approval" (subject to change)
  - If an approver submits a replacement, they have the option to approve it without taking upload credit
  - The replacement creator remains the approver


clragon (changes requested):
Right now, this doesnt allow staff to upload a replacement that is "silent" from the get go. It requires uploading the replacement as pending, then approving it "silently". 

Maybe a checkbox for replacing silently could be added for that purpose.

mm12:
> Right now, this doesnt allow staff to upload a replacement that is "silent" from the get go. It requires uploading the replacement as pending, then approving it "silently".
> 
> Maybe a checkbox for replacing silently could be added for that purpose.

Because this PR will ultimately be overwritten by the primary PR, any feature changes need to either be in both, or be in a new PR to apply after them.

app/controllers/post_replacements_controller.rb
```diff
@@ -58,17 +61,38 @@ def create
...
+    end
+
+    if @post_replacement.errors.any?
+      respond_to do |format|
+        format.json { return render json: { success: false, message: @post_replacement.errors.full_messages.join("; ") }, status: 412 }
+        format.html { return render plain: "Replacement approval failed: #{@post_replacement.errors.full_messages.join('; ')}", status: 400 }
+      end
+    end
 
     if @post_replacement.errors.any?
```
Sindrake: This code us unreachable because of the block above.

app/controllers/post_replacements_controller.rb
```diff
@@ -58,17 +61,38 @@ def create
...
 
-    @post_replacement.approve!(penalize_current_uploader: params[:penalize_current_uploader])
+    approve_options = {}
+    if params.key?(:penalize_current_uploader)
+      approve_options[:penalize_current_uploader] = ActiveModel::Type::Boolean.new.cast(params[:penalize_current_uploader])
+    end
+    approve_options[:credit_replacer] = ActiveModel::Type::Boolean.new.cast(params[:credit_replacer]) if params.key?(:credit_replacer)
+
+    begin
+      @post_replacement.approve!(**approve_options)
```
Sindrake, line 85: The `approve!` method has `penalize_current_uploader:` as a required keyword.
If `approve_options[:penalize_current_uploader]` is not set, this will throw `ArgumentError: missing keyword`.

app/logical/upload_service/replacer.rb
```diff
@@ -59,8 +59,10 @@ def process!(penalize_current_uploader:)
         post.source = "#{replacement.source}\n" + post.source
         post.tag_string = upload.tag_string
         # Reset ownership information on post.
-        post.uploader_id = replacement.creator_id
-        post.uploader_ip_addr = replacement.creator_ip_addr
+        if credit_replacer.to_s.truthy?
```
Sindrake:
```suggestion
        if credit_replacer
```
`credit_replacer` arrives from `approve!` as a boolean.
Why are you casting it to string, and then checking `.truthy?`

app/controllers/post_replacements_controller.rb
```diff
@@ -58,17 +61,38 @@ def create
...
+    rescue ProcessingError => e
+      @post_replacement.errors.add(:base, e.message)
+    end
+
+    if @post_replacement.errors.any?
+      respond_to do |format|
+        format.json { return render json: { success: false, message: @post_replacement.errors.full_messages.join("; ") }, status: 412 }
+        format.html { return render plain: "Replacement approval failed: #{@post_replacement.errors.full_messages.join('; ')}", status: 400 }
```
Sindrake: Why does it throw error 412 in JSON and 400 in HTML?

app/views/post_replacements/partials/show/_post_replacement.html.erb
```diff
@@ -103,7 +103,10 @@
     <% if CurrentUser.can_approve_posts? %>
       <div class="pending-links">
         <% if post_replacement.status == "pending" %>
-          <% if !post_replacement.post.is_deleted? %> <%= link_to "Approve", "#approve", class: "replacement-approve-action", data: { replacement_id: post_replacement.id, penalize: post_replacement.post.uploader != post_replacement.creator } %> <% end %><br />
+          <% if !post_replacement.post.is_deleted? %>
+            <%= link_to "Approve", "#approve", class: "replacement-approve-action", data: { replacement_id: post_replacement.id, penalize: post_replacement.post.uploader != post_replacement.creator } %><br>
+            <%= link_to "Silently", "#silent-approve", class: "replacement-silent-approve-action replacement-action-button", data: { replacement_id: post_replacement.id } %><br><br>
+          <% end %><br />
```
Sindrake: Nitpick: stray `<br />` after `<% end %>`, and the inconsistency of `<br>` vs `<br />`.

app/controllers/post_replacements_controller.rb
mm12: `penalize_current_uploader` is set on `create`, if it isn't available here, is it wrong to error instead? What should it default to?

Sindrake: I would just add a default value to it in the `approve!` method.

PR #1451, opened 2025-10-28
[Replacements] Add ability to transfer replacements between posts

mm12:
> [!IMPORTANT]
> This is PR is a subset of #1207. This PR should be merged _before_ it.
> Degradation info: 
> ☑️ Functions: Appears to function correctly, but some jank may occur. 
> ☑️ UX: UI appears correctly, but may break in some circumstances. 

- Add ability to transfer replacements to another post
  - Only works if post if pending or rejected.
  - Fails if the new replacement has an existing md5 match
  - Adds a new PostEvent to both the old and new posts


app/models/post_replacement.rb
```diff
@@ -270,6 +270,38 @@ def reject!
...
+      set_previous_uploader
+      update_column(:uploader_id_on_approve, uploader_on_approve&.id)
+      create_original_backup
+
+      PostEvent.add(post.id, CurrentUser.user, :replacement_moved, { replacement_id: id, old_post: prev.id, new_post: post.id })
+      PostEvent.add(prev.id, CurrentUser.user, :replacement_moved, { replacement_id: id, old_post: prev.id, new_post: post.id })
+
+      post.update_index
+      prev.update_index
+    end
```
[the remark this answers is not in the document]
mm12: Transferring them should be able to be done without setting the status to pending - ie, if you want to reject it but it is on the wrong post. This way, you can either reject then move, or move then reject - it doesn't matter. If you want to approve it, you can just use "reset to".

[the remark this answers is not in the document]
mm12: It is still rejected, as mentioned.

app/views/post_replacements/partials/show/_post_replacement.html.erb
```diff
@@ -104,6 +104,7 @@
           <%= link_to "Approve", "#approve", class: "replacement-approve-action", data: { replacement_id: post_replacement.id, penalize: post_replacement.post.uploader != post_replacement.creator } %><br />
           <%= link_to "Compare", moderator_post_diff_path(post_a: post_replacement.post_id, post_b: post_replacement.md5) %><br /><br />
           <%= link_to "Reject", "#reject", class: "replacement-reject-action", data: { replacement_id: post_replacement.id } %><br />
+          <%= link_to "Transfer", "#transfer", class: "replacement-transfer-action replacement-action-button", data: { replacement_id: post_replacement.id } %>
```
Sindrake: This link is inside the `<% if post_replacement.status == "pending" %>` block, so it will never be shown on rejected posts.

app/javascript/src/javascripts/post_replacement.js
```diff
@@ -33,6 +33,60 @@ PostReplacement.initialize_all = function () {
...
+  if (rawNewPostId === null) {
+    Utility.notice("No post ID specified. Transfer cancelled.");
+    return;
+  }
+  const newPostId = rawNewPostId.trim();
+  if (!newPostId || isNaN(Number(newPostId))) { // ensure it's a valid number
+    Utility.notice("Invalid post ID. Transfer cancelled.");
+    return;
+  }
+  if (!newPostId) {
```
Sindrake: Dead code - already handled by the block right above it.

app/models/post_replacement.rb
```diff
@@ -270,6 +270,40 @@ def reject!
...
+        errors.add(:md5, "duplicate of existing replacement on post ##{new_post.id}")
+        return
+      end
+
+      Post.transaction do # wrap in transaction to ensure atomicity of the transfer
+        prev = post
+        update(post: new_post, uploader_id_on_approve: nil)
+        set_previous_uploader
+        update_column(:uploader_id_on_approve, uploader_on_approve&.id)
+        create_original_backup
```
Sindrake: The `create_original_backup` method may throw a `ProcessingError` if it fails to create a backup.
As it stands, this error will bubble up to the controller and just return a 500.

Consider rescuing it, and returning a proper error response.

Sindrake (changes requested):

app/views/post_replacements/partials/show/_post_replacement.html.erb
mm12: Intended to be in the next block down, fixed

app/models/post_replacement.rb
```diff
@@ -278,6 +278,45 @@ def reject!
...
+        prev = post
+        update(post: new_post, uploader_id_on_approve: nil)
+        set_previous_uploader
+        update_column(:uploader_id_on_approve, uploader_on_approve&.id)
+
+        begin
+          create_original_backup
+        rescue ProcessingError => e
+          errors.add(:base, "Failed to create backup on new post: #{e.message}")
+        end
```
Sindrake: If a backup failed for any reason, this doesn't roll back the entire transaction.
The replacement is still moved, the PostEvents are logged, and the indexes are updated.

And yet, the controller sees `errors.any?` and returns 412.
The user is told that this action failed while it actually succeeded - and the new post now has a pending replacement without a backup.

```suggestion
        begin
          create_original_backup
        rescue ProcessingError => e
          errors.add(:base, "Failed to create backup on new post: #{e.message}")
          raise ActiveRecord::Rollback
        end
```

Tentative suggestion, untested.

spec/models/post_replacement/processing_methods_spec.rb
```diff
@@ -201,4 +201,119 @@ def stub_approve_deps(replacement)
...
+
+    it "add an error if backup creation fails during transfer" do
+      allow(replacement).to receive(:create_original_backup).and_raise(ProcessingError.new("boom"))
+
+      expect do
+        replacement.transfer(post_alt)
+      end.not_to(change { post_alt.replacements.where(status: "original").count })
+
+      expect(replacement.errors.full_messages).to include("Failed to create backup on new post: boom")
+    end
```
Sindrake, line 334: Doesn't actually test that the `replacement.post_id` stayed the same.
It should.

app/models/post_replacement.rb
Sindrake: Actually, `create_original_backup`'s `throw :abort` might escape this entirely, since that raises `UncaughtThrowError` rather than `ProcessingError`.

All in all, needs testing.

app/models/post_event.rb
```diff
@@ -29,6 +29,7 @@ class PostEvent < ApplicationRecord
     changed_bg_color: 21,
     replacement_penalty_changed: 24,
     owner_changed: 25,
+    replacement_moved: 26,
```
Sindrake, line 32: Missing from the `post_event_decorator.rb`.

app/models/post_replacement.rb
```diff
@@ -278,6 +278,45 @@ def reject!
...
+
+        PostEvent.add(post.id, CurrentUser.user, :replacement_moved, { replacement_id: id, old_post: prev.id, new_post: post.id })
+        PostEvent.add(prev.id, CurrentUser.user, :replacement_moved, { replacement_id: id, old_post: prev.id, new_post: post.id })
+
+        post.update_index
+        prev.update_index
```
Sindrake: I would do this outside the transaction, like how `promote!` does.

Sindrake (dismissed):

mm12:
Yay a 3rd review, will try to take a look this weekend if i get a chance

Sindrake (dismissed):
Fine

PR #1461, opened 2025-11-02
[Static] Rework Sitemap

mm12:
In preparation for scoped permissions, this PR reworks the `/static/site_map` page. Instead of putting conditions inline in the ERB, this switches to building lists of items based on the user's roles.

All permissions view:
image

Anonymous view:
image


mm12:
I'm probably forgetting one of those pages everyone forgets exists

Sindrake (approved):

PR #1492, opened 2025-11-06
[Misc] Add docker-cli-compose to devcontainers

mm12:
It didn't work before, but now it does. Now, you can use 'docker compose' or 'docker-compose' commands right in the devcontainer terminal! 'tasks.json' *could* be updated to use 'compose' commands again, but it is probably slightly better for them to run directly anyway.


Sindrake:
Is this still supposed to be a draft, or is it ready for review?

mm12:
Honestly, I have no idea, its been 4 months.

Sindrake:
Well, it's been a draft for 4 months so

PR #1504, opened 2025-11-11
Make it more obvious when reporting comment/blip/forum post

WaydenceMullins:
Float report button to the right, reduce its font size, change abuse warning background to red to make it more obvious to users that they're reporting and not replying


jmortiger:
Just to add an example of (some of) the effect from your forum post; we'll take a more thorough look when we get a chance. https://imgur.com/wpEvjbI

PR #1552, opened 2025-11-20
[UserNameChangeRequests] Rework the name change process

Sindrake:
The manual approval of name change requests was phased out years ago.
At this point, we might as well streamline the process a little.

As a little bonus, allowed users to change the capitalization in their names directly.


Sindrake:
One day, I'll manage to get this thing back in order.
But I keep coming back to it at 9pm on a Tuesday, so it keeps not happening.

PR #1580, opened 2025-12-20
[Misc] Move overmind lockfile to temp

clragon:
Moves the overmind lockfile into the tmp directory, which is inside the container, and therefore not mirrored back to the host, preventing weirdness when starting the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,4 +54,4 @@ RUN addgroup --gid ${HOST_GID} e621ng && \
 RUN git config --global --add safe.directory $(pwd)
 
 ENTRYPOINT ["/app/docker-entrypoint.sh"]
-CMD ["overmind", "start"]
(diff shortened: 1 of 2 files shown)


Sindrake:
This PR confounds me.

It's a little behind master, and dependency changes prevent the app from starting.
Rebased it, now the app actually starts – but I cannot force push into this branch ("permission denied").
Somehow, the executable bit on `docker-entrypoint.sh` is busted?

And even with all that aside, the `.overmind.sock` file is still being placed in the root directory, not in `/tmp`.

Sindrake:
Good news: the `.overmind.sock` file is now properly in the `/tmp` directory.
Bad news:

Screenshot 2026-02-03 225656

Sindrake:
I cannot get this to work.
The current solution works fine.

PR #1584, opened 2026-01-13
[Posts] Normalise `submissionview.php` Inkbunny sources to `/s/` sources

Parcly-Taxel:
Although the `submissionview.php` sources still work, Inkbunny generates the `/s/` sources in most places in its user interface, and the latter are much shorter.


Parcly-Taxel:

PR #1588, opened 2026-01-18
User preference for automatic upvote when favouriting

### LINKED ISSUE #1795: Auto-Like on Favorite
Upvote posts when favorites. Perhaps as an option that can be turned off. 

References: 
[[1]](https://e621.net/forum_topics/26915) [[2]](https://e621.net/forum_topics/54373) Ask for this feature
[[3]](https://e621.net/forum_topics/41668) Asked because this isn't implemented
[[4]](https://e621.net/forum_topics/59891) "As an option that can be toggled"
Mentioned in discord a few times, can't find it

Pending PR to close

pupgray:
Closes #1795

Forum discussion from a while back: https://e621.net/forum_topics/59891

A few other booru-like boards automatically record an upvote when faving a post. That discussion points at it being useful for most folks, but for people who use favourites as a "see later" list it could be a bit annoying. Seems like a viable user preference with the default being the current behaviour?

It's implemented in the FavoriteManager service object in ruby, just because I assume API favourites + other ways of saving a favorite should also follow this preference. It could just as easily be implemented in the FE though, if you'd prefer!

### Preference in Settings > Basic > Posts
Capture d’écran, le 2026-01-18 à 17 38 13

### And here it is when enabled

> [!IMPORTANT]
> For the moment, there's no visual UI changes on the post page. Wanted everyone's thoughts on that before I give it a go. Discoverability is pretty low when it's a toggle at the bottom of settings.
> Thoughts on including a small dropdown menu on the fave button itself? Would be a similar popover implementation to the search settings.
> exampleflow

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -165,6 +165,7 @@ class UsersController < ApplicationController
       enable_auto_complete
       enable_safe_mode disable_responsive_mode
       forum_notification_dot
+      enable_combined_upvote_and_favorite
     ]
 
     permitted_params += [dmail_filter_attributes: %i[id words]]
--- a/app/javascript/src/javascripts/views/PostsShowToolbar.js
+++ b/app/javascript/src/javascripts/views/PostsShowToolbar.js
@@ -82,27 +82,32 @@ export default class PostsShowToolbar {
     });
   }
 
-  static async vote (direction) {
-    return PostVote.vote(PostsShowToolbar.currentPost.id, direction).then((data) => {
-      // Update Score in Information
-      $(".post-score").text(data.score)
-        .removeClass("score-negative score-neutral score-positive")
-        .addClass(data.score > 0
-          ? "score-positive"
-          : (data.score < 0 ? "score-negative" : "score-neutral"));
+  /**
+   * @param data {{up: Number, down: Number, score: Number, our_score: Number}}
+   */
+  static updateVoteButtons (data) {
+    // Update Score in Information
+    $(".post-score").text(data.score)
+      .removeClass("score-negative score-neutral score-positive")
+      .addClass(data.score > 0
+        ? "score-positive"
+        : (data.score < 0 ? "score-negative" : "score-neutral"));
 
-      // Update button states for the current voting block.
-      $(".ptbr-score").text(data.score);
-      $(".ptbr-breakdown").html(`<span>${data.up}</span><span>${data.down}</span>`);
-      $(".ptbr-vote").attr({
-        "data-score": data.score,
-        "data-up": data.up,
-        "data-down": data.down,
-        "data-state": data.score > 0 ? 1 : (data.score < 0 ? -1 : 0),
-        "data-vote": data.our_score,
-      });
-      return data;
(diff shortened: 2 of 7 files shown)


jmortiger:
This is very nice work, thank you! And thanks for being sure to add tests; it's a big help when it comes to us reviewing these & for avoiding breaking changes down the line.

It's not an objective flaw/improvement, but I would say I'm a bit hesitant to hook in this sort of knock-on effect server-side; I'd prefer this to be kept in client-side logic. There's already third & first party clients/userscripts that manually implement this behavior clientside, & it'd be a pain for them & us to have to change that, plus it's just less clean in my personal opinion.

If that is changed, then I'm also kinda split on migrating this to be stored in LocalStorage instead of server-side, but if the blacklist is stored server-side & solely processed client-side, there's nothing really *wrong* with the same being true for this.

That proposed UI/UX is really nice though, & a lot more elegant. I'd say if that is what is decided on, I'd definitely lean away from the server-side implementation entirely & just make it all client-side. I imagine it'd also just be far simpler to have it check if that flag is set client-side & just call `document.querySelector(".ptbr-vote-button").click()` than to get lost in server-side code.

All in all, this is a really good job, and this would definitely be a welcome addition to the site, no matter the form it ends up taking.

pupgray:
Thanks so much @jmortiger! Yeah I can agree with that. My only real worry with just handling this in the FE is missing some other point in the UI/API where it should do the same thing given I set it up as a User model property buuuuuut if it was just an entirely local option that seems acceptable. Would be a setting for the button specifically, rather than the user.

(also, my day job is in Ruby HAH, force of habit I suppose)

I can whip up the popover UI for this and switch over to just handling it in the FE sometime this week probably

pupgray:
(Sorry! Will eventually got to that UI tweak soon)

Sindrake:
I apologize for not getting to this PR sooner.

A feature like this had been requested for a very long time now.
Unfortunately, it's been up in the air – not just because of the actual work of putting it together, but also because it's not quite clear whether this is something we want.

The purpose of the favorites system is to let users save the posts they like the most.
But that's not how a lot of people use it. Instead, they favorite just about everything that they vaguely like – while some people use it to mark posts that they've already seen. Personally, I don't think that we should be feeding into that, and instead should make voting more prominent.

That's not my final word on the matter, mind you – I am not quite convinced one way or another.

pupgray:
Heya @Sindrake sorry, I'm also rather late to getting back to this! haha

Faaaaair point. If you want, since the default in this branch is the old behaviour, and the like+fave behaviour has to be toggled on in settings, it serves the purpose of having a solution for that minority of users if they come asking for it without making it the default and flooding the site with more votes than normal. Though of course, they could just use re621 so ehhHHHhhH? ¯\\\_(ツ)_/¯

Option remains open to ya!

I was thinking a lot of people in that linked forum post really just haven't fiddled with Sets at all. Maybe there's some tweaks to be made to how Sets work/get added to that would free up favourites to work like... favourites haha

Sindrake:
This is intriguing, but unfortunately I am going to have to say no.

PR #1604, opened 2026-01-25
[DText] Quoting blips/comments/forum posts links to the original message

jmortiger:
When "Reply"/"Respond"ing to a blip/comment/forum post, it actually adds a link to the message being responded to instead of the old method of just quoting it with no way to find to original message beyond searching for the text.

It currently uses the same link as the UI does; e.g.
* forum posts link to the post's anchor on its forum topic's page
* comments link to the comment's anchor on its post's page
* blips link directly to the blip's page


Sindrake:
I am going to reject this, sorry.
I just don't really like how it looks - makes quotes look more busy.

PR #1619, opened 2026-01-31
[Posts] Don't mark one-frame APNGs as animated

### LINKED ISSUE #1467: Only tag apng as `animated` if they have multiple frames
From bug report:
> As for handling these cases, I noticed a bit ago that we actually don't correctly condition the animated flag on the number of frames. We say the file is corrupt if [the number of frames is less than 1](https://github.com/e621ng/e621ng/blob/master/app/logical/apng_inspector.rb#L81), but we mark it as animated if it's equal to one. We should just not mark it if it's "animated" but only has 1 frame.
> 
> I didn't see many cases where someone would intentionally make a 1 frame animation, so I didn't think it was worth bringing up before, but this certainly makes it more pressing.

svehaq:
Fixes #1467.

### DIFF
--- a/app/logical/apng_inspector.rb
+++ b/app/logical/apng_inspector.rb
@@ -81,7 +81,7 @@ class ApngInspector
         if framecount < 1
           actl_corrupted = true
         else
-          @animated = true
+          @animated = framecount > 1
           @frames = framecount
         end
       end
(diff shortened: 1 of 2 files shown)


svehaq:

Sindrake (approved):

PR #1625, opened 2026-02-09
[PostFlags] Add option to change who can view flag notes

jmortiger:
On request from sister site.

Allows choosing between:
* `:staff`: Only flagger & staff can see flag notes (default; current behavior)
* `:uploader`: Only flagger, staff, & post uploader can see flag notes
* `:users`: All logged-in users can see flag notes
* `:all`: Everyone, logged in or not, can see flag notes

Can be configured with `Danbooru::CustomConfiguration`

* Tested
   * HTML response fully manually tested for `Danbooru::CustomConfiguration`
   * For the JSON
      * A partial test & a deferred test for `Danbooru::CustomConfiguration` (need to figure out editing `Danbooru::CustomConfiguration` at runtime for testing to test it working for all values; tests for current value)


Sindrake:
This seems overly complicated.

Why is it necessary to be able to change who can view flag notes on the fly? Just having it as a variable in the config file seems fine to me.

Plus, I don't really like that this PR seemingly both adds options for flag note visibility **and** sets up the framework for overriding config variables with the Setting values. If we were to do that at all, it should probably be a separate PR.

jmortiger:
I didn't know what the different performance implications were of the 2 options, & figured it'd be trivial to revise this to work for either after. Also, I couldn't automate tests for all the values if solely supporting the `Danbooru::CustomConfiguration` solution.

If the framework seems worth adding separately, lmk & I'll strip it out to its own PR. Otherwise, I can pare this back down to just a `Danbooru::CustomConfiguration` field.

Sindrake:
I haven't tested performance.
But yeah, I think it's best to keep it simple, with just pulling the value from the config.

I don't know if we have a use case for this hybrid framework right now, to be completely honest with you.

jmortiger:
Stripped back down; should be good to go now.

Sindrake (approved):

PR #1644, opened 2026-03-09
[Stats] Add count of active users to site stats

mm12:
Adds how many users have their last-logged-in date in the past 3 months.


Sindrake:
Have you considered the performance impact here?
This query would be running a sequential scan on a table with 2.5 million users.

It's not the end of the world, since it's in a background job. But it may time out.
Please, consider adding an index on the `last_logged_in_at` column.

At the same time, consider whether we actually need this information – and whether we want to expose it to the userbase.

mm12:
> Have you considered the performance impact here? This query would be running a sequential scan on a table with 2.5 million users.
> 
> It's not the end of the world, since it's in a background job. But it may time out. Please, consider adding an index on the `last_logged_in_at` column.

Now I'm curious about how bad it is. Didn't get meaningful results with
```
sql = "EXPLAIN (ANALYZE, BUFFERS) SELECT COUNT(*) FROM users WHERE last_logged_in_at >= NOW() - INTERVAL '3 months';"
ActiveRecord::Base.connection.execute(sql).as_json
```
since its just a test instance... I suppose adding an index wouldn't be hard.

app/logical/stats_updater.rb
```diff
@@ -50,6 +50,7 @@ def self.run!
     stats[:unactivated_users] = User.where.not(email_verification_key: nil).count
     stats[:total_dmails] = (Dmail.maximum("id") || 0) / 2
     stats[:average_registrations_per_day] = daily_average.call(stats[:total_users])
+    stats[:active_users] = User.where("last_logged_in_at >= ?", 3.months.ago).count
```
Sindrake, line 53: This needs an index on `last_logged_in_at`.

Sindrake (changes requested):

Sindrake (dismissed):

Issue #10, opened 2019-02-06
Convert delayed_job to resque/sidekiq


kiranoot:
Investigate license issues surrounding sidekiq. If it can't be used, then use resque. Both do approximately the same thing, but are significantly better than delayed_job.

Need to isolate existing delayed_job entries and see what can be salvaged from them.

E621 makes heavy use of background tasks to accomplish slow moving tasks, and these queues often get very large. delayed_job is not likely to cut it for production use when the queues hit sizes in the hundreds of thousands.

PR #1646, opened 2026-03-10
[Embeds] Better newline handling in embeds

### LINKED ISSUE #1645: User profile embed ignores newlines
BUG: When embedding users profiles (such as on discord), newlines are *completely* ignored. This leads to 
 > something something
 > something
 
turning into 
> something somethingsomething

`app/views/static/_embed.html/erb`

mm12:
Works, in theory. I don't have a good way to test if the embed is fixed. 
fixes #1645

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -88,6 +88,11 @@ module ApplicationHelper
     end
   end
 
+  def format_plaintext(text)
+    return "" if text.nil?
+    strip_tags(format_text(text).gsub(/<[^>]+>/, " ")).squish
+  end
+
   def custom_form_for(object, *args, &)
     options = args.extract_options!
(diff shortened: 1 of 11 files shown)


Sindrake:
Issues pointed out by Copilot aside, this won't work.
First of all, rails will convert `&#10` to `&amp;#10` automatically.

But even if it didn't, the description that is being passed on to the embed partial is already missing newlines.
https://github.com/e621ng/e621ng/blob/master/app/views/forum_topics/show.html.erb#L43C107-L43C163
The HTML content generated by `format_text` does not have newlines. It has HTML tags, which then get stripped off.

Sindrake:
Not a perfect solution – newlines are just replaced with spaces.
But it'll work for now, and I ran out of time and patience.

PR #1684, opened 2026-03-22
[Blips][UI?] Rename hide and delete to delete and destroy for Blips, #1358

Miliainare:
An initial attempt to more slowly chip away at  #1358. Tested locally and works fine from what I can see, but I've never used Blips. Please let me know if I missed anything!

PR #1702, opened 2026-03-25
[Blips] Rename hide and delete to delete and destroy respectively

Miliainare:
Again, for #1358  . I probably did something wrong in the previous PR to get it into a closed state sooo... Let's go again?


Sindrake (dismissed):

Sindrake:
Wait how did this end up not including changes from #1684?
Thanks to the merges and force-pushes (that I explicitly asked you not to do), this really ended up being quite a mess.

Miliainare:
Is the preferred strategy for resolving merge conflicts from syncing with `e621ng:master` using merge commits as opposed to rebasing?

Sindrake:
> Is the preferred strategy for resolving merge conflicts from syncing with `e621ng:master` using merge commits as opposed to rebasing?

Sorry, I should have been more clear.
There are two different problems in play here.

1. Force-pushing into PRs after changes are requested makes it harder for me to check what actually changed, and whether the issues were addressed.
2. Aside from that, rebasing the branch is usually fine, but you would need to bump the migration ID to be at the end of the list. Otherwise, things become messy quick.

Push comes to shove, I'll just resolve the issue myself before I squash and merge everything into master.

PR #1706, opened 2026-03-26
[WikiPages] Tag edit changes

DonovanDMC:
Currently wiki pages have two issues with creating unneded data:
* When editing a tag's category via a wiki page, an empty wiki page will be created if one doesn't already exist
* When creating a wiki page, a tag will always be created even if no category is selected

This pr fixes both of those, and restores the prior behavior of preventing creating empty wiki pages. Updating is unchanged.


DonovanDMC:
I'll fix those issues tomorrow, I only just got the notification

.rubocop.yml
```diff
@@ -163,4 +163,11 @@ Style/TrailingCommaInArrayLiteral:
   EnforcedStyleForMultiline: consistent_comma
 
 Style/TrailingCommaInHashLiteral:
-  EnforcedStyleForMultiline: consistent_comma
\ No newline at end of file
+  EnforcedStyleForMultiline: consistent_comma
+
+Rails/DynamicFindBy:
```
Sindrake: Don't do this, it's silly.

First of all, we do want to phase out the methods that override the dynamic find_by methods eventually.
Second, this would suppress the warnings even on non-overriden methods, which we definitely do not want.

app/views/wiki_pages/show.html.erb
```diff
@@ -21,11 +21,11 @@
...
       <div id="wiki-page-body" class="dtext-container">
-        <% if wiki_content.body.present? %>
-          <%= format_text(wiki_content.body, allow_color: true, max_thumbs: 75) %>  
-        <% else %>
+        <% if wiki_content.empty? %>
           <p>This wiki page is empty. <%= link_to "Click here to add details", edit_wiki_page_path(@wiki_page) %>.</p>
-        <% end %>
+        <% else %>
+          <%= format_text(wiki_content.body, allow_color: true, max_thumbs: 75) %>
+      <% end %>
```
Sindrake: Nitpick: indentation is off on this line.

app/controllers/wiki_pages_controller.rb
```diff
@@ -67,8 +67,30 @@ def show
   end
 
   def create
-    @wiki_page = WikiPage.create(wiki_page_params(:create))
-    respond_with(@wiki_page)
+    @wiki_page = WikiPage.new(wiki_page_params(:create))
+    # avoid creating blank wiki pages when only tag attributes are modified
+    if @wiki_page.tag_will_change? && @wiki_page.no_content?
+      WikiPage.transaction do
+        tag_updated = catch(:abort) do
```
Sindrake, line 74: Catching `:abort` seems both fragile and a little odd.
Why not have `create_tag` / `update_tag` return a boolean instead?

Sindrake (changes requested):

.rubocop.yml
DonovanDMC: I only specifically whitelisted `find_by_normalized_name`, `find_by_name_or_id`, and `find_by_name_list`, all names we will almost certainly never have as a column

I've undone it

app/controllers/wiki_pages_controller.rb
DonovanDMC: returning false doesn't trigger a rollback because they're callbacks, not validations*
The only way this would be fragile is if we were to throw a regular error (i.e. not an `ActiveRecord` error) within the functions, but that's both bad practice and I'm pretty certain that would result in an elevated error regardless


\* the return value would also be ignored in validations, but it would check the `errors` object, that doesn't happen with callbacks

DonovanDMC:
Is there something blocking this from being accepted? Implementation too janky for comfort? Or just not a priority?

PR #1718, opened 2026-03-29
[Search] Add metatags relating to flag data

mm12:
Adds metatags relating to flag data. These aim to provide functionality equivalent to that on the `post_flags` index's commit search, but using the post search's more advanced capabilities. 

Filters:
  - `flaggedby:{USER}` (staff-only) – Allows searching by what user created the flag, similar to `deletedby`
  - `flagnote:{contents}` (staff-only) – Allows searching of the flag notes provided by users, with wildcard support
  - `flagreason:{contents}` – Allows searching of the flag reason, with wildcard support. This is a `keyword` rather than `text`.

Orders:
  - `order:flagged` – Allows searching by the most recently flagged post
  - `order:deleted` – Allows searching by the most recently deleted post

<details><summary>Notes to self</summary>

#### Testing fix scripts
1. switch to master
2. run `bundle exec rails runner 'Post.document_store.delete_index!; Post.document_store.create_index!(delete_existing: true); Post.document_store.import'` - deleted current post index entirely and rebuilds it.
3. run fixer scripts, ensure everything works
</details>


app/logical/tag_query.rb
```diff
@@ -1447,6 +1447,18 @@ def parse_query(query, depth: 0, **kwargs)
...
+      # For flag metatags, intentionally allow deleted posts to be included.
+      when "flaggedby", "-flaggedby", "~flaggedby"
+        next unless CurrentUser.is_staff?
+        add_to_query(type, :flagger, user_id_or_invalid(g2))
+
+      when "flagreason", "-flagreason", "~flagreason"
+        add_to_query(type, :flagreason, g2.downcase, wildcard: true)
+
+      when "flagnote", "-flagnote", "~flagnote"
+        next unless CurrentUser.is_staff?
```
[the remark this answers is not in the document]
mm12: Correct wording would be "doesn't exclude" rather than "include"

app/indexes/post_index.rb
```diff
@@ -150,6 +155,14 @@ def import(options = {})
              GROUP BY post_id) pfi
           INNER JOIN post_flags pf ON pf.id = pfi.mid;
         SQL
+        flag_sql = <<-SQL
+          SELECT pf.post_id, pf.creator_id, LOWER(pf.reason) as reason, LOWER(pf.note) as note FROM
+            (SELECT MAX(id) as mid, post_id
+             FROM post_flags
+             WHERE post_id IN (#{post_ids}) AND is_resolved = false AND is_deletion = false
+             GROUP BY post_id) pfi
```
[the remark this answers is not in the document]
mm12: ^^^ Need to investigate if this is related to #1651 at all

app/indexes/post_index.rb
```diff
@@ -164,6 +177,9 @@ def import(options = {})
         deletions        = conn.execute(deletion_sql)
         deleter_ids      = deletions.values.map { |p, did, dr| [p, did] }.to_h
         del_reasons      = deletions.values.map { |p, did, dr| [p, dr] }.to_h
+        flagger_ids      = conn.execute(flag_sql).values.map { |p, fid, fr, fn| [p, fid] }.to_h
+        flag_reasons     = conn.execute(flag_sql).values.map { |p, fid, fr, fn| [p, fr] }.to_h
+        flag_notes       = conn.execute(flag_sql).values.map { |p, fid, fr, fn| [p, fn] }.to_h
```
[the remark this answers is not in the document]
mm12: No preference on if it is done this way or the slightly different way `deletions` does it, but it should be consistent between them

app/indexes/post_index.rb
```diff
@@ -284,6 +306,9 @@ def as_indexed_json(options = {})
       has_children:             has_children,
       has_pending_replacements: options.key?(:has_pending_replacements) ? options[:has_pending_replacements] : replacements.pending.any?,
       artverified:              options.key?(:artverified) ? options[:artverified] : uploader_linked_artists.any?,
+      # use SQL to get most recent flag id for a given post, which will make it in order without needing the date
```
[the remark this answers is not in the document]
mm12: honestly, removing the comment outright is the best move

app/indexes/post_index.rb
```diff
@@ -284,6 +306,9 @@ def as_indexed_json(options = {})
...
+      flagged_at:               ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: false).order(id: :desc).limit(1).pick(:created_at),
+      deleted_at:               ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: true).order(id: :desc).limit(1).pick(:created_at),
```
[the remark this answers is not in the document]
mm12: This is overly complicated... An optimization can be made by using `DISTINCT ON with ORDER BY id DESC on (post_id, is_deletion)` in the SQL above, but it increases complexity a bit more than I would like. Instead, I will try to use the optimization from elsewhere to only use this query as a fallback and get this value from the existing SQL.

app/logical/tag_query.rb
```diff
@@ -1457,6 +1457,18 @@ def parse_query(query, depth: 0, **kwargs)
         q[:show_deleted] ||= true
         add_to_query(type, :deleter, user_id_or_invalid(g2))
 
+      # For flag metatags, intentionally allow deleted posts to be included.
+      when "flaggedby", "-flaggedby", "~flaggedby"
+        next unless CurrentUser.is_staff?
+        add_to_query(type, :flagger, user_id_or_invalid(g2))
```
Sindrake, line 1499: The comment says "intentionally allow deleted posts to be included" but `show_deleted` is never set — unlike `deletedby`/`delreason`, deleted posts are hidden by default. For `flaggedby` in particular, staff searching by flagger will typically be looking for flagged-then-deleted posts. Consider mirroring the `deletedby` pattern:

```suggestion
      when "flaggedby", "-flaggedby", "~flaggedby"
        next unless CurrentUser.is_staff?
        q[:status] ||= "any" unless q[:status_must_not]
        q[:show_deleted] ||= true
        add_to_query(type, :flagger, user_id_or_invalid(g2))
```

app/logical/elastic_post_query_builder.rb
```diff
@@ -135,6 +135,10 @@ def innate_hide_deleted_posts?(at_any_level: !GLOBAL_DELETED_FILTER)
     "comment_bumped" => [{ comment_bumped_at: { order: :desc, missing: :_last } }, { id: :desc }],
     "comment_bumped_asc" => [{ comment_bumped_at: { order: :asc, missing: :_last } }, { id: :desc }],
     # "random" => [{ _score: :desc }],
+    "flagged" => [{ flagged_at: { order: :desc, missing: :_last } }, { id: :desc }],
+    "flagged_asc" => [{ flagged_at: { order: :asc, missing: :_last } }, { id: :asc }],
+    "deleted" => [{ deleted_at: { order: :desc, missing: :_last } }, { id: :desc }],
```
Sindrake, line 141: `order:deleted` sorts by `deleted_at`, but deleted posts are hidden by default — the sort is a no-op unless the user also adds `status:deleted` or `status:any`. Since this order is only meaningful for deleted posts, consider whether it should auto-show deleted posts the way `deletedby` does (i.e. add to `OVERRIDE_DELETED_FILTER_METATAGS` and set `show_deleted` in `tag_query.rb`), or at minimum document that `status:deleted` is required for it to be useful.

app/indexes/post_index.rb
```diff
@@ -260,8 +286,11 @@ def as_indexed_json(options = {})
...
-      deleter:                  options[:deleter]    || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: true).order(id: :desc).first&.creator_id,
-      del_reason:               options[:del_reason] || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: true).order(id: :desc).first&.reason&.downcase,
+      deleter:                  options[:deleter]       || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: true).order(id: :desc).first&.creator_id,
+      del_reason:               options[:del_reason]    || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: true).order(id: :desc).first&.reason&.downcase,
+      flagger:                  options[:flagger]       || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: false).order(id: :desc).first&.creator_id,
+      flag_reason:              options[:flag_reason]   || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: false).order(id: :desc).first&.reason&.downcase,
+      flag_note:                options[:flag_note]     || ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: false).order(id: :desc).first&.note&.downcase,
```
Sindrake: Three separate DB queries hit the same table with identical `WHERE`/`ORDER` clauses. These could share a single query, similar to how the `import` bulk path works:

```ruby
_flag = unless options.key?(:flagger) && options.key?(:flag_reason) && options.key?(:flag_note)
          ::PostFlag.where(post_id: id, is_resolved: false, is_deletion: false).order(id: :desc).first
        end
flagger:     options[:flagger]     || _flag&.creator_id,
flag_reason: options[:flag_reason] || _flag&.reason&.downcase,
flag_note:   options[:flag_note]   || _flag&.note&.downcase,
```

(The same issue exists for `deleter`/`del_reason` above, but this PR makes it worse by adding a third query.)

Sindrake (dismissed):
This PR is missing the `db/fixes/` scripts needed to update the live OpenSearch index mapping. Without a `put_mapping` call for the 5 new fields (`flagged_at`, `deleted_at`, `flagger`, `flag_reason`, `flag_note`), these fields don't exist in any running instance's index — which is why the new metatags silently don't work on existing deployments. A backfill script is also needed to populate the new fields for existing posts (see `db/fixes/114_1_add_opensearch_artverified_index.rb` and `114_2_add_opensearch_artverified_data.rb` for the pattern).

app/logical/elastic_post_query_builder.rb
mm12: Will do for deleted. Want to document that while flags might be similar, showing deleted posts there by default would make things difficult.

mm12:
ignore some of those commits, messed up on a rebase from master

app/logical/tag_query.rb
```diff
@@ -1452,11 +1471,24 @@ def parse_query(query, depth: 0, **kwargs)
...
+        next unless CurrentUser.is_staff?
+        q[:status] ||= "any" unless q[:status_must_not]
+        q[:show_deleted] ||= true
+        add_to_query(type, :flagger, user_id_or_invalid(g2))
+
+      when "flagreason", "-flagreason", "~flagreason"
+        add_to_query(type, :flagreason, g2.downcase, wildcard: true)
+
+      when "flagnote", "-flagnote", "~flagnote"
+        next unless CurrentUser.is_staff?
```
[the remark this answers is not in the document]
mm12: Intentional - flags should not show deleted posts. tbh i would rather that `flaggedby` not either, but that actually have a usecase.

test/test_helper.rb
```diff
@@ -29,6 +29,23 @@ def format_rerun_snippet(result)
...
+        if result.respond_to?(:source_location)
+          result.source_location
+        else
+          result.method(result.name).source_location
+        end
+
+      "#{self.class.executable} #{relative_path_for(location)}:#{line}"
+    end
+  end
+end
```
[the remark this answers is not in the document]
mm12: ugh did that for debugging. realistically that change shouldn't be in this PR but another one. I think I'll remove the guarded one and we can worry about this later.

app/logical/tag_query.rb
```diff
@@ -407,7 +416,17 @@ def self.should_hide_deleted_posts?(query, always_show_deleted: false, at_any_le
...
-    !TagQuery.has_metatags?(query, *OVERRIDE_DELETED_FILTER_METATAGS, prepend_prefix: false, at_any_level: at_any_level, has_all: false)
+    !TagQuery.deleted_filter_order_override?(query, at_any_level: at_any_level) &&
+      !TagQuery.has_metatags?(query, *OVERRIDE_DELETED_FILTER_METATAGS, prepend_prefix: false, at_any_level: at_any_level, has_all: false)
+  end
+
+  def self.deleted_filter_order_override?(query, at_any_level: true)
+    if query.is_a?(TagQuery)
+      query[:order].in?(OVERRIDE_DELETED_FILTER_ORDERS)
+    else
+      order_value = TagQuery.fetch_metatag(query, "order", "-order", prepend_prefix: false, at_any_level: at_any_level)
```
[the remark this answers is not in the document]
Sindrake: Alternative:
```Ruby
all_orders = TagQuery.fetch_metatags(query, "order", "-order", prepend_prefix: false, at_any_level: at_any_level)
order_value = (all_orders["order"] || all_orders["-order"] || []).last
order_value.present? && order_value.in?(OVERRIDE_DELETED_FILTER_ORDERS)
```

app/logical/tag_query.rb
```diff
@@ -366,7 +373,7 @@ def is_grouped_search?
   # * `RuntimeError`: when `q[:children_show_deleted]` is `nil` & any element in `q[:groups]` is a
   # `TagQuery`, as `q[:children_show_deleted]` shouldn't be `nil` if subsearches were processed.
   def hide_deleted_posts?(always_show_deleted: false, at_any_level: false)
-    if always_show_deleted || q[:show_deleted]
+    if always_show_deleted || q[:show_deleted] || q[:order].in?(OVERRIDE_DELETED_FILTER_ORDERS)
```
jmortiger: You should simply update `q[:show_deleted]` when you update `q[:order]` instead of tracking the state independently; this line shouldn't change.

app/logical/tag_query.rb
```diff
@@ -407,7 +416,18 @@ def self.should_hide_deleted_posts?(query, always_show_deleted: false, at_any_le
...
+    !TagQuery.deleted_filter_order_override?(query, at_any_level: at_any_level) &&
+      !TagQuery.has_metatags?(query, *OVERRIDE_DELETED_FILTER_METATAGS, prepend_prefix: false, at_any_level: at_any_level, has_all: false)
+  end
+
+  def self.deleted_filter_order_override?(query, at_any_level: true)
+    if query.is_a?(TagQuery)
+      query[:order].in?(OVERRIDE_DELETED_FILTER_ORDERS)
+    else
+      all_orders = TagQuery.fetch_metatags(query, "order", "-order", prepend_prefix: false, at_any_level: at_any_level)
+      order_value = (all_orders["order"] || all_orders["-order"] || []).last
```
jmortiger: The ethos of removing the deleted filter is that if *anything* implies it should be removed, it's removed, regardless of whether or not it's actually included in the output. That's why `q[:show_deleted]` is always assigned with a `||= true` & never cleared with a `= false`/`= nil`; even if they formatted their search wrong & their `order` is overridden by a later `order`, they expect deleted posts to show up, so they should show up. If that expectation isn't preserved in this case, it needs to be removed for *all* cases; if not, it should be preserved here.

This also enables a performance improvement; returning from the block immediately upon finding a value for `order`/`-order` that would cause deleted posts to appear would cancel the rest of `fetch_metatags`'s execution.

...plus it's way easier to just not worry about where the thing shows up.

```suggestion
      TagQuery.fetch_metatags(query, "order", "-order", prepend_prefix: false, at_any_level: at_any_level) do |_tag, value|
        return true if value.in?(OVERRIDE_DELETED_FILTER_ORDERS)
      end
```

If the current behavior is preferred, the current version would still fail if there were both negated & non-negated `order`s & the final value was a negated one.

app/logical/tag_query.rb
jmortiger: If you want to change if the automatic deleted filter is enabled or disabled with `order` values, then change this to
```suggestion
when "order", "-order" then
  q[:order] = TagQuery.normalize_order_value(g2.downcase, invert: type == :must_not)
  # q[:show_deleted] ||= q[:order].in?(OVERRIDE_DELETED_FILTER_ORDERS)
  # or, since this value is normalized
  q[:show_deleted] ||= q[:order] == "deleted"
```
to preserve the expectations of the other mechanisms for implicitly overriding the filter.

app/logical/tag_query.rb
```diff
@@ -1452,11 +1472,24 @@ def parse_query(query, depth: 0, **kwargs)
...
-      when "deletedby", "-deletedby", "~deletedby"
+      when "deletedby", "-deletedby", "~deletedby", "deleter", "-deleter", "~deleter"
         q[:status] ||= "any" unless q[:status_must_not]
         q[:show_deleted] ||= true
         add_to_query(type, :deleter, user_id_or_invalid(g2))
 
+      when "flaggedby", "-flaggedby", "~flaggedby", "flagger", "-flagger", "~flagger"
+        next unless CurrentUser.is_staff?
+        q[:status] ||= "any" unless q[:status_must_not]
+        q[:show_deleted] ||= true
```
jmortiger: Considering the `flagreason` & `flagnote` metatag does not implicitly remove the automatic deleted post filter, I'd argue this shouldn't either.

jmortiger (changes requested):
Sorry it took so long, I didn't get an email about this.

app/logical/tag_query.rb
mm12: I would agree, but https://github.com/e621ng/e621ng/pull/1718#discussion_r3120388250 makes a good point

app/logical/tag_query.rb
mm12: The previous code used here (https://github.com/e621ng/e621ng/pull/1718#discussion_r3130664571) was 
```rb
def self.deleted_filter_order_override?(query, at_any_level: true)
  if query.is_a?(TagQuery)
    query[:order].in?(OVERRIDE_DELETED_FILTER_ORDERS)
  else
    order_value = TagQuery.fetch_metatag(query, "order", "-order", prepend_prefix: false, at_any_level: at_any_level)
    order_value.present? && order_value.in?(OVERRIDE_DELETED_FILTER_ORDERS)
  end
end
```

However, this fails tests where the `deleted` order comes last
```rb
assert_not(TagQuery.should_hide_deleted_posts?("aaa bbb order:random order:deleted"))
assert_not(TagQuery.should_hide_deleted_posts?("aaa bbb order:random -order:deleted"))
```
Will need to look into a workaround here instead of just going back

jmortiger: Does the provided suggestion address this? It should only return true once a relevant `order` value has been found.

app/logical/tag_query.rb
jmortiger: If you mean
> staff searching by flagger will typically be looking for flagged-then-deleted posts

I'd argue that's overly-presumptive; I can see staff looking for someone's unresolved flags just as well. Besides, for staff, usually the `flagger` & the `deleter` are the same person, so I'd rather them specify `~flagger:user ~deleter:user` when that's what they want; otherwise it feels rife for confusion with minimal subjective benefit. It is a personal preference thing though, neither is objectively wrong or anything; I'll defer to your judgement.

jmortiger (dismissed):
Aside from personal preference on `flaggedby` (which we're already discussing), it all looks good to me.

app/logical/tag_query.rb
mm12: Funny how i wrote that exact code before i saw this

app/logical/tag_query.rb
mm12: `deleter` and `flagger` are almost _never_ the same person. Deletions are credited to the staff member who deleted the post.

Sindrake (dismissed):

PR #1750, opened 2026-04-01
[Pools] Pool post validation

### LINKED ISSUE #1722: Pool creation crashes with out-of-range post IDs (no validation on user-supplied values)
### Problem
When creating a new pool, supplying an excessively large post ID value (e.g., 62062876206289) causes a crash:

```
ActiveModel::RangeError
62062876206289 is out of range for ActiveModel::Type::Integer with limit 4 bytes

app/controllers/pools_controller.rb:43:in `create'
lib/middleware/parameter_sanitizer.rb:27:in `call'
```

#### Insights
- The root cause is that post IDs supplied by the user are somewhat normalized, but not validated for type or range.
- This allows values outside the signed 32-bit integer range (such as 62062876206289) to make it to ActiveRecord, resulting in a RangeError and a 500 error.

### Steps to Reproduce
1. Go to the new pool creation page.
2. Supply an out-of-range post ID in the Post IDs field (e.g., 62062876206289).
3. Submit the form.
4. Observe the application error.

### Expected Behavior
- Pool creation should fail gracefully with a validation error if any supplied post ID is not present in the database or is out of bounds for the post ID type.
- The user should receive an actionable error message indicating which post IDs were invalid.

### Actual Behavior
- The server crashes with an unhandled `ActiveModel::RangeError` when an out-of-bounds post ID is included.

### Suggestion
- Add explicit validation to the `post_ids` attribute on `Pool` model to ensure that:
    - All post IDs are within the integer range supported by Postgres and ActiveRecord.
    - Each post ID is a positive integer and refers to an existing post.
- Prevent invalid values from causing database-level errors by surfacing appropriate model validation errors instead.

----

References:
- [pools_controller.rb#L43](https://github.com/e621ng/e621ng/blob/master/app/controllers/pools_controller.rb#L43)
- [pool.rb](https://github.com/e621ng/e621ng/blob/master/app/models/pool.rb)

Tarrgon:
Fixes #1722

### DIFF
--- a/app/models/pool.rb
+++ b/app/models/pool.rb
@@ -224,7 +224,7 @@ class Pool < ApplicationRecord
     invalid_ids = []
 
     added.each do |id|
-      safe_id = ParseValue.safe_id(safe_id)
+      safe_id = ParseValue.safe_id(id)
       if safe_id <= 0 || !Post.exists?(id)
         invalid_ids.push(safe_id)
       end
--- a/app/models/pool.rb
+++ b/app/models/pool.rb
@@ -222,12 +222,21 @@ class Pool < ApplicationRecord
     return if added.empty?
 
     invalid_ids = []
+    sanitized_ids = []
 
     added.each do |id|
       safe_id = ParseValue.safe_id(id)
       if safe_id <= 0 || !Post.exists?(id)
-        invalid_ids.push(safe_id)
+        invalid_ids.push(id)
+      elsif safe_id > 0
+        sanitized_ids.push(safe_id)
       end
+    end
+
+    if sanitized_ids.any?
+      existing_ids = Post.where(id: sanitized_ids).pluck(:id)
+      missing_ids = sanitized_ids - existing_ids
+      invalid_ids.concat(missing_ids)
     end
 
     return if invalid_ids.empty?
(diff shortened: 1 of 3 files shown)


Tarrgon:
I forgot this may require updating other tests now that some posts will no longer work

app/models/pool.rb
```diff
@@ -214,6 +215,36 @@ def validate_number_of_posts
...
+    added = post_ids - post_ids_before
+
+    return if added.empty?
+
+    invalid_ids = []
+    sanitized_ids = []
+
+    added.each do |id|
+      safe_id = ParseValue.safe_id(id)
+      if safe_id <= 0 || !Post.exists?(id)
```
Sindrake: This kinda works, but it's a bad idea.
You are running an extra DB query for each added post.
User adds 200 valid posts – you get 200 extra queries.

app/models/pool.rb
```diff
@@ -214,6 +215,36 @@ def validate_number_of_posts
...
+        invalid_ids.push(id)
+      elsif safe_id > 0
+        sanitized_ids.push(safe_id)
+      end
+    end
+
+    if sanitized_ids.any?
+      existing_ids = Post.where(id: sanitized_ids).pluck(:id)
+      missing_ids = sanitized_ids - existing_ids
```
Sindrake, line 238: This is pointless: `missing_ids` will always be empty.
Line 229 guarantees that posts with provided IDs exist.

Sindrake (changes requested):
This works, but creates a massive N+1 query issue.

You should first filter out all blatantly invalid IDs.
Then, bulk look up the rest of the IDs. The difference between the list of IDs you provide to that query and the list it returns will be invalid.

Sindrake (approved):

PR #1784, opened 2026-04-08
[Posts] Delete with reason DMails and deletion DMails titles

DRLaDRLa:
- Replace the basic `window.confirm()` dialog with something more useful
  - Send deletion DMail checkbox (always uses the default template)
  - Prefilled reason text box so it can be modified (doesn't show when reason is `from-flag`)
  - Enter and escape key shortcuts still work as before
- Implement deletion DMail titles
  - New field in the deletion confirmation page so it can be edited

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1621,6 +1621,19 @@ class Post < ApplicationRecord
     def pending_flag
       flags.unresolved.order(id: :desc).first
     end
+
+    def substitute_deletion_dmail_template(text, reason = nil)
+      unless text.presence
+        return nil
+      end
+      if reason
+        text = text.gsub("%REASON%", reason)
+      end
+      text.gsub("%POST_ID%", id.to_s)
+          .gsub("%STAFF_NAME%", CurrentUser.name)
+          .gsub("%STAFF_ID%", CurrentUser.id.to_s)
+          .gsub("%UPLOADER_ID%", uploader_id.to_s)
+    end
   end
 
   module VersionMethods
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1623,9 +1623,7 @@ class Post < ApplicationRecord
     end
 
     def substitute_deletion_dmail_template(text, reason = nil)
-      unless text.presence
-        return nil
-      end
+      return nil if text.blank?
       if reason
         text = text.gsub("%REASON%", reason)
       end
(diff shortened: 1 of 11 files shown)


DRLaDRLa:

app/views/moderator/post/posts/confirm_delete.html.erb
```diff
@@ -41,22 +41,20 @@
...
+              data-dmail-title="<%= @post.substitute_deletion_dmail_template(message[:title]) %>">
               <%= key.to_s.titleize %>
             </option>
           <% end %>
         <% end %>
       </select>
     </label>
 
+    <input type="text" size="50" name="dmail_title" id="dmail-title">
     <%# TODO: Figure out using <%= f.input :dmail, as: :dtext, limit: Danbooru.config.dmail_max_size %>
```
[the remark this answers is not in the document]
DRLaDRLa: Unnecessary.

app/models/post.rb
```diff
@@ -1621,6 +1621,19 @@ def deletion_flag
     def pending_flag
       flags.unresolved.order(id: :desc).first
     end
+
+    def substitute_deletion_dmail_template(text, reason = nil)
+      unless text.presence
```
Sindrake: Nitpick: This could just be `return nil if text.blank?`

app/views/moderator/post/posts/_delete_with_reason_dialog.html.erb
```diff
@@ -0,0 +1,20 @@
+<form class="simple_form" id="delete-with-reason-dialog" data-title="Delete Post" data-width="500" data-position="center">
+  <div class="input delete-with-reason-dialog-inputs">
+    <input type="text" id="delete-with-reason-dialog-input" placeholder="Enter reason here...">
+    <label for="delete-with-reason-dialog-enable-dmail">
+      Send DMail (default template)? <input type="checkbox" id="delete-with-reason-dialog-enable-dmail"
+        <%= " checked" if Danbooru.config.enable_post_deletion_dmail %>
+        <% message = Danbooru.config.post_deletion_dmail_templates.values.first %>
```
Sindrake: I think I would like to see some validation here, to make sure that there is at least one `post_deletion_dmail_templates` value configured. Both here and on the detailed deletion page.

Maybe don't even render the checkbox if no templates are configured.

app/views/moderator/post/posts/_delete_with_reason_dialog.html.erb
```diff
@@ -0,0 +1,20 @@
+<form class="simple_form" id="delete-with-reason-dialog" data-title="Delete Post" data-width="500" data-position="center">
+  <div class="input delete-with-reason-dialog-inputs">
+    <input type="text" id="delete-with-reason-dialog-input" placeholder="Enter reason here...">
```
Sindrake, line 4: Could you add a `<label>` for this element, and make the input stretch the full width?

Image

Sindrake (changes requested):
Pretty good, solid improvement overall.
Needs some minor tweaks.

DRLaDRLa:
CR changes made.

Sindrake (approved):

PR #1881, opened 2026-04-18
[UserFeedbacks] Move UserFeedback previews into a reusable ViewComponent

mm12:
Moves a reusable component for UserFeedback previews into it's own ViewComponent. 

#### "badge" style
**Staff view**
image
image

**Non-staff view**
image

#### "inline" style
**all users** (when shown by the calling view)
image


app/components/user_feedback_component.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      data-total="<%= positive - negative %>"
+    >
+      <%= svg_badge(deleted, "deleted") if deleted > 0 %>
+      <%= svg_badge(negative, "negative") if negative > 0 %>
+      <%= svg_badge(neutral, "neutral") if neutral > 0 %>
+      <%= svg_badge(positive, "positive") if positive > 0 %>
+    </a>
+  <% end %>
+
+  <% if CurrentUser.is_moderator? && CurrentUser.user != user && active == 0 %>
```
Sindrake: This is a nitpick, but for consistency's sake this should be `CurrentUser.user.is_moderator?` like it is in the `deleted` method.

Sindrake (changes requested):
Overall, looks good.

One thing  –  the `_feedback_badge.erb` file has not been removed.
Please, verify that it's not being used anymore, and remove it if that's the case.

app/components/user_feedback_component.html.erb
mm12: Note - looks like that was what feedback_badge was using before.

app/components/user_feedback_component.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      data-neutral="<%= neutral %>"
+      data-positive="<%= positive %>"
+      data-total="<%= positive - negative %>"
+    >
+      <%= svg_badge(deleted, "deleted") if deleted > 0 %>
+      <%= svg_badge(negative, "negative") if negative > 0 %>
+      <%= svg_badge(neutral, "neutral") if neutral > 0 %>
+      <%= svg_badge(positive, "positive") if positive > 0 %>
+    </a>
+  <% end %>
```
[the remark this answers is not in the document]
mm12: Will add, but since we are migrating out of our current test structure, I'll try to keep that in consideration.

... `context` -> `describe`, `should` -> `it`, `assert_equal` -> `assert`. Should be easy porting.

Sindrake (approved):

mm12:
gah, i forgot to copy in the css move from the local branch for 1289. Oh well, doesn't effect anything here.

Sindrake:
Followed up in #1890.

PR #1884, opened 2026-04-20
[ModActions] Fix errors caused by values unexpectedly containing nil values

mm12:
When a `mod_action`'s action is not in `known_actions`, the decorator logic does one of two things:
1. If the user is an admin, `"Unknown action #{object.action}: #{object.values.inspect}"`
2. Otherwise, `"Unknown action #{object.action}"`

However, `object.values` (from `ModAction.values` also does one of two things: 
1. If the user is an admin, it passed back the original values
2. Otherwise, it parses it to process what information needs to be sanitized to show the user (for example, removing ticket ids for non-moderators). 

If an unknown action exists and a non-admin attempts to view it, the `values` method attempts to process it's keys, resulting an an exception when it tries to `slice` (or `to_sym`) the nil value. 

To fix this, an early return for all of `values` is added when the original value is nil. Additionally, add safely operators to decorator cases that assume a kv pair exist.

----
Note: When we get tests for these, this should be a test.

### DIFF
--- a/app/decorators/mod_action_decorator.rb
+++ b/app/decorators/mod_action_decorator.rb
@@ -135,7 +135,7 @@ class ModActionDecorator < ApplicationDecorator
         "Changed #{user} level to #{vals['level']}"
       end
     when "user_flags_change"
-      "Changed #{user} flags. Added: [#{vals['added'].join(', ')}] Removed: [#{vals['removed'].join(', ')}]"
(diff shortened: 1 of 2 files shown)


mm12:
Well, I have realized my code doesn't do what I thought it did. It does solve the problem, just for a different reason. 

But now, I can't repro the issue undoing the changes. Even clearing rails and browser cache.

PR #1891, opened 2026-04-21
[ModActions] Normalize legacy mod actions

Sindrake:
Converts a bunch of pre-2021 mod action records to a modern format.
Required some tweaks to the mod action decorator to account for differences in value format.

The following actions were not converted:
* `locked_forum_post`, `unlocked_forum_post`, `stickied_forum_post`, `unstickied_forum_post`
  These store `forum_post_id` rather than `forum_topic_id`.
* `created_report_reason`, `edited_report_reason`
  These store `reason_id` rather than a strong `reason`

Also fixed some Rubocop alerts.

### DIFF
--- a/app/decorators/mod_action_decorator.rb
+++ b/app/decorators/mod_action_decorator.rb
@@ -22,7 +22,11 @@ class ModActionDecorator < ApplicationDecorator
       ### Pool ###
 
     when "pool_delete"
-      "Deleted pool ##{vals['pool_id']} (named #{vals['pool_name']}) by #{user}"
+      if vals["pool_name"]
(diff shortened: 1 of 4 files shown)


Sindrake:
I am losing the will to live with this thing.

app/decorators/mod_action_decorator.rb
```diff
@@ -338,7 +364,7 @@ def format_description
     when "report_reason_delete"
       "Deleted post report reason #{vals['reason']} by #{user}"
```
DRLaDRLa, line 365:
```suggestion
      "Deleted post report reason #{vals['reason']}"
```
Pretty sure this is redundant.

PR #1897, opened 2026-04-22
[Users] Allow admins to rename users with blank names

Sindrake:
The validation has been tightened to prevent users from making accounts with blank names.
But something needs to be done about existing accounts.


Sindrake:
This is wrong, since it makes admin name changes bypass logging in UserNameChangeRequests.

PR #1898, opened 2026-04-22
[UserPasswordResets] FIx an error caused by a blank password

### LINKED ISSUE #1853: NoMethodError in reset_user! when password param name mismatches form field name
## Summary

The password reset form submits the new password as `text`, but the controller reads `params[:password]`. This mismatch passes `nil` to `reset_user!`, where `ActiveSupport::SecurityUtils.secure_compare` calls `bytesize` on `nil` and raises a `NoMethodError`.

## Stack Trace

```
NoMethodError: undefined method `bytesize' for nil

app/models/user_password_reset_nonce.rb:19:in `reset_user!'
app/controllers/maintenance/user/password_resets_controller.rb:29:in `update'
```

## Root Cause

The form submits `params[:text]` for the new password, but `PasswordResetsController#update` ([app/controllers/maintenance/user/password_resets_controller.rb:29](app/controllers/maintenance/user/password_resets_controller.rb#L29)) calls `reset_user!(params[:password], params[:password_confirm])`. `params[:password]` is `nil`, and `ActiveSupport::SecurityUtils.secure_compare` does not handle `nil` arguments — it calls `bytesize` immediately, crashing before any comparison is made.

## Expected Behavior

Either the controller should read the correct param key (`params[:text]`), or the form field should be named `password`. Additionally, `reset_user!` could guard against `nil` arguments to avoid a crash in place of a clean "passwords do not match" redirect.

Sindrake:
Fixes #1853

### DIFF
--- a/app/models/user_password_reset_nonce.rb
+++ b/app/models/user_password_reset_nonce.rb
@@ -16,6 +16,8 @@ class UserPasswordResetNonce < ApplicationRecord
   end
 
   def reset_user!(pass, confirm)
+    return false if pass.blank? || confirm.blank?
+    return false if pass.length < 8
     return false unless ActiveSupport::SecurityUtils.secure_compare(pass, confirm)
     user.upgrade_password(pass)
     true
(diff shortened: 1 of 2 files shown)


Sindrake:
This is bad and stupid.
Start from scratch.

PR #1900, opened 2026-04-22
[Pagination] Prevent navigation buttons from linking to invalid sequential pages

### LINKED ISSUE #1899: PaginationError on page `a`
**Describe the bug**
Page `a` produces `Danbooru::Paginator::PaginationError exception raised`

**To Reproduce**
1. Go to `/mod_actions?page=a0` (doesn't need to be `mod_actions`)
2. Scroll down and Click on 'Next [page]' (`?page=b1`)
3. Click "Prev [page]" (`?page=a`)

**Expected behavior**
- ~~Handle exceptions of invalid pages~~ May not be practical here. Users should know they requested a bad page.
- paginator should not direct users to invalid pages

**Screenshots**
Internal

**Additional context**
Internal

mm12:
Fixes e621ng/e621ng#1899

Implementation results in `a` in some edge cases. Fallback to `0` to avoid invalid page numbers.

potential issues:
- Still does not prevent `PaginationError` from being delivered to users
  - Users should know if a page is invalid. 
- No tests check this fallback


mm12:
page `b1` has neither a forward nor a back button now, caused by `@records.last.id <= 1` / `last.id <= 1`

Sindrake:
I hate everything Copilot just said.

Sindrake (approved):
This does work, but it does not fix the underlying issue, in my opinion.

PR #1912, opened 2026-04-23
[Users] Fix an error caused by an invalid bcrypt hash

Sindrake:
Still not quite sure how they ended up like this.
https://us5.datadoghq.com/error-tracking/issue/34cd2eb6-d534-11ef-a522-da7ad0900000


Sindrake:
Bad idea, did not think it through.
I'll come back to it later.

PR #1920, opened 2026-04-25
[PostFlags] Move flag reasons into the DB

DRLaDRLa:
Finish PR https://github.com/e621ng/e621ng/pull/1380 by Cinder. Sort of, this doesn't attempt to merge post flags and reports into one.

- Grandfathering was moved to flag reasons. It works by setting a date and tag filter on a flag reason, with the option to negate the check. It achieves the same result of hiding a flag reason on posts matching certain criteria, but it's generalized.
  - I also considered just using a regular search query for filtering, but that means running a post search and checking if the post is returned. Something like `Post.tag_match("id:#{id} #{filter_query}").limit(1).count_only == 1` where for example `filter_query = "~date:=<2015-01-01 ~grandfathered_content"`. Simpler to implement and much more flexible, but performance is a concern. Maybe overkill.
-  Flags would do a `/uploading_guidelines/` regex match on the reason text to disable "delete with given reason", essentially forcing staff to give their own reason. I replaced this by a `needs_staff_reason` column. On existing flags it's filled by running that exact regex check using Postgres syntax.
- Flags would do a text search for "inferior" (actually it was done three different ways) to enable "move flag to parent post", toggle "move favorites" automatically, among other things. This was mostly replaced by `needs_parent_id`, except for moving the flag. That requires knowing the original reason name, because now you can have multiple flag reasons requiring that field. To solve that I added a `reason_name` column to post flags, again filling it by using the same matching logic in Postgres. It only recovers the name for the ` uploading_guidelines` and `inferior` flags since those are the only ones with specific matching logic, and they're also the only ones that need that value. It's null on other flags.
  - The hard coded flag message was also changed to `"#{flag_reason.reason} (post ##{parent_post.id})"` (originally `"Inferior version/duplicate of post ##{parent_post.id}"`).
  - I changed the inferior input field to behave like the note field, since it not longer applies exclusively to a single flag reason.
- I didn't have a good solution for the AI check automatic flag, since it hard codes a reason name (`uploading_guidelines`). I at least moved it to the config, but it still needs to match a reason name in the DB. It should be moved to the DB but I'm not sure where. Honestly I don't think it's worth worrying about.
- I made the necessary changes to get the tests passing again, but I'm not sure I've done it the best way possible. I'm not familiar with these kinds of testing frameworks.


app/controllers/post_flag_reasons_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def create
+    PostFlagReason.transaction do
+      @reason = PostFlagReason.create(reason_params)
+      ModAction.log(:flag_reason_create, { reason: @reason.reason, text: @reason.text })
+    end
+    flash[:notice] = @reason.valid? ? "Post flag reason created" : @reason.errors.full_messages.join("; ")
+    redirect_to post_flag_reasons_path
+  end
```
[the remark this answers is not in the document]
DRLaDRLa: TODO in original PR said it should be done like in `PostReportReason`, and that's how it's implemented there. No idea if the automated review comment is correct or not.

DRLaDRLa:
CR changes made.

Sindrake:
Looks like the tests update made a mess here. Sorry.

DRLaDRLa:
> Looks like the tests update made a mess here. Sorry.

Yeah I expected that. But at least getting the original tests passing helped me find a few bugs.

I'll update maybe tonight or later this week.

DRLaDRLa:
> Looks like the tests update made a mess here. Sorry.

I rebased and updated the tests once more. I'm not 100% I did it the best way, but they're passing.

app/javascript/src/js/pages/post_flags/post_flags.js
```diff
@@ -25,8 +25,14 @@ PostFlags.initFlagForm = function () {
     .addClass("required-indicator")
     .text(" (required)")
     .appendTo(noteLabel);
-  form.on("change", "input[type='radio']", updateNoteRequired);
+  form.on("change", "input[type='radio']", onReasonRadioChange);
```
Sindrake:
```suggestion
  form.on("change", "input[type='radio']", updateNoteRequired);
```

The function below can just be removed, its only purpose is to add a debug message.

app/views/posts/partials/show/content/notices/_notices.html.erb
```diff
@@ -11,11 +11,11 @@
...
         <% end %>
-        <% unless post.pending_flag&.reason =~ /uploading_guidelines/ %>
-           | <%= tag.a "Delete with given reason", href: "#", class: "delete-with-reason-link button btn-neutral", data: { "post-id": post.id, "reason": "", "from-flag": true, "prompt": "given reason", "move-favs": (post.pending_flag&.reason =~ /Inferior/).present? } %>
+        <% unless post.pending_flag&.needs_staff_reason? %>
+           | <%= tag.a "Delete with given reason", href: "#", class: "delete-with-reason-link button btn-neutral", data: { "post-id": post.id, "reason": "", "from-flag": true, "prompt": "given reason", "move-favs": post.pending_flag&.needs_parent_id?.present? } %>
         <% end %>
-        <% if !post.parent_id.nil? && post.pending_flag&.reason =~ %r{Inferior version/duplicate of post #\d+} %>
-           | <%= tag.a "Move flag to parent", href: "#", class: "move-flag-to-parent-link button btn-neutral", data: { "pid": post.id, "parent-id": post.parent_id } %>
+        <% if !post.parent_id.nil? && PostFlagReason.needs_parent_id?(post.pending_flag.reason_name) %>
```
Sindrake:
```suggestion
        <% if !post.parent_id.nil? && PostFlagReason.needs_parent_id?(post.pending_flag&.reason_name) %>
```

app/models/post_flag_reason.rb
```diff
@@ -0,0 +1,63 @@
...
+  # Cached list of reasons for name lookup
+  def self.by_name(reason_name)
+    name_lookup = Rails.cache.fetch("post_flag_reasons:for_by_name") { ordered.index_by(&:name) }
+    name_lookup[reason_name.to_s]
+  end
+
+  # Cached check for whether a reason requires an explanation
+  def self.needs_explanation?(reason_name)
+    explanation_map = Rails.cache.fetch("post_flag_reasons:for_needs_explanation") { ordered.pluck(:name, :needs_explanation).to_h }
+    !!explanation_map[reason_name.to_s]
```
Sindrake:
```suggestion
    by_name(reason_name)&.needs_explanation? || false
```

Is this not easier?
No need to cache the full list of PostFlagReason records four times.

app/models/post_flag_reason.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  # Cached check for whether a reason requires a parent id
+  def self.needs_parent_id?(reason_name)
+    needs_parent_id_map = Rails.cache.fetch("post_flag_reasons:for_needs_parent_id") { ordered.pluck(:name, :needs_parent_id).to_h }
+    !!needs_parent_id_map[reason_name.to_s]
```
Sindrake:
```suggestion
    by_name(reason_name)&.needs_parent_id? || false
```

Same as above.

app/models/post_flag_reason.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+class PostFlagReason < ApplicationRecord
+  validates :name, presence: true, uniqueness: { case_sensitive: false }, exclusion: { in: %w[deletion] }
+  validates :reason, presence: true
+  validates :index, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :target_date_kind, inclusion: { in: %w[before after], allow_blank: true }
+
```
Sindrake, line 11:
```suggestion
  validates :target_date_kind, presence: true, if: -> { target_date.present? }
  validates :target_date, presence: true, if: -> { target_date_kind.present? }

```
I would add validation to make sure that `target_date` and `target_date_kind` are not set independently of each other.

app/models/post_flag_reason.rb
```diff
@@ -0,0 +1,63 @@
...
+  after_destroy -> { self.class.invalidate_cache }
+  after_save -> { self.class.invalidate_cache }
+
+  scope :ordered, -> { order(index: :asc, id: :asc) }
+
+  def applies_to_post?(post)
+    grandfathered_tag = target_tag&.strip
+    if grandfathered_tag.present?
+      if grandfathered_tag[0] == "-"
```
Sindrake, line 20: Might want to add a validation to ensure that a `grandfathered_tag` cannot be set to just `-`.

spec/models/post/approval_methods_spec.rb
```diff
@@ -139,6 +139,7 @@
 
       it "resolves all open flags on the post" do
         post = create(:flagged_post)
+        create(:post_flag_reason)
```
Sindrake: The repeated `create(:post_flag_reason)` every time you need to create a post flag is noisy and error-prone.
Consider adding a `before(:create)` hook to the post flag factory, or something to that extent.

app/views/post_flag_reasons/_reason_row.html.erb
```diff
@@ -0,0 +1,15 @@
+<% depth ||= 0 %>
+<% css_class = depth > 0 ? "child-reason depth-#{depth}" : "" %>
```
Sindrake: This PR does not include nested reasons.
Please, clean this up for clarity.

db/migrate/20260505200714_create_post_flag_reasons.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class CreatePostFlagReasons < ActiveRecord::Migration[6.1]
```
Sindrake:
```suggestion
class CreatePostFlagReasons < ActiveRecord::Migration[8.1]
```

db/fixes/133_populate_post_flag_reasons_from_config.rb
Sindrake, line 1: This file needs to be renamed to `133_populate_post_flag...`

app/javascript/src/js/pages/post_flags/post_flags.js
```diff
@@ -25,8 +25,14 @@ PostFlags.initFlagForm = function () {
...
 
+  function onReasonRadioChange (event) {
+    console.log("target", event.target);
+    updateNoteRequired();
+  }
+
+  //   Set whether the note field is required
   let isNoteRequired = false;
   function updateNoteRequired () {
```
Sindrake: Note that this function does not seem to work anymore.
The "(required)" indicator does not appear.

Sindrake (changes requested):

db/migrate/20260505200714_create_post_flag_reasons.rb
DRLaDRLa: What's this number anyway? I couldn't figure out what it's supposed to be.

app/models/post_flag_reason.rb
DRLaDRLa: Yeah, honestly I wasn't sure how to use the caches correctly so I hoped you'd tell in the CR.

app/views/post_flag_reasons/_reason_row.html.erb
DRLaDRLa: I knew I had missed some.

spec/models/post/approval_methods_spec.rb
DRLaDRLa: I have a really hard time understanding how this factory stuff works, but I think I found a solution.

DRLaDRLa:
I think I should move the AI automatic flag reason to the settings instead of config, and have it selectable on the flag reasons index.
Edit: I'll do that as a follow up instead.

app/controllers/post_flag_reasons_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def edit
+    @reason = PostFlagReason.find(params[:id])
+  end
+
+  def create
+    PostFlagReason.transaction do
+      @reason = PostFlagReason.create(reason_params)
+      ModAction.log(:flag_reason_create, { reason: @reason.reason, text: @reason.text })
```
Sindrake:
```suggestion
      ModAction.log(:flag_reason_create, { reason: @reason.reason, text: @reason.text }) if @reason.valid?
```
Without this, it will create mod actions on invalid flag reasons.
See: `update`, it has this guard.

app/views/post_flag_reasons/index.html.erb
```diff
@@ -0,0 +1,32 @@
...
+        <th>Staff Reason?</th>
+        <th>Target Date?</th>
+        <th>Target Tag?</th>
+        <th>Updated</th>
+        <th></th>
+      </tr>
+      </thead>
+
+      <tbody>
+      <%= render partial: "reason_row", collection: @reasons, as: :flag_reason, locals: { depth: 0 } %>
```
Sindrake:
```suggestion
      <%= render partial: "reason_row", collection: @reasons, as: :flag_reason %>
```
Don't have depth anymore.

spec/models/post_flag/validations_spec.rb
```diff
@@ -118,70 +118,70 @@ def as_user(user)
...
+    it "is invalid for 'needs_parent_id' when the parent post does not exist" do
+      flag = build(:needs_parent_id_post_flag, reason: "something")
       flag.parent_id = 99_999_999
       expect(flag).not_to be_valid(:create)
       expect(flag.errors[:parent_id]).to include("must exist")
     end
 
-    it "is invalid for 'uploading_guidelines' when the post has the grandfathered_content tag" do
-      skip "uploading_guidelines reason not configured" unless Danbooru.config.flag_reasons.any? { |r| r[:name].to_s == "uploading_guidelines" }
+    it "is invalid for targe tag '-grandfathered_content' when the post has the grandfathered_content tag" do
```
Sindrake:
```suggestion
    it "is invalid for target tag '-grandfathered_content' when the post has the grandfathered_content tag" do
```

Nitpick.

Sindrake (changes requested):
Minor bug, and a couple nitpicks.
Otherwise, looks good to ship.

Sindrake:
Bumping this to next week, since tomorrow's release is getting crowded.

DRLaDRLa:
> Bumping this to next week, since tomorrow's release is getting crowded.

Yeah, it's a bit too high risk to merge this late, and I'll need to rebase again.

Sindrake (approved):

Issue #1780, opened 2026-04-07
Metatags that cause changes on other posts should add an edit reason to affected posts


mm12:
Metatags that cause changes to another post's properties generate a PostVersion on that post. Similar to e621ng/e621ng#1707, an edit reason specifying that the action was indirect would be beneficial. 

Metatags that effect other posts: 
`child`: `child:{id}`, `-child:{id}`, and `child:{none}`

PR #1923, opened 2026-04-28
[Posts] Add indirect edit reasons for child metatags

PlutonusDev:
When adding or removing a child using metadata tags, add an edit reason to the child/ren.

Implements #1780

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -981,19 +981,19 @@ class Post < ApplicationRecord
 
         when /^child:none$/i
           children.each do |post|
-            set_remove_child_edit_reason(post)
+            remove_child_edit_reason(post)
             post.update!(parent_id: nil)
           end
 
         when /^-child:(.+)$/i
           children.numeric_attribute_matches(:id, $1).each do |post|
-            set_remove_child_edit_reason(post)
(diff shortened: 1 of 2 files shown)


PlutonusDev:

app/models/post.rb
```diff
@@ -1476,6 +1479,14 @@ def set_merge_edit_reason
       return unless parent_id.present?
       parent.edit_reason = "Merged from post ##{self.id}"
     end
+
+    def set_remove_child_edit_reason(post)
```
Sindrake: Rubocop should complain about the names of these methods, since they go against the [Naming/AccessorMethodName](https://docs.rubocop.org/rubocop/latest/cops_naming.html) cop.
This rule is disabled for the `post.rb` file because there are a bunch of other methods that break it.
But we should be fixing issues like this, not propagating them.

Sindrake (changes requested):
Looks good.
Minot style nitpick.

app/models/post.rb
PlutonusDev: Resolved in 4d7ba37.

Sindrake (dismissed):

Issue #1763, opened 2026-04-04
[Uploads] Reject sources that are not URLs


jmortiger:
Add HTML form validation to reject sources that are not URLs

PR #1925, opened 2026-04-28
[Uploads] Reject sources that are not URLs

mint-dog:
This fulfills #1763. It uses the standard JS URL parser, but we could be more specific if desired.

Screenshot:
image


mint-dog:

PR #1926, opened 2026-04-28
[Uploader] Reject non-URL sources

mint-dog:
This fulfills #1763. It uses the standard JS URL parser, but we could be more specific if desired.

Screenshot:
image


app/javascript/src/js/pages/uploads/new/sources.vue
```diff
@@ -93,18 +96,30 @@
...
           return source.length > 0;
         }).length;
         return !this.noSource && (validSourceCount === 0);
       },
+      nonUrlSourceWarning: function() {
+        const nonUrlSourceCount = this.sources.filter(function (source) {
+          return source.length > 0 && !URL.canParse(source)
+        }).length;
+        return nonUrlSourceCount > 0;
+      },
```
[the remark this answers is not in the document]
mint-dog: Applied. Thank you robot

app/javascript/src/js/pages/uploads/new/sources.vue
```diff
@@ -93,18 +96,30 @@
...
     watch: {
-      sourceWarning: {
+      missingSourceWarning: {
+        immediate: true,
+        handler() {
+          this.$emit("missingSourceWarning", this.missingSourceWarning);
+        }  
+      },
+      nonUrlSourceWarning: {
         immediate: true,
```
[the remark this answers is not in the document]
mint-dog: I think you're standing too close to a magnet, mister robot. That syntax should be valid

app/javascript/src/js/pages/uploads/new/sources.vue
```diff
@@ -93,18 +96,30 @@
...
         const validSourceCount = this.sources.filter(function (source) {
           return source.length > 0;
         }).length;
         return !this.noSource && (validSourceCount === 0);
       },
+      nonUrlSourceWarning: function() {
+        const nonUrlSourceCount = this.sources.filter(function (source) {
+          return source.length > 0 && !URL.canParse(source)
+        }).length;
+        return nonUrlSourceCount > 0;
```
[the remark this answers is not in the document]
mint-dog: I'll leave this comment open for a human reviewer. Do we care about this?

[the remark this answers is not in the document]
Sindrake: Yeah, it's pretty new.
We don't have a strictly defined policy for this kind of thing, but my normal approach is "we need to support browsers that were released in the past 5 years".

https://caniuse.com/?search=URL.canParse
Looks like it was added in 2023.

app/javascript/src/js/pages/uploads/new/sources.vue
```diff
@@ -93,18 +96,38 @@
       },
     },
     computed: {
-      sourceWarning: function() {
+      missingSourceWarning: function() {
         const validSourceCount = this.sources.filter(function (source) {
           return source.length > 0;
         }).length;
```
Sindrake, line 102: `URL.canParse()` accepts any syntactically valid URL, including `javascript:`, `data:`, and `ftp:` schemes. The server only accepts `http`/`https` (see `post_replacement.rb:45`), so a non-http URL will pass this client-side check but be rejected server-side with a confusing error.

Consider tightening the filter to match the server expectation:
```js
const url = URL.canParse(source) && new URL(source);
return source.length > 0 && !(url && (url.protocol === "http:" || url.protocol === "https:"));
```
This would also let the error message be more specific — e.g. _"The source must be an http or https URL."_

mint-dog: I fixed this in a commit pushed just seconds ago. Great minds and all that!

mint-dog: I'll update the error message

app/javascript/src/js/pages/uploads/new/sources.vue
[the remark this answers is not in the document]
mint-dog: Okay, I can switch to regex to be safe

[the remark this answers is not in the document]
mint-dog: done

app/javascript/src/js/pages/uploads/new/sources.vue
mint-dog: done

app/javascript/src/js/pages/uploads/new/sources.vue
```diff
@@ -93,18 +96,35 @@
...
         return !this.noSource && (validSourceCount === 0);
       },
+      nonUrlSourceWarning: function() {
+        if (this.noSource) return false;
+
+        const nonUrlSourceCount = this.sources.filter(function (source) {
+          // Basic regex to check if the source starts with http or https and is followed by a valid domain structure
+          const httpUrlRegex = /^(https?:\/\/)([a-z0-9-]+\.)+[a-z]{2,}(:\d+)?(\/[^\s]*)?$/i;
```
DRLaDRLa: Parsing a URL correctly using regex is really complicated, this will still allow invalid URLs. Why not use `new URL(url)` surrounded by a `try/catch`? That's equivalent to `URL.canParse`: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#url. That API became widely available in April 2021, so it's exactly 5 years old.

mint-dog: I'm happy to use whichever of the two that the formal dev team would like. If you look earlier in this PR, I discussed this with Sindrake; I actually originally used URL.canParse. We could put that in a try/catch, but what behavior do we want when the user's browser lacks the support and we get an exception?

As for the regex, I am not  real experienced with regex, but my understanding is that this regex string is pretty solid. Do you have a significant counterexample? Presumably I could just adjust the regex if so. I played around with URL.canParse earlier, and it seemed much more lenient than the regex.

DRLaDRLa: Mainly it's reinventing the wheel, so to speak. The [RFC that defines URLs](https://datatracker.ietf.org/doc/html/rfc1738) is a long document, you can't implement it with a regex. So let `new URL` do it for you.

But more specifically it doesn't handle alternate encodings (non-normalized forms). 
You can try it here: https://jsfiddle.net/8e2kv4L9/.
These are valid URLs that shouldn't be rejected. https://regex101.com/r/lETGJk/1

mint-dog: I hear you. What should happen when an exception is thrown there, then? Use the regex as a fallback?

DRLaDRLa: The URL constructor throws when the URL string is invalid. The URL simply shouldn't be accepted.

mint-dog: Oh, I see what you mean with treating an exception as failed validation in general. I'll just use what's in that jsfiddle link

mint-dog: Done

app/javascript/src/js/pages/uploads/new/sources.vue
```diff
@@ -93,18 +96,37 @@
...
         const validSourceCount = this.sources.filter(function (source) {
           return source.length > 0;
         }).length;
         return !this.noSource && (validSourceCount === 0);
       },
+      nonUrlSourceWarning: function() {
+        if (this.noSource) return false;
+
+        const nonUrlSourceCount = this.sources.filter(function (source) {
+          let isValidUrl = false;
```
Sindrake, line 109: The updated logic looks correct — `new URL()` + protocol check handles `javascript:`, `data:`, and `ftp:` properly, and the early `if (this.noSource) return false` is a nice readability improvement.

Sindrake (approved):

PR #1928, opened 2026-04-28
[Posts] Editor source URL client-side validation and backend validation improvements

DRLaDRLa:
- Add client-side validation for source URLs in the post editor
- Improve backend validation to better handle weird URLs, and just reduce the amount of regex usage in favour of proper URL parsing.

### DIFF
--- a/app/helpers/posts_helper.rb
+++ b/app/helpers/posts_helper.rb
@@ -19,18 +19,37 @@ module PostsHelper
     end
   end
 
-  def post_source_tag(source)
+  def try_parse_http_url(source)
+    # This will do a better job at handling technically invalid URLs like
+    # http:example.com, http:/example.com or just example.com
+    url = Addressable::URI.heuristic_parse(source)
     # Only allow http:// and https:// links. Disallow javascript: links.
-    if source =~ %r{\Ahttps?://}i
-      source_link = decorated_link_to(source.sub(%r{\Ahttps?://(?:www\.)?}i, ""), source, target: "_blank", rel: "nofollow noreferrer noopener")
+    if %w[http https].include?(url.scheme)
+      url
+    end
+  rescue Addressable::URI::InvalidURIError
+    nil
+  end
+
+  def post_source_tag(source)
+    if source.start_with?("-")
+      tag.s(source[1..])
+    elsif (url = try_parse_http_url(source)).present?
+      # remove http(s): and any leading and trailing slashes
+      short_url = url.omit(:scheme).to_s.sub(%r{^/+}, "").sub(%r{/+$}, "")
+      source_link = decorated_link_to(short_url, url.to_s, target: "_blank", rel: "nofollow noreferrer noopener")
 
       if CurrentUser.is_janitor?
-        source_link += " ".html_safe + link_to("»", posts_path(tags: "source:#{source.sub(%r{[^/]*$}, '')}"), rel: "nofollow")
+        # remove ?query=test#example
+        url_no_final_path = url.omit(:query, :fragment)
+        # remove last /part
+        url_no_final_path.path = url_no_final_path.path.sub(%r{[^/]*$}, "")
+        # remove any remaining trailing slashes
+        url_no_final_path = url_no_final_path.to_s.sub(%r{/+$}, "")
+        source_link += " ".html_safe + link_to("»", posts_path(tags: "source:#{url_no_final_path}"), rel: "nofollow")
       end
 
       source_link
-    elsif source.start_with?("-")
-      tag.s(source[1..])
     else
       tag.span(source, class: "source-invalid")
     end
--- a/app/javascript/src/js/pages/posts/posts.js
+++ b/app/javascript/src/js/pages/posts/posts.js
@@ -552,28 +552,44 @@ Post.initialize_post_sections = function () {
     }
   });
 
-  const allUrlsValid = (urls) => {
-    if (!urls) {
-      urls = $("#post_source").val();
+  const isUrlValid = (url, ignoreUrls = []) => {
+    url = url.trim();
+    if (url.length <= 0) {
+      return true;
     }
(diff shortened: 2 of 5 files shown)


Sindrake:
~~Rejected in favor of #1926.~~
Forgot that this PR does other stuff, oops.

DRLaDRLa:
~~Reminder to apply this fix too: #1990~~ Done

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -549,6 +549,33 @@ Post.initialize_post_sections = function () {
...
+  const allUrlsValid = (urls) => {
+    if (!urls) {
+      urls = $("#post_source").val();
+    }
+    const lines = urls.split(/\r?\n/);
+    return lines.every(line => {
+      // Allow dead source links prefixed with `-`
+      if (line[0] === "-") line = line.substring(1);
+      try {
+        const url = new URL(line);
```
Sindrake: Empty source fields fail validation, as do blank lines.

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -549,6 +549,33 @@ Post.initialize_post_sections = function () {
...
+        return url.protocol === "http:" || url.protocol === "https:";
+      } catch {
+        // Exception occurs if the URL constructor fails to parse the string, which means it's not a valid URL
+        return false;
+      }
+    });
+  };
+
+  const updateForUrlChange = (event) => {
+    const validUrls = allUrlsValid(event.target.value);
```
Sindrake: This is correct when listening to `change.danbooru`, but not `danbooru:open-post-edit-tab`.
The latter is bound to `$(document)`, the `event.target.value` will be `undefined`.

This works accidentally right now, but probably just because of the previous bug.

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -549,6 +549,33 @@ Post.initialize_post_sections = function () {
...
+    $("#post-edit-invalid-url")[0].style.display = validUrls ? "none" : "";
+    $("#edit #form input[type=\"submit\"]")[0].disabled = !validUrls;
```
Sindrake: I think that the editor should only warn on invalid URLs to start with, not outright disable editing.
I am a little hesitant to outright break saving changes on a myriad of posts on e6 that already have invalid links.

app/helpers/posts_helper.rb
```diff
@@ -19,18 +19,37 @@ def prev_page_url
     end
   end
 
-  def post_source_tag(source)
+  def try_parse_http_url(source)
```
Sindrake, line 22: Please, add a spec file that tests the changed methods.

Sindrake (changes requested):

app/javascript/src/js/pages/posts/posts.js
DRLaDRLa: My thoughts was that users would fix, remove or "-" invalid urls on the next edit, intentionally forcing them to take a bit of time to cleanup.
I guess the editor could automatically prefix invalid entries with "-" when it first loads? That or it saves the existing list, then skips validation for matching entries in the new list.

app/javascript/src/js/pages/posts/posts.js
DRLaDRLa: It works because the callback does this:
```js
    if (!urls) {
      urls = $("#post_source").val();
    }
```
but honestly it's maybe just confusing, I'll just remove the argument.

app/javascript/src/js/pages/posts/posts.js
DRLaDRLa: Better solution: existing invalid sources will be allowed, but with a warning.

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -551,6 +551,49 @@ Post.initialize_post_sections = function () {
...
+    url = url.trim();
+    if (url.length <= 0) {
+      return true;
+    }
+    // Allow dead source links prefixed with `-`
+    if (url[0] === "-") {
+      url = url.substring(1);
+    }
+    // So existing invalid source links are skipped
+    if (ignoreUrls.includes(url)) {
```
Sindrake, line 561: Edge case here: since the `-` prefix is stripped before the `ignoreUrls` bypass check, dead-linked non-URL old sources will still block submission.

Example: if `oldSources` has `-example.com`, this check would fail.

app/javascript/src/styles/views/posts/show/partials/_edit_form.scss
```diff
@@ -9,3 +9,21 @@
...
+  border: 1px solid themed("color-foreground");
+  width: fit-content;
+  margin-bottom: 1em;
+  padding: 0.5em;
+  font-weight: bold;
+}
+
+#post-edit-invalid-url-error {
+  background-color: themed("color-danger-darken-10");
+  @include post-edit-invalid-url
```
Sindrake:
```suggestion
  @include post-edit-invalid-url;
```

app/javascript/src/styles/views/posts/show/partials/_edit_form.scss
```diff
@@ -9,3 +9,21 @@
...
+}
+
+#post-edit-invalid-url-warning {
+  background-color: themed("color-warning-darken-5");
+  @include post-edit-invalid-url
```
Sindrake:
```suggestion
  @include post-edit-invalid-url;
```

Sindrake (changes requested):
Minor nitpicks, otherwise looks clean.

app/javascript/src/js/pages/posts/posts.js
DRLaDRLa: Good catch.

Sindrake (approved):

PR #1930, opened 2026-04-29
[Onboarding] Initial implementation

### LINKED ISSUE #1770: [Users] Onboarding process
https://e621.net/forum_topics/63200?page=1#forum_post_488250

New users get a few screens to help better prep them for their experience. I'm thinking of an initial implementation using a single column on the `users` table (great excuse to add a JSON column) to track how deep they are in the onboarding process & if they chose to skip it.

#### Theme Selection

This is a fairly self-contained project, so it's a good option for someone outside the dev team interested in taking own a huge project, but it's still a big project.

PlutonusDev:
Implements a multi-step onboarding process for new users covering blacklists, privacy, and notifications. Designed to be non-blocking and easily extensible.

Implements #1770.


app/controllers/sessions_controller.rb
```diff
@@ -18,11 +18,17 @@ def create
...
       respond_to do |fmt|
-        fmt.html { redirect_to(url || posts_path) }
-        fmt.json { render(json: { url: url || posts_path }) }
+        if @user.onboarding_completed?
+          fmt.html { redirect_to(url || posts_path) }
+          fmt.json { render(json: { url: url || posts_path }) }
+        else
+          fmt.html { redirect_to(onboarding_path) }
+          fmt.json { render(json: { url: onboarding_path }) }
+        end
```
jmortiger: Since onboarding is only feasible in HTML (& I don't think you can make a new user through the API anyways), there's no value in not honoring a user's redirect attempt when creating a session through the API.
```suggestion
        if @user.onboarding_completed?
          fmt.html { redirect_to(url || posts_path) }
        else
          fmt.html { redirect_to(onboarding_path) }
        end
        fmt.json { render(json: { url: url || posts_path }) }
```

app/javascript/src/js/pages/onboarding/show.js
```diff
@@ -0,0 +1,92 @@
+import Page from "@/utility/Page";
+
+const Onboarding = {};
+
+Onboarding.init = function () {
+  const root = document.getElementById("onboarding-root");
+  if (!root) return;
+
+  // this sucks, will find a more robust way to do this in the future
```
jmortiger: If you want to get fancy, you can make this a secret form, make the buttons that trigger the AJAX request `submit` buttons, catch the `submit` event, `preventDefault`, & then asynchronously send the `FormData` object via JavaScript; then you can use the same strategy as [the settings page](https://e621.net/users/settings).

Going that far is probably needlessly obtuse, but you can still take a similar approach for retrieving the proper endpoint (having the server give it to you directly) & for what elements each step contains (using the fields of a form element); that would be more data-driven & flexible than solely allowing preset fields to be altered.

app/javascript/src/js/pages/onboarding/show.js
```diff
@@ -0,0 +1,92 @@
+import Page from "@/utility/Page";
+
```
jmortiger: Importing the flash utility for errors & such.
```suggestion
import Flash from "@/utility/Flash";

```

app/javascript/src/js/pages/onboarding/show.js
```diff
@@ -0,0 +1,92 @@
...
+      data: JSON.stringify({
+        user: {
+          blacklisted_tags: selectedTags, 
+          enable_privacy_mode: $("#enable_privacy_mode").prop("checked"),
+          disable_user_dmails: $("#disable_user_dmails").prop("checked"),
+          receive_email_notifications: $("#receive_email_notifications").prop("checked"),
+        },
+      }),
+    }).fail(err => {
+      console.error("Error updating settings:", err);
```
jmortiger: It might be best to have this flash the notification instead, or alongside it.

```suggestion
      Flash.error(`Error updating settings: ${err}`);
```

app/javascript/src/js/pages/onboarding/show.js
```diff
@@ -0,0 +1,92 @@
...
+        contentType: "application/json",
+        dataType: "json",
+        data: JSON.stringify({}),
+      })
+        .done(data => { 
+          console.log("Onboarding complete response:", data);
+          window.location.href = data.redirect_url || "/posts"; 
+        })
+        .fail((xhr, status, error) => { 
+          console.error("Onboarding complete failed:", status, error, xhr.responseText);
```
jmortiger: Same as above.
```suggestion
          Flash.error(`Onboarding complete failed: ${status} ${error}\n${xhr.responseText}`);
```

app/views/onboardings/show.html.erb
```diff
@@ -0,0 +1,102 @@
+<% content_for(:html_header) do %>
+  <%= vite_typescript_tag "v_onboarding" %>
+<% end %>
+
+<div id="c-onboardings"><div id="a-show">
+  <div class="onboarding-container">
+    <section class="section" id="onboarding-root" data-user-id="<%= @user.id %>">
+      <div class="section-header">
+        <h1>Welcome to e621, <%= CurrentUser.name %>!</h1>
```
jmortiger: We store the site's name in the config files; helps for downstream forks.
```suggestion
        <h1>Welcome to <%= Danbooru.config.app_name %>, <%= CurrentUser.name %>!</h1>
```

db/migrate/20260421120000_add_onboarding_completed_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddOnboardingCompletedToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :onboarding_completed, :boolean, default: false, null: false
```
jmortiger: I wouldn't immediately try & make this change (see main comment), but if we do track onboarding progress server-side, we'd likely either store it inside the user preferences bitflag or make it an integer that stores the step number they've completed; the latter would, just as an example, allow us to add steps & then put up an indicator if their stored number is less than the total number.

That said, I'd hold off on trying either of these changes until Cinder makes the call on storing this server-side at all.

app/controllers/sessions_controller.rb
```diff
@@ -18,11 +18,17 @@ def create
     session_creator = SessionCreator.new(request, session, cookies, sparams[:name], sparams[:password], sparams[:remember].to_s.truthy?)
 
     if session_creator.authenticate
+      @user = User.find(session[:user_id])
```
jmortiger: Minor optional nitpick, but I'd actually just tweak [`SessionCreator`](https://github.com/e621ng/e621ng/blob/master/app/logical/session_creator.rb) to allow retrieving the user; it's weird it doesn't already do that.
```suggestion
      @user = session.user
```
```rb
# app/logical/session_creator.rb:4
attr_reader :request, :session, :cookies, :name, :password, :remember, :user

# app/logical/session_creator.rb:17
@user = User.find_by_name(name)
```

jmortiger (changes requested):
_Disclaimer: I haven't run this locally._

As noted in the comments under the initial issue, there's been some internal disagreements about persisting the state of the onboarding process; that's on us for not updating the issue description to reflect this. However, that part of this could be easily removed should our lead deem it necessary, so I wouldn't worry too much about that.

It's good you didn't do this before, since we actually just entirely changed our test suite, but it's always a good idea to look into adding some tests when adding new functionality, & especially new endpoints. It's not a *requirement*, & I'd be lying if I said we did it as stringently as we should, so I'm not going to hold you to a standard we ourselves don't exemplify; just wanted to mention it for the future.

Other than that & the specific comments, this all looks wonderful; great job! 👍

app/controllers/sessions_controller.rb
PlutonusDev: Resolved in bd924e2.

app/javascript/src/js/pages/onboarding/show.js
PlutonusDev: Resolved in bd924e2.

app/javascript/src/js/pages/onboarding/show.js
PlutonusDev: Resolved in bd924e2.

app/javascript/src/js/pages/onboarding/show.js
PlutonusDev: Resolved in bd924e2.

app/views/onboardings/show.html.erb
PlutonusDev: Resolved in bd924e2.

app/controllers/sessions_controller.rb
PlutonusDev: Implemented and resolved in bd924e2.

app/controllers/onboardings_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  respond_to :html, :json
+
+  def show
+    @user = CurrentUser.user
+    respond_to do |format|
+      format.html { render :show }
+      format.json do
+        render json: {
+          user_id: @user.id,
+          steps: steps_with_user_values(@user)
```
jmortiger: Here's your typo; should be
```suggestion
          steps: steps_with_user_values(@user),
```

app/logical/session_creator.rb
```diff
@@ -14,7 +14,7 @@ def initialize(request, session, cookies, name, password, remember = false)
 
   def authenticate
     if User.authenticate(name, password)
-      user = User.find_by_name(name)
+      @user = User.find_by_name(name)
```
Sindrake: Is there a reason to do this?
The normal pattern is to just use `CurrentUser.user` – no need to pass the global variable around.

Also, making this a global variable breaks the code below (line 19).

app/controllers/onboardings_controller.rb
Sindrake:
```suggestion
          steps: steps_with_user_values(@user),
```
Pretty sure the controller would throw a SyntaxError without this.

app/controllers/onboardings_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      step_with_values = step.dup
+
+      case step[:type]
+      when "blacklist"
+        current_blacklist = user.blacklisted_tags.to_s.split(/\s+/)
+        step_with_values[:current_value] = current_blacklist
+      when "settings"
+        step_with_values[:fields] = step[:fields].map do |field|
+          field_dup = field.dup
+          field_dup[:current_value] = user.send(field[:id])
```
Sindrake: This seems like a bad idea.
`field[:id]` is config-controlled, so it's not catastrophic, but I would still add an explicit allowlist at the very list.

app/javascript/src/js/pages/onboarding/show.js
Sindrake, line 1: These days, we prefer TS over plain JS.

app/javascript/src/js/pages/onboarding/show.js
```diff
@@ -0,0 +1,234 @@
...
+      }
+      */
+
+      const linksContainer = $(`<div style="display: flex; gap: 10px;"></div>`);
+      step.links.forEach(link => {
+        const linkEl = $(`
+          <a href="${link.href}" class="btn btn-secondary" ${link.target ? `target="${link.target}"` : ""}>
+            ${escapeHtml(link.label)}
+          </a>
+        `);
```
Sindrake:
```suggestion
        const linkEl = $("<a class='st-button'>")
          .attr("href", link.href)
          .text(link.label);
```
I'm surprised that github has not flagged this as an XSS risk.
It's usually pretty anal about stuff like this.

Sindrake (changes requested):

db/migrate/20260421120000_add_onboarding_completed_to_users.rb
PlutonusDev: Removed this change in bc7be00.

app/controllers/onboardings_controller.rb
PlutonusDev: This has been resolved.

app/logical/session_creator.rb
PlutonusDev: This has been resolved.

app/controllers/onboardings_controller.rb
PlutonusDev: This has been resolved.

app/javascript/src/js/pages/onboarding/show.js
PlutonusDev: Refactored to TS in 0dc77d9.

app/controllers/onboardings_controller.rb
PlutonusDev: Resolved in 76c9521.

app/javascript/src/js/pages/onboarding/show.js
PlutonusDev: Resolved in 0dc77d9.

app/javascript/src/js/pages/onboarding/show.js
PlutonusDev: I do like this idea, and it's something to consider down the road if we need to scale the onboarding process. That said, I'm going to stick with the current implementation for now to keep this PR focused.

Thanks for the great feedback!

app/views/users/partials/edit/_basic.html.erb
```diff
@@ -31,6 +31,26 @@
...
+
+<tab-entry tab="<%= tab_name %>" group="account" class="with-button" search="onboarding restart welcome tutorial help">
+  <tab-head>Onboarding</tab-head>
+  <tab-body>
+    <%= link_to "Restart Onboarding Process", restart_onboarding_path, method: :post, class: "st-button" %>
+  </tab-body>
+  <tab-hint>
+    If you're feeling lost, you can restart the initial setup guide at any time.
+  </tab-hint>
+</tab-entry>
```
Sindrake, line 42: Duplicate?
```suggestion
```

app/javascript/src/js/pages/onboarding/show.ts
```diff
@@ -0,0 +1,250 @@
...
+interface OnboardingField {
+  id: string;
+  label: string;
+  help_text: string;
+  type: "checkbox";
+  current_value?: boolean;
+}
+
+interface OnboardingStep {
+  id: string;
```
Sindrake, line 21:
```suggestion
  id: string;
  type: "blacklist" | "settings" | "info";
```

Missing `type`, TS will error.

app/javascript/src/js/pages/onboarding/show.ts
```diff
@@ -0,0 +1,250 @@
...
+      url: "/onboarding.json",
+      dataType: "json",
+    })
+      .done((data: OnboardingResponse) => {
+        this.steps = data.steps;
+        this.renderSteps();
+        this.showStep(0);
+      })
+      .fail(() => {
+        E621.Toast.error("Failed to load onboarding steps. Please try again later.");
```
Sindrake:
```suggestion
        E621.Toast.alert("Failed to load onboarding steps. Please try again later.");
```

The `error` static method had been renamed to `alert`, to match Rails.

app/controllers/onboardings_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def complete
+    respond_to do |format|
+      format.json { render json: { success: true, redirect_url: posts_path } }
+    end
+  end
+
+  def restart
+    respond_to do |format|
+      format.html { redirect_to(onboarding_path, notice: "You have restarted the onboarding process") }
```
Sindrake:
```suggestion
      format.html { redirect_to(onboarding_path) }
```
No need to state the obvious.

app/controllers/users_controller.rb
```diff
@@ -209,6 +221,11 @@ def avatar_menu
     end
   end
 
+  def restart_onboarding
+    @user = CurrentUser.user
+    redirect_to onboarding_path, notice: "You have restarted the onboarding process"
+  end
+
```
Sindrake:
```suggestion
```

I don't think that this is used anywhere.

config/routes.rb
```diff
@@ -371,6 +371,12 @@
       get :avatar_menu
     end
   end
+  resource :onboarding, only: [:show] do
+    collection do
```
Sindrake:
```suggestion
    member do
```

More semantically correct for a singular resource, I think.

spec/requests/onboardings_controller_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+      let(:user) { create(:user) }
+
+      before do
+        sign_in_as user
+        ActionController::Base.allow_forgery_protection = false
+      end
+
+      after do
+        ActionController::Base.allow_forgery_protection = true
+      end
```
Sindrake: Not only is this unnecessary, it's actually harmful.
Forgery protection is disabled globally in the test suite. See: `config/environments/test.rb`.
By re-enabling it here, you may break other tests elsewhere.

spec/requests/onboardings_controller_spec.rb
```diff
@@ -0,0 +1,178 @@
...
```
Sindrake: Same as above.

spec/requests/onboardings_controller_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+
+  # ---------------------------------------------------------------------------
+  # POST /onboarding/restart — restart
+  # ---------------------------------------------------------------------------
+
+  describe "POST /onboarding/restart" do
+    context "as anonymous" do
+      it "returns 403 (forbidden)" do
+        post restart_onboarding_path(format: :html)
+        expect(response).to have_http_status(:forbidden)
```
Sindrake: Not quite sure how this is passing in CI.
Anonymous users are redirected to the login page (302) instead of returning 403.

config/danbooru_default_config.rb
```diff
@@ -892,6 +892,81 @@ def post_recommendations_enabled?
...
+        },
+        {
+          id: "theme",
+          title: "Site Personalization",
+          description: "Looking for a different look? #{app_name} supports multiple themes and custom CSS.",
+          type: "info",
+          links: [
+            { label: "Open Settings", href: "/users/settings", target: "_blank" },
+          ],
+        },
```
Sindrake: I would drop this step entirely.
It directs users away from onboarding mid-process, which could be confusing.

Sindrake (changes requested):

config/danbooru_default_config.rb
PlutonusDev: I agree. Removed this step in 3afd8fa.

spec/requests/onboardings_controller_spec.rb
PlutonusDev: Corrected in 3afd8fa.

spec/requests/onboardings_controller_spec.rb
PlutonusDev: Corrected in 3afd8fa.

spec/requests/onboardings_controller_spec.rb
PlutonusDev: Corrected in 3afd8fa.

config/routes.rb
PlutonusDev: Corrected in 3afd8fa.

app/controllers/users_controller.rb
PlutonusDev: Removed in 3afd8fa.

app/controllers/onboardings_controller.rb
PlutonusDev: Resolved in 3afd8fa.

app/javascript/src/js/pages/onboarding/show.ts
PlutonusDev: Resolved in 3afd8fa.

app/views/users/partials/edit/_basic.html.erb
PlutonusDev: Resolved in 3afd8fa.

app/javascript/src/js/pages/onboarding/show.ts
PlutonusDev: Resolved in 3afd8fa.

PR #1931, opened 2026-04-29
[Forum] Make topic update time consistent

### LINKED ISSUE #1793: [Forum] Editing a topic title updates inconsistently
Staff editing a topic title updates the user that last edited it, but not the time it was edited; ideally, neither would change

maremom:
This completes #1793 by making the update time consistent. While not the ideal solution in the long-term, it keeps consistency with the user being updated.

### DIFF
--- a/app/controllers/forum_topics_controller.rb
+++ b/app/controllers/forum_topics_controller.rb
@@ -68,7 +68,7 @@ class ForumTopicsController < ApplicationController
   def update
     check_privilege(@forum_topic)
     @forum_topic.assign_attributes(forum_topic_params)
-    @forum_topic.save touch: false
+    @forum_topic.save
     respond_with(@forum_topic)
   end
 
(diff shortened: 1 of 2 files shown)


maremom:

Sindrake:
This actually has a couple side effects that are worth considering.

* **Topic bumping.** `updated_at` is used for forum ordering — `sticky_first` and `default_order` both sort by it. A staff member editing a topic's title will now silently bump it to the top of the topic list, exactly like posting a new reply.
* **Unread marking.** `read_by?` and `mark_as_read!` compare `updated_at` against the user's last-read timestamps. A title edit will now mark the topic as unread for all users, triggering the "new post" indicator even though no new content exists.
* **Double timestamp write.** The `after_update :update_original_post` callback also sets `updated_at = Time.now` on the original `ForumPost`. So a title edit now touches timestamps on both the topic record and the original post.

Not the end of the world, but meaningful behavioral changes.
I am still of the opinion that the consistency is worth it.

Sindrake (approved):

PR #1937, opened 2026-04-29
[Posts] Add a comment count to mobile tab

### LINKED ISSUE #1935: [UI] Show number of comments on post view's comment tab for mobile
> [!NOTE]
> Split off from #1922 
 
Not too long after the recommendations feature came the tab split between tags and comments on posts on the mobile layout. The number of comments was removed on post thumbnails from the mobile layout in searches/post lists several months ago, so with comments being in their own tab now I think there's a good opportunity to have the number of comments on the tab itself instead to bring back that indicator in some way for the mobile layout.

For the comments tab, I wanted to suggest showing the number of comments on the tab button for comments itself.

**Additional context**
I have provided screenshots for demonstration purposes that were achieved by modifying elements on the page, not with any actual working code that I could provide.

For the comments demonstrations, I just manually added in the comments number to the element for the tab text to show what I mean. 

### [Comment Tab] Example 1
Image

### [Comment Tab] Example 2
Image

mint-dog:
Adds a count (including 0) to the comments tab under a post, for mobile. 
This fulfills #1935.
Screenshot:
image

### DIFF
--- a/app/views/posts/show.html.erb
+++ b/app/views/posts/show.html.erb
@@ -189,7 +189,7 @@
       >Tags</button>
       <% unless CurrentUser.hide_comments? %>
         <button
-          type="button" 
+          type="button"
           role="tab"
           class="post-mobile-tab"
           data-action="comments"
--- a/app/views/posts/show.html.erb
+++ b/app/views/posts/show.html.erb
@@ -189,13 +189,14 @@
       >Tags</button>
       <% unless CurrentUser.hide_comments? %>
         <button
-          type="button"
+          type="button" 
           role="tab"
           class="post-mobile-tab"
           data-action="comments"
           aria-controls="post-show-display"
           aria-selected="<%= selected_tab == "comments" %>"
-        >Comments</button>
+          aria-label="<%= "Comments (#{(@comments&.size || 0)})" %>"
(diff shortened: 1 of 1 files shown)


app/views/posts/show.html.erb
```diff
@@ -195,7 +195,8 @@
           data-action="comments"
           aria-controls="post-show-display"
           aria-selected="<%= selected_tab == "comments" %>"
-        >Comments</button>
+          aria-label="<%= "Comments (#{(@comments&.size || 0)})" %>"
+        ><%= "Comments (#{(@comments&.size || 0)})" %></button>
```
Sindrake:
```suggestion
        ><%= "Comments (#{@post.comment_count})" %></button>
```

The `aria-label` attribute is not necessary if its text will match the contents of the button.
I would also recommend using `@post.comment_count` instead of actually counting the comments. It's already saved on the post model, includes comments that are below the vote threshold, and should not be nil.

Sindrake (changes requested):

app/views/posts/show.html.erb
mint-dog: 👍  I committed your code

Sindrake (approved):

PR #1941, opened 2026-04-29
[DMails] Add staff link sharing

jmortiger:


app/javascript/src/js/pages/dmails/copy_links_to_clipboard.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  static init () {
+    const shareLink = document.querySelector<HTMLAnchorElement>(`a[href^=${document.location.pathname.replace(/\//g, "\\/")}\\?key]`);
+    if (!shareLink) return;
+    this.url = shareLink.href;
+    shareLink.onclick = (e) => {
+      e.preventDefault();
+      e.stopImmediatePropagation();
+      try {
+        navigator.clipboard.writeText(this.url).then(() => Flash.notice("Copied to clipboard!")).catch(() => this.catchCopyLinkToClipboardError);
```
Sindrake:
```suggestion
        navigator.clipboard.writeText(this.url)
          .then(() => Flash.notice("Copied to clipboard!"))
          .catch((e) => this.catchCopyLinkToClipboardError(e));
```

The current code is hard to read, and also has a bug - it does not pass the error to the handler.

app/models/dmail.rb
```diff
@@ -239,9 +240,69 @@ def update_recipient_unread_count
...
+  def visible_to?(user, key = nil)
+    owner_id == user.id ||
+      (user.is_staff? && (
+          is_accessible_by_key?(key) ||
+          (user.is_moderator? && is_accessible_to_moderators?) ||
+          (user.is_admin? && is_accessible_to_admins?)
+        ))
+  end
+
+  USE_VERSIONING = false
```
Sindrake: All of this is noise.
Please, don't include dead code in PRs that you want to be merged.

The only code that does any work here is this:
```Ruby
def generate_key
  "v1::#{Digest::SHA256.hexdigest("#{id}::#{from.id}::#{to.id}::#{body}")}"
end

def key_matches?(key)
  key == generate_key
end
```

app/models/dmail.rb
```diff
@@ -239,9 +240,69 @@ def update_recipient_unread_count
...
+    key.present? && (to.is_staff? || from.is_staff?) && key_matches?(key)
+  end
+
+  if USE_VERSIONING
+    def key_matches?(key)
+      (v = VERSION_REGEX.match(key)&.match(1)) && (key == generate_key(v))
+    end
+  else
+    def key_matches?(key)
+      key == generate_key
```
Sindrake: This is open to a timing attack.

```suggestion
      ActiveSupport::SecurityUtils.secure_compare(key, generate_key)
```

app/models/dmail.rb
```diff
@@ -239,9 +240,69 @@ def update_recipient_unread_count
...
+    def generate_key(...)
+      send(:"generate_key_v#{CURRENT_VERSION}")
+    end
+  end
+
+  private
+
+  # Includes the id of the DMail, so only works on that copy; if deleted, can't be used on the other party's copy. As there's no way to get their copy's id, currently doesn't matter.
+  def generate_key_v1
+    "v1::#{Digest::SHA256.hexdigest("#{id}::#{from.id}::#{to.id}::#{body}")}"
```
Sindrake: Raw SHA256 over known inputs.
Body being private is the only practical secret here, which seems fragile.

Rails provides `Rails.application.secret_key_base`. Proper practice would be something like this:


```suggestion
    OpenSSL::HMAC.hexdigest("SHA256", Rails.application.secret_key_base, "#{id}::#{from.id}::#{to.id}::#{body}")
```

app/javascript/src/js/pages/dmails/copy_links_to_clipboard.ts
```diff
@@ -0,0 +1,28 @@
...
+
+class DMails {
+  private static url?: string;
+  private static catchCopyLinkToClipboardError (error: any, url = this.url) {
+    console.warn("Failed! %o", error);
+    alert(`Failed to automatically copy sharable link.\n\nLink: ${url}`);
+  }
+
+  static init () {
+    const shareLink = document.querySelector<HTMLAnchorElement>(`a[href^=${document.location.pathname.replace(/\//g, "\\/")}\\?key]`);
```
Sindrake: This just seems fragile.
I would add a `data-share-link` attribute on the anchor in the ERB template, and use that instead.

app/controllers/dmails_controller.rb
```diff
@@ -15,20 +15,22 @@ def index
     respond_with @dmails.to_json
   end
 
+  # TODO: Test endpoint with key
```
Sindrake: Actually test the endpoint with a key, rather than leave TODOs.

spec/models/dmail/permissions_spec.rb
```diff
@@ -58,44 +83,44 @@ def make_dmail(overrides = {})
...
-      #   # Create a ticket referencing this dmail, bypassing ticket validations
-      #   ticket = Ticket.new(qtype: "dmail", disp_id: dmail.id,
-      #                       reason: "Test report", accused_id: sender.id)
-      #   ticket.creator_id = moderator.id
-      #   ticket.save!(validate: false)
-      #
-      #   expect(dmail.visible_to?(moderator)).to be true
-      # end
+      it "is only visible to a moderator when an associated dmail Ticket exists" do
+        # TODO: After the Ticket factory gets made, use factory instead
```
Sindrake, line 103: We do have a Ticket factory.

Sindrake (changes requested):

app/models/dmail.rb
jmortiger: This is prep for when we overhaul DMails to kill this multiple copy stuff, but ok, I'll trim it.

app/models/dmail.rb
jmortiger: The point is that if you already know what the DMail says, who it's from, & who it's to, then you already know everything; what are you protecting at that point?

I'll add it if you really want, but it seems like curing the symptom of the proposed scenario instead of the actual problem.

spec/models/dmail/permissions_spec.rb
jmortiger: I was going off what was there prior. I'll update it.

Sindrake (approved):

PR #1942, opened 2026-04-30
[WikiSearch] Unhandled ActiveRecord::RangeError when out-of-range integer passed as search param

### LINKED ISSUE #1845: Unhandled ActiveRecord::RangeError when out-of-range integer passed as search param
## Summary

Passing an integer value that exceeds PostgreSQL's `integer` range (max ~2.1 billion) as a search parameter triggers an unhandled `ActiveRecord::RangeError`, resulting in a 500 instead of a user-facing 400/404.

## Stack Trace

```
ActiveRecord::RangeError
PG::NumericValueOutOfRange: ERROR: value "995859912741" is out of range for type integer

app/logical/danbooru/paginator/active_record_extension.rb:54:in `records'
app/views/wiki_page_versions/_page_listing.html.erb:20:in `each_with_index'
app/controllers/wiki_page_versions_controller.rb:8:in `index'
```

## Triggering Request

```
GET /wiki_page_versions?search[wiki_page_id]=995859912741
```

## Root Cause

The `wiki_page_id` param is passed directly into a PostgreSQL query bound as an `integer`. The value `995859912741` exceeds the 32-bit integer limit, causing Postgres to raise `PG::NumericValueOutOfRange`. This is not rescued anywhere in the paginator or controller, so it surfaces as a 500.

## Expected Behavior

Out-of-range integer search params should be caught and treated as a no-results response or a user-facing 400/422 error. `ActiveRecord::RangeError` could also be rescued globally in `ApplicationController` alongside similar invalid-input errors.

mint-dog:
Ensure wiki_page_id is a safe int, and rescue ActiveRecord::RangeError high above (for safety).
This fulfills #1845.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -91,6 +91,8 @@ class ApplicationController < ActionController::Base
       render_error_page(500, exception, message: "The database timed out running your query.")
     when ActionDispatch::Http::Parameters::ParseError, ActionController::BadRequest, PostVersion::UndoError
       render_error_page(400, exception)
+    when ActiveRecord::RangeError
+      render_expected_error(400, "Invalid request parameters")
     when SessionLoader::AuthenticationFailure
       session.delete(:user_id)
       cookies.delete :remember
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -91,8 +91,6 @@ class ApplicationController < ActionController::Base
       render_error_page(500, exception, message: "The database timed out running your query.")
     when ActionDispatch::Http::Parameters::ParseError, ActionController::BadRequest, PostVersion::UndoError
       render_error_page(400, exception)
-    when ActiveRecord::RangeError
-      render_expected_error(400, "Invalid request parameters")
     when SessionLoader::AuthenticationFailure
       session.delete(:user_id)
       cookies.delete :remember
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -91,6 +91,8 @@ def rescue_exception(exception)
       render_error_page(500, exception, message: "The database timed out running your query.")
     when ActionDispatch::Http::Parameters::ParseError, ActionController::BadRequest, PostVersion::UndoError
       render_error_page(400, exception)
+    when ActiveRecord::RangeError
+      render_expected_error(400, "Invalid request parameters")
```
Sindrake:
```suggestion
```

I don't think that a global guard like this is a good idea.
`ActiveRecord::RangeError` fires when malformed input reaches a DB query unguarded. That's a bug at the call site, not an expected user-error condition.

Rescuing it globally converts what would be a logged 500 (visible in the ExceptionLog) into a silent 400. That means that we would not be able to detect unsanitized inputs like this effectively.

Sindrake (changes requested):

app/controllers/application_controller.rb
mint-dog: Makes sense. Committed your change

Sindrake (approved):

PR #1961, opened 2026-05-02
[Posts] Rework automatically added video tags

### LINKED ISSUE #1345: Rework automatically added video tags
Right now, all posts with webm video files receive both `webm` and `animated` tags.
There is also a `video` tag, which is aliased to `animated`.

This was somewhat acceptable when we only supported the webm format, but now that mp4 is supported as well it's causing a mess. The way I see it, here is the best course of action:
1. Un-alias `video` from `animated`.
2. [Automatically add](https://github.com/e621ng/e621ng/blob/f528781c1f14cc4101881e89bbc37dd375c3f502/app/models/post.rb#L804) the `video` tag to [any video posts](https://github.com/e621ng/e621ng/blob/f528781c1f14cc4101881e89bbc37dd375c3f502/app/logical/file_methods.rb#L32-L34).
3. Re-direct both `webm` and `mp4` in searches to `type:webm` and `type:mp4`.

This would also help solve issues like #1341.

yivarld:
Closes #1345

### DIFF
--- a/app/logical/tag_query.rb
+++ b/app/logical/tag_query.rb
@@ -1509,6 +1509,7 @@ class TagQuery
       add_to_query(type, :filetype, tag)
       true
     end
+    nil
   end
 
   # Adds the tag to the query object based on its prefix and if it contains a wildcard.
--- a/app/logical/tag_query.rb
+++ b/app/logical/tag_query.rb
@@ -1507,7 +1507,7 @@ class TagQuery
   def intercept_metatag_alias(tag, type)
     if FileMethods::FILE_TYPE.value?(tag)
       add_to_query(type, :filetype, tag)
-      true
+      return true
     end
     nil
   end
--- a/app/logical/tag_query.rb
+++ b/app/logical/tag_query.rb
@@ -1185,10 +1185,15 @@ class TagQuery
 
   private
 
-  METATAG_SEARCH_TYPE = Hash.new(:must).merge({
+  TAG_SEARCH_TYPE = Hash.new(:must).merge({
     "-" => :must_not,
     "~" => :should,
   }).freeze
+
+  def get_tag_name_with_search_type(tag)
+    type = TagQuery::TAG_SEARCH_TYPE[tag[0]]
+    [(type == :must ? tag : tag[1..]).downcase, type]
+  end
 
   # The maximum number of nested groups allowed before either cutting off processing or triggering a
   # `TagQuery::DepthExceededError`.
@@ -1286,7 +1291,7 @@ class TagQuery
           next if group.blank?
           q[:children_show_deleted] ||= !group.hide_deleted_posts?(at_any_level: true) if kwargs[:process_groups]
           q[:groups] ||= {}
-          search_type = METATAG_SEARCH_TYPE[match[1]]
+          search_type = TAG_SEARCH_TYPE[match[1]]
           q[:groups][search_type] ||= []
           q[:groups][search_type] << group
         else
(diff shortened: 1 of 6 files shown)


yivarld:

app/models/post.rb
```diff
@@ -832,7 +832,7 @@ def remove_negated_tags(tags)
     def add_automatic_tags(tags)
       return tags unless Danbooru.config.enable_dimension_autotagging?
 
-      tags -= %w[thumbnail low_res hi_res absurd_res superabsurd_res huge_filesize wide_image tall_image long_image flash webm mp4 long_playtime short_playtime]
+      tags -= %w[thumbnail low_res hi_res absurd_res superabsurd_res huge_filesize wide_image tall_image long_image flash webm mp4 video long_playtime short_playtime]
```
jmortiger: Now that we automatically convert the tags for all valid file extensions automatically, none of them should be tagged.
```suggestion
      tags -= %w[thumbnail low_res hi_res absurd_res superabsurd_res huge_filesize wide_image tall_image long_image flash video long_playtime short_playtime].push(*FileMethods::FILE_TYPE.values)
```

jmortiger (commented):
This is wonderful work, & I appreciate the consideration & foresight to place the constant in `FileMethods`. Only notes I have would be that I'd like a test for the autotagging behavior & that I'd like to see the rest of the filetypes accounted for there. As for the test, this *should* work, assuming the other suggestion was added:
```rb
# spec/requests/post_controller_spec.rb:250

      context "editing autotags" do
        it "correctly removes filetype tags" do
          filetype_tags = %w[jpg png webm mp4 swf webp gif flash video animated_gif animated_png].freeze
          post = create(:post, file_ext: "jpg")
          put post_path(post, format: :json), params: { post: { tag_string_diff: filetype_tags.join(" ") } }
          expect(response).to have_http_status(:ok)
          post.reload
          expect(post.tag_array).not_to include(*filetype_tags)
        end
      end
```

app/models/post.rb
yivarld: Oh that is true! Didn't think of that.

app/logical/tag_query.rb
```diff
@@ -1497,41 +1503,52 @@ def parse_query(query, depth: 0, **kwargs)
   # Same as `TagQuery::REGEX_VALID_TAG_CHECK`, but disallows `*`
   REGEX_VALID_TAG_CHECK_2 = /[\*\,\#\$\%\\]/
 
+  # Checks if a certain tag should be transformed into a metatag, and adds it accordingly if so.
+  def intercept_metatag_alias(tag, type)
+    if FileMethods::FILE_TYPE.value?(tag)
+      add_to_query(type, :filetype, tag)
+      true
+    end
+  end
```
Sindrake:
```suggestion
      true
    end
    nil
  end
```
I would return `nil` explicitly, rather than implicitly.

spec/requests/posts_controller_spec.rb
```diff
@@ -248,6 +248,17 @@
         expect(response.parsed_body).to include("post" => hash_including("id" => post.id))
       end
 
+      context "editing autotags" do
```
Sindrake: I feel like instead of having this in the controller spec, it may be better to have a more atomic test for `add_automatic_tags` method specifically – making sure it strips the filetype tags, and adds the `video` tag when appropriate.

There are already a couple tests for it in `spec/models/post/tag_methods_spec.rb`.

Sindrake (changes requested):

Sindrake:
BUR: https://e621.net/bulk_update_requests/14104

Sindrake (approved):

PR #1972, opened 2026-05-05
[UI] Change post thumbnail colored borders to corner ribbons

yivarld:
Removes the post thumbnail borders and replaces them with ribbons, similar to the ones in re621:
image


yivarld:
Made some slight tweaks;
- Badges are properly shifted down to let the left ribbon exist.
- Tooltips are shown if hovered over the ribbons, showing what they mean.
image

app/components/post_thumbnail_component.rb
```diff
@@ -63,16 +68,46 @@ def preview_classes
...
+  }.freeze
+
+  def get_ribbon_background_position_for_conditions(condition_a, condition_b)
+    RIBBON_BACKGROUND_POSITION_MAPPING[[condition_a, condition_b]] || false
+  end
+
+  def get_ribbon_background_position(side)
+    case side
+    when :left
+      get_ribbon_background_position_for_conditions(post.has_visible_children?, post.parent_exists?)
```
Sindrake:
```suggestion
      get_ribbon_background_position_for_conditions(post.has_visible_children?, post.parent_id.present?)
```

`parent_exists?` method looks up the parent post in the DB, which results in an N+1 query problem.

app/components/post_thumbnail_component.rb
```diff
@@ -63,16 +68,46 @@ def preview_classes
...
+    end
+  end
+
+  def get_ribbon_tooltip(side)
+    content = []
+
+    case side
+    when :left
+      content << "Has children post/s." if post.has_visible_children?
+      content << "Has parent post." if post.parent_exists?
```
Sindrake:
```suggestion
      content << "Has parent post." if post.parent_id.present?
```
Same as above.

app/javascript/src/styles/base/_palette.scss
```diff
@@ -34,12 +34,10 @@ $palette: (
   "text-grey": #aaaaaa,
 
   // Just here for the post borders, nothing else
-  "plain-white": #fff,
-  "plain-black": #000,
-  "plain-green": #0F0,
-  "plain-yellow": #CC0,
-  "plain-blue": #00F,
-  "plain-red": #F00,
+  "plain-green": #65ff3f,
```
Sindrake:
```suggestion
  "plain-white": #fff,
  "plain-black": #000,
  "plain-green": #65ff3f,
```
`plain-white` and `plain-black` are still used in a few places in the repo.

app/components/post_thumbnail_component.rb
```diff
@@ -4,6 +4,11 @@ class PostThumbnailComponent < ViewComponent::Base
   include IconHelper
   with_collection_parameter :post
 
+  RIBBON_SIDE = {
+    left: "left",
+    right: "right",
+  }.freeze
```
Sindrake:
```suggestion
  RIBBON_SIDE = %i[left right].freeze
```
Just use `side.to_s` to get the CSS class in the template.

app/components/post_thumbnail_component.rb
```diff
@@ -63,16 +68,46 @@ def preview_classes
...
+    [true, false] => "0%",
+    [false, true] => "100%",
+    [true, true] => "50%",
+  }.freeze
+
+  def get_ribbon_background_position_for_conditions(condition_a, condition_b)
+    RIBBON_BACKGROUND_POSITION_MAPPING[[condition_a, condition_b]] || false
+  end
+
+  def get_ribbon_background_position(side)
```
Sindrake: By convention, the `get_` prefix is typically omitted in Ruby.
`ribbon_background_position(side)` and `ribbon_tooltip(side)` read better.

Just a nitpick.

app/javascript/src/styles/components/post_thumbnail_component.scss
```diff
@@ -409,67 +411,38 @@ body[data-st-contain="true"] article.thumbnail img {
...
-}
-
-// Four states: all edges different
-article.thumbnail[data-border-states="4"] {
-  > a.thm-link {
-    border: 2px solid transparent;
-    border-color: $border-has-children $border-has-parent $border-is-flagged $border-is-pending;
-  }
+article div.ribbon {
+   --transform-scalar: 1; // changed to -1 for the transforms below
```
Sindrake:
```suggestion
  --transform-scalar: 1; // changed to -1 for the transforms below
```

... 3 spaces???

Sindrake (changes requested):

app/components/post_thumbnail_component.rb
yivarld: Oh that's sweet! Will do

app/components/post_thumbnail_component.html.erb
```diff
@@ -9,6 +8,16 @@
         <%= tag.img src: preview_urls[1], alt: alt_text, **image_attributes %>
       <% end %>
     <% end %>
+
+    <% RIBBON_SIDE.each do |side| %>
+      <% if should_show_ribbon? side %>
+        <div
+          class="ribbon <%= side.to_s %>"
+          style="--bg-pos: <%= ribbon_background_position side %>"
+          title="<%= ribbon_tooltip side %>">
```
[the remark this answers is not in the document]
yivarld: Not too sure whether this is a good idea or not since `title` is what is being used everywhere else. The `data-hover-text` attribute is not used anywhere. Will do if requested though!

Sindrake (changes requested):
# Code Review

**PR #1972**: [UI] Change post thumbnail colored borders to corner ribbons

## Summary

Solid follow-up that addresses most of the previous review's concerns (the parent_exists? N+1, RIBBON_SIDE constant naming, plain-white/plain-black restoration, get_ prefix). The ribbon UI itself is clean. Two real issues remain: a behavior regression in posts.js where Post.approve / Post.update_data still maintain the now-dead data-border-states attribute but no longer visually update the right ribbon when a post is approved, and the SCSS multi-line transform still uses the 3-space indentation that was flagged in the prior round.

## Critical Issues - Must Fix

1. **Approving a pending post no longer hides its right ribbon**
   - File: `app/javascript/src/js/pages/posts/posts.js`
   - Post.approve still does `removeClass("pending")` and decrements `data-border-states`, but the right ribbon is a server-rendered `<div class="ribbon right">` element whose presence is the only thing that controls visibility. Removing the class no longer hides the pending indicator. Previously, the bottom-blue border did disappear because the border CSS keyed off `.pending` + `[data-border-states]`. After this PR, approving a post from the listing leaves the right ribbon visually stuck until a hard reload. The dead `data-border-states` writes in Post.update_data (line 578-580) and Post.approve (line 799) should either be removed and replaced with explicit ribbon DOM manipulation, or noted as a known regression.
   - Fix: Either remove the dead data-border-states writes entirely (if you accept that approve no longer updates the visual indicator until reload), or update the JS to remove `article.thumbnail > a.thm-link > div.ribbon.right` on approval (and rebuild it conditionally when has_visible_children/parent_id change in update_data).

2. **Multi-line `transform:` still uses 3-space continuation**
   - File: `app/javascript/src/styles/components/post_thumbnail_component.scss`
   - Flagged in the previous review round. The transform block sits at 2-space class-body indentation, but the `translate(...)` / `rotate(...)` continuation lines are indented with 3 spaces, which is neither standard nesting (4) nor aligned-after-colon. Rest of the file uses 2-space increments consistently.
   - Fix: Bump the continuation lines to 4 spaces (one nesting level deeper) so the transform reads as a normal multi-line property.

## Important Issues - Should Fix

1. **Ribbon tooltip relies on `title` on a non-interactive `<div>`**
   - File: `app/components/post_thumbnail_component.html.erb`
   - The ribbon is an empty decorative `<div>` with a `title` attribute. Screen readers generally do not announce `title` on non-focusable, non-labeled elements, so the only signal that a post is pending / flagged / has parent / has children becomes invisible to AT users. The previous border-based system was also non-accessible, so this isn't strictly a regression, but it's worth either adding `role="img" aria-label="..."` to the ribbon `<div>` or surfacing the same information into the existing tooltip_text on the thm-link.
   - Suggestion: Add `role="img"` and `aria-label="<%= ribbon_tooltip side %>"` (alongside or instead of `title`), or merge the ribbon meanings into tooltip_text so it shows up on the link's hover_text.

2. **Joining multiple ribbon reasons with `\n` is browser-inconsistent**
   - File: `app/components/post_thumbnail_component.rb`
   - `ribbon_tooltip` joins reasons with `\n` and passes them through the HTML `title` attribute. Browsers vary on whether they preserve newlines in `title` - Chromium tends to collapse them to a space, Firefox honors them, native macOS varies. When a left ribbon shows both 'Has child posts.' and 'Has parent post.', users may see those run together or on separate lines depending on the browser.
   - Suggestion: Either join with `" "` (or `" / "`) for a consistent inline tooltip, or move to a real tooltip mechanism (data-hover-text + JS) so multi-line is reliable.

## Positive Notes

- Nice job picking up almost every item from the prior review - `parent_exists?` N+1 fixed, `RIBBON_SIDE = %i[left right]` symbols, restored `plain-white`/`plain-black`, dropped the `get_` prefix.
- The two-color gradient + background-position trick to handle 1- vs 2-state ribbons in a single element is a clean approach; reads better than four separate elements.
- Test coverage on `should_show_ribbon?` (single flag / both flags / shift-badge presence) is a nice addition over the previous border_state_count tests.

## Recommendation

Request changes - critical issues need to be addressed before merging.

yivarld:
I am not sure it is a great idea to make the ribbon title text appear in screen readers by adding the `aria-label` attribute. Screen readers would announce this for every single post, and I do not think that will be very useful for a user, especially since we also already have another title tag for the thumbnail itself, containing all the information regarding the post.

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -801,11 +803,10 @@ Post.approve = function (post_id, callback) {
       var message = $.map(data.responseJSON.errors, (msg) => msg).join("; ");
       E621.Toast.alert("Error: " + message);
     }).done(function () {
-      var $post = $(`article.thumbnail[data-id="${post_id}"]`).first();
+      const $post = Post.getThumbnail(post_id).first();
```
Sindrake:
```suggestion
      const $post = Post.getThumbnail(post_id);
```

`getThumbnail` already calls `first()` internally.

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -599,6 +596,10 @@ Post.tagScript = function (post_id, tags) {
   Post.update(post_id, { "post[tag_string_diff]": tag_string });
 };
 
+Post.getThumbnail = function (post_id) {
+  return $(`article.thumbnail[data-id="${post_id}"]`).first();
```
Sindrake: Actually, is there a reason to only pick the first thumbnail?
We currently don't have any pages where multiple posts with the same ID can be displayed, that is true. But it does not hurt to be slightly more future-thinking and account for that situation anyways. Should not be that complicated.

app/javascript/src/styles/components/post_thumbnail_component.scss
```diff
@@ -45,6 +45,7 @@ article.thumbnail {
   display: flex;
   flex-flow: column;
   align-items: center;
+  position: relative;
```
Sindrake: Ribbons are inside the `a.thm-link` element, not the `article.thumbnail`.
Does this actually do anything?

spec/components/post_thumbnail_component/preview_classes_spec.rb
```diff
@@ -79,34 +79,47 @@ def component(post, **options)
...
+
+    it "shifts down the badge if the left ribbon is present" do
+      allow(post).to receive(:has_visible_children?).and_return(true)
+      expect(component(post).send(:preview_classes)).to include("shift-badge")
     end
 
-    it "sums all four active state flags to 4" do
-      allow(post).to receive_messages(has_visible_children?: true, parent_id: 1, is_pending?: true, is_flagged?: true)
-      expect(component(post).send(:border_state_count)).to eq(4)
+    it "does not shift down the badge if the left ribbon is present" do
```
Sindrake:
```suggestion
    it "does not shift down the badge if the left ribbon is not present" do
```

app/components/post_thumbnail_component.html.erb
```diff
@@ -9,6 +8,16 @@
         <%= tag.img src: preview_urls[1], alt: alt_text, **image_attributes %>
       <% end %>
     <% end %>
+
+    <% RIBBON_SIDE.each do |side| %>
+      <% if should_show_ribbon? side %>
+        <div
+          class="ribbon <%= side.to_s %>"
+          title="<%= ribbon_tooltip side %>"
+          role="img">
```
Sindrake:
```suggestion
          aria-hidden="true"
        >
```

The comment you left regarding this is slightly incorrect.

This actually does cause screen readers to announce the ribbon regardless of whether `aria-label` is present: per ARIA, title is the accessible-name fallback for role="img".
I would just keep the ribbons fully-decorative by dropping the `role="img"` and adding `aria-hidden="true"`.

Sindrake (changes requested):

app/javascript/src/styles/components/post_thumbnail_component.scss
yivarld: Good catch! That is something I forgot to remove when i was originally experimenting on putting them inside the thumbnail element directly.

app/javascript/src/styles/themes/_palette_deutranomaly.scss
```diff
@@ -20,7 +20,6 @@ $palette: (
   "text-green": #9f98ff,
   "text-grey": #aaaaaa,
 
-  "plain-black": #000,
   "plain-green": #9e1fff,
   "plain-yellow": #CC0,
```
Sindrake: Note that while the default value for `plain-yellow ` was changed in the main `palette.scss` file, the deuteranomaly file retains the old one.

Please, verify whether this is correct, and that the corner ribbons work with all palettes.

Sindrake (changes requested):
Looks good.
Just one minor nitpick.

app/javascript/src/styles/themes/_palette_deutranomaly.scss
yivarld: I have adjusted the yellow for both color blindness palettes, especially so that it is as differentiable as possible from green (in the case you see a post thumbnail marked as has parent and children)

Sindrake (approved):

PR #1978, opened 2026-05-05
[Favorites] Fix an N+1 query problem by batch-loading uploader records

Sindrake:
This didn't actually hit the database most the time, but it did cause repeated cache calls.

### DIFF
--- a/app/logical/post_sets/favorites.rb
+++ b/app/logical/post_sets/favorites.rb
@@ -22,7 +22,10 @@ module PostSets
     def posts
       @post_count ||= ::Post.tag_match("fav:#{@user.name} status:any").count_only
       @posts ||= begin
-        favs = ::Favorite.for_user(@user.id).includes(:post).order(created_at: :desc).paginate_posts(page, total_count: @post_count, limit: @limit)
+        favs = ::Favorite.for_user(@user.id)
+                         .includes(post: :uploader)
+                         .order(created_at: :desc)
+                         .paginate_posts(page, total_count: @post_count, limit: @limit)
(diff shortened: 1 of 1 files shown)


Sindrake:
# Code Review

**PR #1978**: [Favorites] Fix an N+1 query problem by batch-loading uploader records

## Summary

Small, focused optimization that swaps per-post `User.id_to_name` cache lookups for batch-loading uploader records via `.includes(post: :uploader)` on the favorites listing. The change is correct, idiomatic Rails, and aligns with the existing pattern used in PostSets::Post (which already passes `includes: [:uploader]` to its paginator). Behavior is preserved: ordering, pagination, total_count, and the returned post collection are unchanged. `Post#uploader_name` already short-circuits to the loaded association when present, so this change actually flips it onto the fast path. Coverage on the file remains at 86.4% per octocov, and existing favorites controller specs exercise the happy path.

## Positive Notes

- Nested includes syntax `includes(post: :uploader)` is exactly right here - Favorite belongs_to :post, Post belongs_to :uploader.
- Consistent with the eager-loading pattern already used in app/logical/post_sets/post.rb (paginate_posts ... includes: [:uploader]).
- `Post#uploader_name` already checks `association(:uploader).loaded?` and falls back to `User.id_to_name` otherwise, so this change activates the existing fast path without touching the model. Clean separation of concerns.
- Behavior-preserving: only an `includes` is added, so ordering, pagination, and the returned PaginatedArray are untouched.

## Recommendation

Approve - the code looks good. Minor items can be addressed optionally.

PR #1981, opened 2026-05-06
[Frontpage] SVG counter digits

### LINKED ISSUE #1757: Make counters SVGs
I believe this is being somewhat worked on by staff (including @mm12), but the low-quality originals aren't helping. We may need to track down the original files before this can be done properly.

> [FR] [Assets] (very) Low priority: would be nice if the counters were high quality, perhaps even SVGs (someday) 
> Image

Parcly-Taxel:
We convert the old PNG files for the 10 digits shown on the front page to SVG files.

Closes #1757.


Sindrake:
There is something slightly cursed about SVGs having aliasing in them.

Sindrake (approved):

PR #1985, opened 2026-05-07
[Posts] Fix failed video conversions due to invalid colorspace values

Sindrake:
Post#6316864 failed to get processed due to an invalid metadata value.
Other videos with proper color space metadata should remain unaffected by this override.


Sindrake:
This actually just isn't worth fixing, unless these issues get more common.

PR #1988, opened 2026-05-08
[AutomodDmails] Permit replies to messages addressed to automod

DRLaDRLa:
Automod DMails now have:
- Respond
- Forward
- Mark as read/unread
- Report

These work as-if the DMail was your own.

Mark read/unread has different behaviour when the recipient is the automod: it reloads the page instead of redirecting to the index.

DMail permissions now allow all staff to view when the recipient is the automod, which is consistent with the new automod DMails route. This is necessary to make the actions work, because you need to be able to view a DMail to respond, forward or ticket it (otherwise it'd be trivial to leak the contents). This could be simplified a bit by merging it with the moderator permission to see DMails sent by the automod, so all staff can see both from and to automod. It makes sense when you consider that any automod DMail could be replied to by a user, exposing the original message to all staff.

All staff can create a ticket about a DMail sent to automod. View permissions could also be changed to allow all staff to see tickets about DMails sent to the automod, but I left this out to avoid making the ticket permissions more complicated than necessary.

### DIFF
--- a/app/controllers/admin/automod_dmails_controller.rb
+++ b/app/controllers/admin/automod_dmails_controller.rb
@@ -40,7 +40,7 @@ module Admin
     private
 
     def find_system_dmail(id)
-      @dmail = Dmail.where("owner_id = ?", User.system.id).includes(:to, :from).find(id)
+      Dmail.where("owner_id = ?", User.system.id).includes(:to, :from).find(id)
     end
   end
 end
--- a/app/controllers/admin/automod_dmails_controller.rb
+++ b/app/controllers/admin/automod_dmails_controller.rb
@@ -15,8 +15,32 @@ module Admin
 
     def show
       @user = User.system
-      @dmail = Dmail.where("owner_id = ?", @user.id).includes(:to, :from).find(params[:id])
+      @dmail = find_system_dmail(params[:id])
       respond_with(@dmail)
+    end
+
+    def mark_as_read
+      @dmail = find_system_dmail(params[:id])
+      @dmail.mark_as_read!
+      respond_to do |format|
+        format.html { redirect_to(admin_automod_dmail_path(@dmail), notice: "Message marked as read") }
+        format.json
+      end
+    end
+
+    def mark_as_unread
+      @dmail = find_system_dmail(params[:id])
+      @dmail.mark_as_unread!
+      respond_to do |format|
+        format.html { redirect_to(admin_automod_dmail_path(@dmail), notice: "Message marked as unread") }
+        format.json
+      end
+    end
+
+    private
+
+    def find_system_dmail(id)
+      @dmail = Dmail.where("owner_id = ?", User.system.id).includes(:to, :from).find(id)
     end
   end
 end
--- a/app/controllers/dmails_controller.rb
+++ b/app/controllers/dmails_controller.rb
@@ -57,15 +57,28 @@ class DmailsController < ApplicationController
     @dmail = Dmail.find(params[:id])
     check_privilege(@dmail)
     @dmail.mark_as_read!
+    if @dmail.to_id == User.system.id
+      respond_to do |format|
+        format.html { redirect_to(admin_automod_dmail_path(@dmail), notice: "Message marked as read") }
+        format.json
+      end
+    end
   end
 
   def mark_as_unread
     @dmail = Dmail.find(params[:id])
     check_privilege(@dmail)
     @dmail.mark_as_unread!
-    respond_to do |format|
-      format.html { redirect_to(dmails_path, notice: "Message marked as unread") }
-      format.json
+    if @dmail.to_id == User.system.id
+      respond_to do |format|
+        format.html { redirect_to(admin_automod_dmail_path(@dmail), notice: "Message marked as unread") }
+        format.json
+      end
+    else
+      respond_to do |format|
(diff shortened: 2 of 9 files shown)


jmortiger:
> DMail permissions now allow all staff to view when the recipient is the automod, which is consistent with the new automod DMails route.

I'll highlight that whether this inconsistency stem from the DMail permissions being too restrictive (as this assumes) or from the Automod DMail permissions being too permissive is a question that needs to be answered. If the answer isn't as expected, this third line should be changed so that it says from or to the system user, instead of adding another line that expands access to their received messages. That fixes the problem without expanding permissions in a way that is an open question.

```rb
   def visible_to?(user) 
     return true if user.is_janitor? && to_id == User.system.id 
      return true if user.is_moderator? && (from_id == User.system.id || Ticket.where(qtype: "dmail", disp_id: id).exists?)
```

app/controllers/dmails_controller.rb
```diff
@@ -57,15 +57,28 @@ def mark_as_read
     @dmail = Dmail.find(params[:id])
     check_privilege(@dmail)
     @dmail.mark_as_read!
+    if @dmail.to_id == User.system.id
```
Sindrake: Don't add exceptions and special cases to the main dmails controller for the sake of this.
Add another action to the automod DMails controller to do this properly.

app/models/dmail.rb
```diff
@@ -240,6 +240,7 @@ def update_recipient_unread_count
   end
 
   def visible_to?(user)
+    return true if user.is_janitor? && to_id == User.system.id
```
Sindrake, line 243: Why is this necessary?..

Sindrake (changes requested):

app/models/dmail.rb
DRLaDRLa: The dmail reply route has a visibility check, you can't reply to a dmail you can't view. Since all staff can view dmails sent to the automod through the automods dmails page, I figured the simplest solution to that problem was to also reflect this permission in the dmail visibility. Without this the reply button is a 403 for any user.
This is just what solution I could come up with, I don't know of it's necessarily the best solution.

jmortiger: As DRLa says, the `visible_to?` call gates all direct access to DMails received by automod outside of the automod dmail route, even for moderators (who only have access to message automod *sends*, not receives. To allow access in the forwarding & replying actions, this must be changed to allow access.

If janitors shouldn't see DMails sent to automod (which makes sense for responses to resolved tickets), then the automod access should be modified (by which I obviously mean the automod endpoint), & the solution from [my earlier comment](https://github.com/e621ng/e621ng/pull/1988#issuecomment-4421172095) should replace this; if janitors should be able to see DMails sent to automod (which makes sense for deletion DMails), then this solution should be preserved.

DRLaDRLa: To fully remove access to DMails sent to the automod, you would also need to change the user level for the `admin/automod_dmails` route because at the moment it's "all staff". I believe for e6ai we would prefer to keep it like that (because it's mostly replies to deletion dmails anyway), so making it configurable would be better, but please confirm with our lead admin.

Sindrake: No need to be deliberately dense, you two.
The original explanation of "you can't reply to a dmail you can't view" was fine, thank you.

app/controllers/dmails_controller.rb
DRLaDRLa: Done. I knew there must have been a better way to do this...

app/controllers/admin/automod_dmails_controller.rb
```diff
@@ -15,8 +15,32 @@ def index
...
+      respond_to do |format|
+        format.html { redirect_to(admin_automod_dmail_path(@dmail), notice: "Message marked as unread") }
+        format.json
+      end
+    end
+
+    private
+
+    def find_system_dmail(id)
+      @dmail = Dmail.where("owner_id = ?", User.system.id).includes(:to, :from).find(id)
```
Sindrake:
```suggestion
      Dmail.where(owner_id: User.system.id).includes(:to, :from).find(id)
```

No need to assign it twice.

app/views/admin/automod_dmails/show.html.erb
```diff
@@ -16,13 +16,17 @@
...
         <%= format_text(@dmail.body) %>
-
-        <% if @dmail.is_automated? %>
-          <p class="info">
-            This is an automated message. Responses will not be seen. If you have any questions either message a moderator or ask in the forum.
-          </p>
-        <% end %>
       </div>
+      <p>
+        <%= link_to "Respond", new_dmail_path(respond_to_id: @dmail) %>
```
Sindrake, line 21: Do you think you could add a spec to verify that a janitor can load a `new_dmail_path` when responding to a system DMail?

Sindrake (changes requested):

app/views/admin/automod_dmails/show.html.erb
DRLaDRLa: Sure, done.

DRLaDRLa:
BTW I wasn't sure whether the mark read/unread should just refresh the page or send back to the index, I went for the former as the "least surprising" option, but I'm open to change it (or just wait and see what people think after it goes live).

Sindrake (approved):
lgtm

Issue #1760, opened 2026-04-03
Clicking on a spoiler tag on mobile should not cause a click on a link inside it


jmortiger:
> Would it be appropriate to remove the spoiler-tags if a post-description has them around links? 
As someone who mostly uses the site on mobile, it's annoying to accidentally be opening links when just wanting to know what is said behind a spoiler.

https://discord.com/channels/431908090883997698/1392682818890498059/1419771953597907096

If this persists, a simple `preventDefault`/`stopPropagation`/`stopImmediatePropagation` could do the trick, but I haven't checked for certain.

PR #1989, opened 2026-05-09
[UI] Fix improper spoiler tag link navigation on mobile

maremom:
This completes https://github.com/e621ng/e621ng/issues/1760 by preventing link navigation on touch devices during the user's initial tap on a spoiler tag. It also cleans up and removes the old faulty CSS-only attempt.


Sindrake (approved):
lgtm, thank you!

PR #1993, opened 2026-05-09
[Takedowns] Allow approvers the same data access as moderators

mm12:
Allow approvers to view the same information that moderators can.

### DIFF
--- a/app/controllers/takedowns_controller.rb
+++ b/app/controllers/takedowns_controller.rb
@@ -97,7 +97,7 @@ class TakedownsController < ApplicationController
 
   def search_params
     permitted_params = %i[status]
-    permitted_params += %i[source reason creator_id creator_name creator_logged_in reason_hidden instructions post_id notes] if CurrentUser.is_moderator?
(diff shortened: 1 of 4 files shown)


Sindrake (changes requested):
There is another gate on line 109 of `app/views/takedowns/show.html.erb` that was not widened:
```ERB
        <% if !@takedown.reason_hidden || CurrentUser.is_moderator? %>
```

mm12:
> ```
> <% if !@takedown.reason_hidden ||
> ```

Must have removed it when debugging. 
But also, doesn't even seem to do it's job: 

```rb
  <% if !@takedown.reason_hidden || CurrentUser.is_moderator? %>
    <%= format_text(@takedown.notes) %>
    <% if @takedown.reason_hidden %><span class="text-error">(HIDDEN)</span><% end %>
  <% else %>
    <span class="text-error">[Admin notes hidden]</span>
  <% end %>
```

Automoderator view: 
image

After changing to add ` || CurrentUser.is_approver? `, it works fine. This bewilders me. 

EDIT: ah, its because ruby is weird.

Sindrake (approved):

PR #1996, opened 2026-05-10
[Post Search] Fixes bug where tag count in groups is counted twice

jmortiger:
https://discord.com/channels/431908090883997698/1501212717514494093/1503048699235008724

Latest version made on mobile, not tested locally.

### DIFF
--- a/app/logical/elastic_post_query_builder.rb
+++ b/app/logical/elastic_post_query_builder.rb
@@ -38,6 +38,7 @@ class ElasticPostQueryBuilder < ElasticQueryBuilder
     @always_show_deleted = always_show_deleted
     @always_show_deleted ||= !query.hide_deleted_posts?(at_any_level: true) if GLOBAL_DELETED_FILTER && @depth <= 0
     @error_on_depth_exceeded = kwargs.fetch(:error_on_depth_exceeded, ERROR_ON_DEPTH_EXCEEDED)
+    @process_groups = kwargs[:process_groups]
     super(query)
   end
 
@@ -67,7 +68,7 @@ class ElasticPostQueryBuilder < ElasticQueryBuilder
         x = TagQuery.new(
           x,
           resolve_aliases: @resolve_aliases,
-          free_tags_count: @free_tags_count + @q.tag_count,
+          free_tags_count: @free_tags_count + (@process_groups || TagQuery.will_count_group_tags? ? 0 : @q.tag_count),
           error_on_depth_exceeded: @error_on_depth_exceeded,
           depth: @depth + 1,
           hoisted_metatags: nil,
@@ -77,7 +78,7 @@ class ElasticPostQueryBuilder < ElasticQueryBuilder
       temp = ElasticPostQueryBuilder.new(
         x,
         resolve_aliases: @resolve_aliases,
-        free_tags_count: @free_tags_count + @q.tag_count,
+        free_tags_count: @free_tags_count + (@process_groups || TagQuery.will_count_group_tags? ? 0 : @q.tag_count),
         enable_safe_mode: @enable_safe_mode,
         always_show_deleted: GLOBAL_DELETED_FILTER ? true : asd_cache,
         error_on_depth_exceeded: @error_on_depth_exceeded,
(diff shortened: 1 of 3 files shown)


app/logical/elastic_post_query_builder.rb
```diff
@@ -67,7 +68,7 @@ def add_group_search_relation(groups)
         x = TagQuery.new(
           x,
           resolve_aliases: @resolve_aliases,
-          free_tags_count: @free_tags_count + @q.tag_count,
+          free_tags_count: @free_tags_count + (@process_groups || TagQuery.will_count_group_tags? ? 0 : @q.tag_count),
```
Sindrake:
```suggestion
          free_tags_count: @free_tags_count + ((@process_groups || TagQuery.will_count_group_tags?) ? 0 : @q.tag_count),
```
Feels a little confusing. Same thing on line 81.
Might even want to extract it into a local variable for clarity, and reuse that in both cases.

spec/logical/elastic_post_query_builder/initialization_spec.rb
```diff
@@ -32,10 +32,29 @@ def build_query(query_string, **opts)
       end.to raise_error(TagQuery::DepthExceededError)
     end
 
-    it "raises DepthExceededError when depth exceeds DEPTH_LIMIT" do
+    it "does not error when depth doesn't exceed DEPTH_LIMIT" do
       expect do
-        build_query("cute", depth: TagQuery::DEPTH_LIMIT + 1)
-      end.to raise_error(TagQuery::DepthExceededError)
+        build_query("cute", depth: TagQuery::DEPTH_LIMIT - 2)
+      end.not_to raise_error(TagQuery::DepthExceededError)
```
Sindrake:
```suggestion
      end.not_to raise_error
```
This is an antipattern that's been deprecated for a while now.
Using `.not_to raise_error` with a class means that the test passes if any *other* error is raised, which can hide regressions.

spec/logical/elastic_post_query_builder/initialization_spec.rb
```diff
@@ -32,10 +32,29 @@ def build_query(query_string, **opts)
...
+        expect do
+          build_query("#{search} -( caprine faun wooloo dubwool skiddo gogoat mareep flaaffy caprity deer faun_(spyro) antelope pronghorn camelid giraffid pinniped dewgong scolipede mega_scolipede )")
+        end.to raise_error(TagQuery::CountExceededError)
+      end
+
+      it "does not error when count doesn't exceed limit" do
+        allow(Danbooru.config.custom_configuration).to receive(:tag_query_limit).and_return(40)
+        expect do
+          build_query(search)
+        end.not_to raise_error(TagQuery::CountExceededError)
```
Sindrake:
```suggestion
        end.not_to raise_error
```
Same as above.

app/logical/elastic_post_query_builder.rb
```diff
@@ -38,6 +38,7 @@ def initialize( # rubocop:disable Metrics/ParameterLists
     @always_show_deleted = always_show_deleted
     @always_show_deleted ||= !query.hide_deleted_posts?(at_any_level: true) if GLOBAL_DELETED_FILTER && @depth <= 0
     @error_on_depth_exceeded = kwargs.fetch(:error_on_depth_exceeded, ERROR_ON_DEPTH_EXCEEDED)
+    @process_groups = kwargs[:process_groups]
```
Sindrake: I am confused by this variable's existence.
It does not look like `:process_groups` is ever set on the ElasticPostQueryBuilder.
It's also not being passed on into the recursion on lines 78-87, so it disappears anyways.

Sindrake (changes requested):

app/logical/elastic_post_query_builder.rb
jmortiger: It's an override to force TagQuery to process groups that's currently unused, but I'd rather unnecessarily account for it properly here then have this bug randomly rear its head again. I'll likely trim it with the rest of the configurable bloat from TagQuery, but until that time, I'd rather ensure it work properly if someone uses it for whatever reason.

app/logical/elastic_post_query_builder.rb
jmortiger: I tried to do that and RuboCop yelled at me.

app/logical/elastic_post_query_builder.rb
Sindrake: Alright, sounds good.

spec/logical/elastic_post_query_builder/initialization_spec.rb
Sindrake: I do realize that I introduced that pattern in the tests, yes.

app/logical/elastic_post_query_builder.rb
Sindrake: Ah, so it does. How silly.
I would just extract `@process_groups || TagQuery.will_count_group_tags?` into a variable or something.

It's a nitpick. Feel free to dismiss if you don't believe that it's worth addressing.

Sindrake (approved):

Issue #1753, opened 2026-04-03
Add more source sanitizers


jmortiger:
As part of the link sanitizer upgrades, abadbird has helpfully provided additional sites that could benefit from cleaning. 
Information below, pulled from [this thread](https://discord.com/channels/431908090883997698/1415540337786748928).

### Conversions

- [ ] [InkBunny old submissionview.php post pages](https://e621.net/posts?tags=source%3A*submissionview.php*) ~61K

```

https://inkbunny.net/submissionview.php?id=1173899&page=3#pictop
https://inkbunny.net/submissionview.php?id=1173899&page=3
https://inkbunny.net/s/1173899-p3

```

- [ ] [Pixiv old account style](https://e621.net/posts?tags=source%3A*pixiv*member.php*) ~11K

```

https://www.pixiv.net/member.php?id=923628
https://www.pixiv.net/users/923628

```

- [ ] [Pixiv /en/ account](https://e621.net/posts?tags=source%3A*pixiv*en*user*) ~7.1K

```

https://www.pixiv.net/en/users/107143906
https://www.pixiv.net/users/107143906

```

- [ ] [Pixiv /en/ dead links](https://e621.net/posts?tags=source%3A*-*pixiv*en*artwork*) ~150

```

-https://www.pixiv.net/en/artworks/25320628
-https://www.pixiv.net/artworks/25320628

```

- [ ] [Facebook mobile site](https://e621.net/posts?tags=source%3A*m.facebook*) ~600

```

https://m.facebook.com/photo.php?fbid=3058563794264742&id=896121230509020&set=a.2916025081851948&source=43&refid=13&__tn__=%2B%3D
https://www.facebook.com/photo.php?fbid=3058563794264742

```

- [ ] [Webtoons mobile site](https://e621.net/posts?tags=source%3A*https%3A%2F%2Fm.webtoons*) ~300

```

https://m.webtoons.com/en/canvas/forestdale/forestdale-314-missed-birthday/viewer?title_no=856922&episode_no=331
https://www.webtoons.com/en/canvas/forestdale/forestdale-314-missed-birthday/viewer?title_no=856922&episode_no=331

```

- [ ] [Tapas mobile site](https://e621.net/posts?tags=source%3A*m.tapas*) ~150

```

https://m.tapas.io/episode/189498
https://tapas.io/episode/189498

```

- [ ] [Imgur mobile site](https://e621.net/posts?tags=source%3A*m.imgur*) ~70

```

[...]

PR #2004, opened 2026-05-12
[Posts] Improve and add more source sanitizers

maremom:
Implements some of the items in https://github.com/e621ng/e621ng/issues/1753 and more by extending the existing source URL sanitizers as well as adding new ones. 

Completes (from said issue):

- Pixiv old account style (Convert deprecated member.php?id= to /users/ format)

- Pixiv /en/ account (Canonicalize /en/users/ to /users/)

- Facebook mobile site (Convert mobile subdomain to www.facebook.com)

- Webtoons mobile site (Convert mobile subdomain to www.webtoons.com)

- Tapas mobile site (Convert mobile subdomain to tapas.io)

- Imgur mobile site (Convert mobile subdomain to imgur.com)

- FurAffinity SFW subdomain (Convert sfw subdomain to www.furaffinity.net)

- Facebook fbid & (Strip unneeded parameters, keep only fbid identifier)

Additionally:
- Facebook: Canonicalize friendly-link /photo/ into only /photo.php.
- Facebook: Strip unneeded parameters from /story.php, keep only story_fbid and id.
- Facebook: Convert web.facebook.com into www.facebook.com.

Some Notes: Facebook sources take on additional forms but the number of posts make those negligible. It would be nice to canonicalize story.php into photo.php, but each has a unique fbid resolved only through the /photo/ endpoint.

### DIFF
--- a/app/logical/sources/alternates/furaffinity.rb
+++ b/app/logical/sources/alternates/furaffinity.rb
@@ -15,6 +15,9 @@ module Sources
         # Handle old CDN or old broken CDN
         if ["d.facdn.net", "d2.facdn.net"].include?(@parsed_url.host)
           @parsed_url.host = "d.furaffinity.net"
+        # Handle SFW subdoamain
+        elsif @parsed_url.host == "sfw.furaffinity.net"
+          @parsed_url.host = "www.furaffinity.net"
         end
         # Convert /full/ submission links to /view/ links
         if @parsed_url.path.start_with?("/full/")
--- a/app/logical/sources/alternates/furaffinity.rb
+++ b/app/logical/sources/alternates/furaffinity.rb
@@ -15,7 +15,7 @@ module Sources
         # Handle old CDN or old broken CDN
         if ["d.facdn.net", "d2.facdn.net"].include?(@parsed_url.host)
           @parsed_url.host = "d.furaffinity.net"
-        # Handle SFW subdoamain
+        # Handle SFW subdomain
         elsif @parsed_url.host == "sfw.furaffinity.net"
           @parsed_url.host = "www.furaffinity.net"
         end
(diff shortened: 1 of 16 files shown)


app/logical/sources/alternates/furaffinity.rb
```diff
@@ -15,6 +15,9 @@ def original_url
         # Handle old CDN or old broken CDN
         if ["d.facdn.net", "d2.facdn.net"].include?(@parsed_url.host)
           @parsed_url.host = "d.furaffinity.net"
+        # Handle SFW subdoamain
```
Sindrake:
```suggestion
        # Handle SFW subdomain
```

Nitpick.

app/logical/sources/alternates/imgur.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Sources
+  module Alternates
+    class Imgur < Base
+      def force_https?
+        true
```
Sindrake: This is correct.
However, the Base class that this alternate is extending also hardcodes `imgur.com` in the `force_https?` method. That should probably be removed.

spec/logical/sources/alternates_spec.rb
```diff
@@ -11,8 +11,8 @@
   # .all
   # -------------------------------------------------------------------------
   describe ".all" do
-    it "returns exactly 7 handler classes" do
-      expect(described_class.all.length).to eq(7)
+    it "returns exactly 11 handler classes" do
```
Sindrake: I feel like we should just drop this test.
It hard-codes the number of handler classes - while checking the exact same thing that the test right below it already does.

Sindrake (changes requested):
Couple minor nitpicks.
Otherwise, looks good!

Sindrake:
That's not what I meant at all.

You should remove `imgur.com` from here:
<https://github.com/e621ng/e621ng/blob/master/app/logical/sources/alternates/base.rb#L33>
And put this back into the Imgur alternate:
```Ruby
      def force_https?
        true
      end
```

Sindrake (approved):
lgtm

PR #2023, opened 2026-05-14
[Posts] Add deletion flag ID placeholder to deletion DMails

DRLaDRLa:
Add `%FLAG_ID%` so you can have a link to the appeal form in a deletion DMail.
Also minor improvement to the deletion DMail UI, to stop it overwriting your changes even when there's no "reason" (pun intended).

### DIFF
--- a/app/javascript/src/js/pages/moderator/post/posts/post_delete.js
+++ b/app/javascript/src/js/pages/moderator/post/posts/post_delete.js
@@ -38,12 +38,13 @@ PostDeletion.init = function () {
       // Don't overwrite the input's text unless necessary
       let titleChanged = force, messageChanged = force;
       if (!Utility.blank(newReason)) {
-        if (newTitle.indexOf("%REASON%") >= 0) {
-          newTitle = newTitle.replaceAll("%REASON%", newReason);
+        const reasonPlaceholder = "%REASON%";
+        if (newTitle.indexOf(reasonPlaceholder) >= 0) {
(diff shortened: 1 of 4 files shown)


app/javascript/src/js/pages/moderator/post/posts/post_delete.js
```diff
@@ -31,21 +31,33 @@ PostDeletion.init = function () {
...
+        dMailTitleInput.val(newTitle);
+      }
+      if (messageChanged) {
+        dMailTextArea.val(newMessage);
       }
-      dMailTitleInput.val(newTitle);
-      dMailTextArea.val(newMessage);
     };
     dMailCheckBox.addEventListener("click", () => updateDMailActivation());
     dMailTemplate.addEventListener("change", () => updateDMailReason());
```
Sindrake:
```suggestion
    dMailTemplate.addEventListener("change", () => updateDMailReason(true));
```
With `force = false`, selecting a different template when the reason field is empty does nothing - the title and message inputs are not updated to the new template's content.

Fix, or confirm that the behavior is intentional.

Sindrake (changes requested):

app/javascript/src/js/pages/moderator/post/posts/post_delete.js
DRLaDRLa: Mistake, it's fixed.

Sindrake (approved):

PR #2043, opened 2026-05-17
[UI] Update displayed times

jmortiger:
This both updates displayed times & adds the ability to switch the displayed times & the title attribute's timestamp.

# Updating
If you spend enough time for the displayed label to be inaccurate, this PR will have the label update accordingly, matching the formatting & time ranges of the server.

# Toggle
You can switch the displayed time & the hover-over title attribute's timestamp by clicking/tapping the time. To enable this for time elements inside a link, an element is placed outside the link to trigger the toggle. An option to disable adding this element can be made if desired.

Clickable time element:
image
Non-clickable time element:
image
Toggled time element:
image


app/javascript/src/js/components/Timestamp.ts
```diff
@@ -0,0 +1,382 @@
...
+    return temp.substring(0, temp.length - 1);
+  }
+
+  public static timeDeltaInWords (fromTime: Date, options: object) {
+    return this.findDeltaFormat(fromTime, new Date(), options);
+  }
+
+  private static findDeltas (fromTime: Date, toTime: Date) {
+    if (fromTime.valueOf() > toTime.valueOf()) [fromTime, toTime] = [toTime, fromTime];
+    const deltaMs = (toTime.valueOf() - fromTime.valueOf()) / 1000,
```
Sindrake: `deltaMs` is actually a value in seconds rather than milliseconds, which will cause confusion later.

app/javascript/src/js/components/Timestamp.ts
```diff
@@ -0,0 +1,382 @@
...
+      this.timerID = undefined;
+      return false;
+    }
+  }
+
+  private static findUpdateInterval (range: FormatSpecification, from: Date, to = new Date()) {
+    let interval = range.updateIntervalMs;
+    if (interval > this.MAX_UPDATE_INTERVAL_MS) return undefined;
+    interval = Math.max(this.MIN_UPDATE_INTERVAL_MS, interval);
+    return interval - Math.round(Timestamp.findDeltas(from, to).deltaMs);
```
Sindrake: `interval` is a value in milliseconds, `deltaMs` is a value in seconds.

app/javascript/src/js/components/Timestamp.ts
```diff
@@ -0,0 +1,382 @@
...
+    let to_year = toTime.getFullYear();
+    if (toTime.getMonth() < 3)
+      to_year -= 1;
+
+    let leap_years: number;
+    if (from_year > to_year)
+      leap_years = 0;
+    else {
+      const fyear = from_year - 1;
+      leap_years = ((to_year / 4) - (to_year / 100) + (to_year / 400)) - ((fyear / 4) - (fyear / 100) + (fyear / 400));
```
Sindrake: This is using float division without any rounding, which is then passed on into the display string.
That will result in counts like "about 2.0625 years ago".

Sindrake (changes requested):
Started reviewing it, but the whole thing is kinda questionable.

Toggle switch for the date format (relative / absolute): **good**, useful, let's do it.
Automatically updating the displayed value as time passes: adds a large amount of complexity for the sake of something that's not really even a problem.

I would suggest stripping this down to just the toggle.

app/javascript/src/js/interfaces/E621.ts
```diff
@@ -31,4 +32,44 @@ export default interface E621Type {
...
+  // compatibility aliases
+  error: (message: string) => void,
+  notice: (message: string, permanent?: boolean) => void,
+  Flash: {
+    error: (message: string) => void,
+    notice: (message: string, permanent?: boolean) => void,
+  },
+}
+
+export function makeE621Instance (): E621Type {
```
Sindrake: What's the point of this?

jmortiger: Currently, we have to update the interface & the instantiation separately (which I've failed to do in prior PRs), but there's no clear reason to do this; we can just put the instantiation in a function exported from the same place the interface is defined.

This drastically reduces both the chances of the 2 falling out of sync & the things to keep track of for these kinds of changes, & also removes the need to edit both files every time these additions are made, reducing the review burden of PRs.

jmortiger:
The reason I made this was that I personally have trouble with this all the time when replying to things on mobile, plus going back on Chrome for Android doesn't reload the page, so it's even further out of date. I know I'm particularly slow, but having to refresh the page to check the times is extremely frustrating.

I know it's added complexity, but I separated this into its own class specifically to keep that complexity isolated & self-contained. Almost everything is private, & it solely touches `<time>` elements that match the expected string format; it's not going to cross any streams.

Additionally, if we do eventually take advantage of Turbo frames, this will keep comment timestamps from being disjointed when one of them is edited.

jmortiger:
I'll take that as a rejection of my justification for the central feature & revise it to solely support the approved tertiary feature. I can't update this on mobile when closed, so I'll just keep it as a draft for now.

Sindrake:
Yes, that is a rejection of the self-updating feature.
I am aware that you've probably spend quite a lot of time on it, but it's still complexity and future tech debt for the sake of something that is not really a problem for the vast majority of people.

jmortiger:
I'd understand the concern more if it wasn't a completely self-contained class that nothing else should ever need to touch nor really maintain, & if this wasn't a standard feature that e621 wasn't the sole site I know of that doesn't do this, but I'll defer. I'll scuttle the whole class & just compress it into a function soon.

PR #2047, opened 2026-05-18
[ExceptionLog] Improve scrubbing of sensitive information

mm12:
Replaces IPv4, IPv6, and emails with `[PROTECTED]`-like scrubs

### DIFF
--- a/app/models/exception_log.rb
+++ b/app/models/exception_log.rb
@@ -99,4 +99,40 @@ class ExceptionLog < ApplicationRecord
 
     total
   end
+
+  def viewable_message
+    if CurrentUser.is_admin?
+      message
+    else
+      scrub_emails(scrub_ips(message))
+    end
+  end
+
+  def viewable_extra_params
+    if CurrentUser.is_admin?
+      extra_params
+    else
+      extra_params.deep_transform_values do |value|
+        # exclude user agent from scrub since it will likely contain version numbers:
+        if value == extra_params[:user_agent]
+          value
+        elsif value.is_a?(String)
+          scrub_emails(scrub_ips(value))
+        else # rubocop:disable Lint/DuplicateBranch
+          value
+        end
+      end
+    end
+  end
+
+  private
+
+  def scrub_ips(text)
+    text.gsub(/\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b/, "[IP PROTECTED]") # IPv4
+        .gsub(/\b(?:[A-F0-9]{1,4}:){7}[A-F0-9]{1,4}\b/i, "[IP PROTECTED]") # IPv6
+  end
+
+  def scrub_emails(text)
+    text.gsub(/\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b/i, "[EMAIL PROTECTED]")
+  end
 end
--- a/app/models/exception_log.rb
+++ b/app/models/exception_log.rb
@@ -143,7 +143,7 @@ class ExceptionLog < ApplicationRecord
 
     text.gsub(candidate_regex) do |match|
       token = Regexp.last_match(1)
-      candidate = token.sub(/\d+\z/, "").sub(/%.+\z/, "")
+      candidate = token.sub(%r{/\d+\z}, "").sub(/%.+\z/, "")
 
       # Skip the literal "::" separator to avoid false positives in log messages
       if candidate == "::"
--- a/app/models/exception_log.rb
+++ b/app/models/exception_log.rb
@@ -113,14 +113,9 @@ class ExceptionLog < ApplicationRecord
       extra_params
     else
       extra_params.deep_transform_values do |value|
-        # exclude user agent from scrub since it will likely contain version numbers:
-        if value == extra_params[:user_agent]
-          value
-        elsif value.is_a?(String)
-          scrub_emails(scrub_ips(value))
-        else # rubocop:disable Lint/DuplicateBranch
-          value
-        end
+        next value unless value.is_a?(String)
+        next value if value == extra_params["user_agent"]
+        scrub_emails(scrub_ips(value))
       end
     end
   end
--- a/app/models/exception_log.rb
+++ b/app/models/exception_log.rb
@@ -114,8 +114,8 @@ class ExceptionLog < ApplicationRecord
     else
       extra_params.deep_transform_values do |value|
         next value unless value.is_a?(String)
-        next value if value == extra_params["user_agent"]
-        scrub_emails(scrub_ips(value))
+        next scrub_emails(value) if value == extra_params["user_agent"]
+        scrub_ips(scrub_emails(value))
       end
     end
   end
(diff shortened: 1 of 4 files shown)


spec/models/exception_log/instance_methods_spec.rb
```diff
@@ -44,4 +44,52 @@ def make_log(overrides = {})
...
+    end
+
+    context "when not admin" do
+      include_context "as member"
+
+      it "preserves the user agent but scrubs emails and IPs in other fields" do
+        log = make_log(extra_params: raw_extra)
+        ve = log.viewable_extra_params
+
+        expect(ve["user_agent"]).to eq("Mozilla/5.0")
```
Sindrake: Please, add a test that ensures that the values within `user_agent` do not get stripped away.
For example: `MyAwesomeBot/1.2.3.4`

app/models/exception_log.rb
```diff
@@ -99,4 +99,40 @@ def self.prune!(older_than: 1.year, batch_size: 1_000)
...
+        end
+      end
+    end
+  end
+
+  private
+
+  def scrub_ips(text)
+    text.gsub(/\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b/, "[IP PROTECTED]") # IPv4
+        .gsub(/\b(?:[A-F0-9]{1,4}:){7}[A-F0-9]{1,4}\b/i, "[IP PROTECTED]") # IPv6
```
Sindrake: This is the reason why I haven't done this myself.
The regex here works for fully-expanded 8-group IPv6.
It does not for compressed ones: `2001:db8::1`.
I am sure there are other edge cases like this.

Sindrake: Also, `\b` word boundaries are somewhat unreliable around colons (non-word characters).
Anchoring with lookahead/behind might be safer.

app/views/admin/exceptions/index.html.erb
```diff
@@ -23,11 +23,11 @@
         <td><%= link_to_user(exception_log.user) %></td>
         <td><%= exception_log.code %></td>
         <td>
-          <%= exception_log.extra_params.dig("params", "controller") %>/<%= exception_log.extra_params.dig("params", "action") %>
+          <%= exception_log.viewable_extra_params.dig("params", "controller") %>/<%= exception_log.viewable_extra_params.dig("params", "action") %>
```
Sindrake: Is this necessary?
Controller and action are not sensitive as far as I know.

app/models/exception_log.rb
```diff
@@ -99,4 +99,40 @@ def self.prune!(older_than: 1.year, batch_size: 1_000)
...
+    else
+      extra_params.deep_transform_values do |value|
+        # exclude user agent from scrub since it will likely contain version numbers:
+        if value == extra_params[:user_agent]
+          value
+        elsif value.is_a?(String)
+          scrub_emails(scrub_ips(value))
+        else # rubocop:disable Lint/DuplicateBranch
+          value
+        end
```
Sindrake:
```suggestion
        extra_params.deep_transform_values do |value|
          next value unless value.is_a?(String)
          next value if value == extra_params["user_agent"]
          scrub_emails(scrub_ips(value))
        end
```

The `extra_params` value is JSON, which only has string keys.
So, `extra_params[:user_agent]` is just going to be nil.

The suggested fix also gets rid of the rubocop skip.
Untested.

Sindrake (changes requested):
Looks okay.
One bug, a couple of nitpicks.

app/models/exception_log.rb
mm12: Yeah I was thinking of something like that but had some issues, I will take another look at it when I get a chance.

app/views/admin/exceptions/index.html.erb
mm12: Will rollback then

spec/models/exception_log/instance_methods_spec.rb
mm12: Ah, I fixed it in actual test but didn't write the unit. My code coverage file went all weird again too, thats annoying.

app/models/exception_log.rb
mm12: I knew about this shortcoming, yes. Does our code every even produce compressed IPs? I couldn't get it to. I'll mess around with it more later.

mm12: Tempted to just `try` making a new `IPAddr` from them and if it succeeds, remove it.

Sindrake: I can't say for sure whether they've been in the exception log.
But I've definitely seen them on user activity. So I'm going to say yes.

app/models/exception_log.rb
```diff
@@ -99,4 +99,67 @@ def self.prune!(older_than: 1.year, batch_size: 1_000)
...
+    end
+  end
+
+  private
+
+  def scrub_ips(text)
+    return text unless text.is_a?(String)
+
+    # First, explicitly replace dotted IPv4 addresses (with optional port/brackets/CIDR)
+    ipv4_regex = %r{/(?:\b|\[)((?:\d{1,3}\.){3}\d{1,3})(?:\]|\b)(?::\d{1,5})?/}
```
Sindrake: With `%r{...}` delimiters, this regex would only match IPs that are wrapped in slashes.
For example: `/1.2.3.4/`, but not `1.2.3.4`.

app/models/exception_log.rb
```diff
@@ -99,4 +99,67 @@ def self.prune!(older_than: 1.year, batch_size: 1_000)
...
+        match
+      end
+    end
+
+    # Then handle IPv6 and other candidate tokens (bracketed or bare), stripping CIDR/zone ids
+    candidate_regex = %r{(?:\b|\[)([0-9A-Fa-f:.%/]+)(?:\]|\b)(?::\d{1,5})?}
+
+    text.gsub(candidate_regex) do |match|
+      token = Regexp.last_match(1)
+      candidate = token.sub(%r{/\d+\z}, "").sub(%r{/%.+\z/}, "")
```
Sindrake: Same issue as above in the second `.sub`.
The slashes at the start and end of the regex are interpreted as literal symbols.

app/models/exception_log.rb
```diff
@@ -99,4 +99,67 @@ def self.prune!(older_than: 1.year, batch_size: 1_000)
...
+  end
+
+  def viewable_extra_params
+    if CurrentUser.is_admin?
+      extra_params
+    else
+      extra_params.deep_transform_values do |value|
+        next value unless value.is_a?(String)
+        next scrub_emails(value) if value == extra_params["user_agent"]
+        scrub_ips(scrub_emails(value))
```
Sindrake, line 118: The `viewable_message` scrubs IPs first, then emails.
The `viewable_extra_params` does it the other way around.

Not a dealbreaker, but consistency would be nice.

Sindrake (changes requested):

Sindrake (approved):

PR #2050, opened 2026-05-18
[Posts] Replace pool string with pool and set ID arrays

DoubleThePsycho:
# Replace `pool_string` with `pool_ids` and `set_ids` arrays

## Summary

This replaces `posts.pool_string` with separate structured membership columns:

- `posts.pool_ids integer[]`
- `posts.set_ids bigint[]`

The goal is to remove the mixed token format currently used by `pool_string`, where pool and post set memberships are stored together as text tokens such as:

```text
pool:1 pool:2 set:1 set:5
```

After this change, pool and set memberships are stored separately while preserving the existing denormalized “post knows what groups it belongs to” lookup behavior.

## Motivation

`pool_string` currently stores multiple concepts in one text field and requires token parsing for pool/set membership logic. Splitting this into structured arrays makes the data model clearer and avoids mixing public pool membership with internal post set membership.

This follows the earlier discussion about replacing the mixed `pool_string` format with structured columns while keeping the current fast lookup/cache behavior.

## Changes

- Adds `posts.pool_ids` as `integer[]`
- Adds `posts.set_ids` as `bigint[]`
- Backfills both columns from existing `pool_string` tokens and authoritative pool/post set membership arrays
- Updates pool and post set membership methods to use the new arrays
- Updates post set cleanup/sync jobs to mutate `set_ids` / `pool_ids`
- Updates `pool:any`, `pool:none`, and `inpool:*` queries to check `pool_ids`
- Removes `posts.pool_string` and its token GIN index
- Keeps `set_ids` hidden from API output
- Preserves public post JSON `pools` output

## Notes

Previously, `pool:any`, `pool:none`, and `inpool:*` checked whether `pool_string` was empty, which could include post set tokens. These checks now use `pool_ids`, so they reflect only actual pool membership.

A full OpenSearch reindex should not be required because the public/indexed pool membership values are preserved; this changes the backing storage, not the semantic value of `post.pool_ids`.

Includes a migration that adds `posts.pool_ids` and `posts.set_ids`, backfills them from existing membership data, removes the old `pool_string` token index, and drops `posts.pool_string`.

The main risk is migration cost on production-sized data, since this backfills post membership arrays across existing posts.

## Validation

- Ran focused specs for pool/set methods, post set jobs, pool query behavior, pool creation, and API hidden attributes
- Ran RuboCop on touched files
- Verified migrated schema has `pool_ids` / `set_ids` and no `pool_string`
- Verified locally that post JSON still exposes `pools` and does not expose `set_ids` or `pool_string`

### DIFF
--- a/app/jobs/post_set_cleanup_job.rb
+++ b/app/jobs/post_set_cleanup_job.rb
@@ -4,23 +4,23 @@ class PostSetCleanupJob < ApplicationJob
   queue_as :default
   sidekiq_options lock: :until_executing
 
-  # General cleanup for post membership tokens in pool_string.
+  # General cleanup for post membership arrays.
   def perform(type, obj_id)
     type = type.to_sym
 
     case type
     when :set
-      token_prefix = "set:"
+      column = :set_ids
       removal_method = :remove_set!
     when :pool
-      token_prefix = "pool:"
+      column = :pool_ids
       removal_method = :remove_pool!
     else
       raise ArgumentError, "Invalid type: #{type.inspect}"
     end
 
-    tag = "#{token_prefix}#{obj_id}"
-    scope = Post.where("string_to_array(pool_string, ' ') @> ARRAY[?]::text[]", tag)
+    array_type = type == :set ? "bigint" : "integer"
+    scope = Post.where("#{column} @> ARRAY[?]::#{array_type}[]", obj_id)
 
     # Pools check for whether the user account is older than 7 days.
     CurrentUser.as_system do
--- a/app/jobs/post_set_posts_sync_job.rb
+++ b/app/jobs/post_set_posts_sync_job.rb
@@ -9,15 +9,13 @@ class PostSetPostsSyncJob < ApplicationJob
   end
 
   def perform(set_id)
-    set   = PostSet.find(set_id)
-    token = "set:#{set_id}"
+    set = PostSet.find(set_id)
 
     current_ids = set.post_ids.to_set
-    token_ids   = Post.where("string_to_array(pool_string, ' ') @> ARRAY[?]::text[]", token)
-                      .pluck(:id).to_set
+    synced_ids  = Post.where("set_ids @> ARRAY[?]::bigint[]", set_id).pluck(:id).to_set
 
-    to_add    = (current_ids - token_ids).to_a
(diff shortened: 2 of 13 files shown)


DoubleThePsycho:

Sindrake:
Thank you.

Unfortunately, in its current state, this PR is not viable.
The underlying logic is reasonable, but the lack of a well thought-out migration strategy makes this a complete non-starter.

Here are some stats from production:
6.4 million posts
53k pools, 11 average posts per pool
69k post sets, 338 average posts per set

Backfilling pool data is somewhat reasonable, although it will still take some time.
Backfilling post set data and then restoring the `pool_string` will affect millions of rows, and will take 30-45 minutes each to complete. In total, this is up to 2 hours of downtime, which is unacceptable.

In addition to that, we run multiple app instances behind a load balancer in production.
This approach means that we would not be able to do a seamless rollout, and would instead be forced to take all of them down while the migration runs.

DoubleThePsycho:
Thanks, that makes sense.

I focused too much on getting the final data shape working locally and not enough on making the rollout safe for production. The multi-instance deployment issue especially makes sense; dropping `pool_string` in the same step as switching reads would make old and new app versions incompatible during rollout.

I’ll rework this as a phased expand/backfill/contract approach across multiple PRs:

1. Add nullable `pool_ids` / `set_ids` while keeping `pool_string`
2. Make the app compatible with both formats and dual-write where needed
3. Add a later batched backfill step/job that can migrate existing posts incrementally while the app stays online
4. Switch reads/queries to the new columns only after the backfill is complete and verified
5. Drop `pool_string` and the old token index in a final cleanup PR

Would you prefer I close this PR and open a smaller schema-only first step, or convert this PR into the first phased PR?

DoubleThePsycho:
I opened a smaller phased first step here: [[Posts] Add pool and set ID arrays alongside pool_string](https://github.com/e621ng/e621ng/pull/2078)

This one only adds nullable `pool_ids` / `set_ids` alongside `pool_string` and begins dual-writing membership changes. It does not backfill existing rows, switch reads/search to the new columns, drop `pool_string`, or remove the old token index.

I’ll close this original all-in-one PR to avoid confusion.

PR #2059, opened 2026-05-19
[Posts] Improve favorite performance and add votes property to Post lists

clragon:
Removes `fav_string` from the Posts query, since it is potentially large and bad for performance, as well as complicated to maintain. Instead replaces it with querying the Favorites table directly. 

`fav_string` is not dropped from the table, and is currently still being written when a user favorites a post. This is a safety measure to ensure we can rollback with zero migration. A future PR can remove `fav_string` completely.

Additionally, we also query for post votes. `vote` is now available on the post json as an integer of `[-1, 0, 1]`. 

A follow up of this could be to allow users to blacklist their own voted posts. This could potentially be very convenient for users who do not wish to see posts they have downvoted.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -1254,6 +1254,8 @@ class Post < ApplicationRecord
 
     def reload(*)
       @fav_string = nil
+      @favorited_status_cache = nil
+      @vote_by_cache = nil
       super
     end
 
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -5,6 +5,18 @@ class Post < ApplicationRecord
   class DeletionError < Exception ; end
   class TimeoutError < Exception ; end
 
+  # Exclude fav_string from default SELECTs to avoid loading the (potentially huge)
+  # blob on every post fetch. Write paths transparently fall through to Post#fav_string,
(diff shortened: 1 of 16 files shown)


app/models/post.rb
```diff
@@ -1210,58 +1216,97 @@ def avoid_posting_tags
...
+    # Lazy reader: the column is ignored by AR (see ignored_columns at the top of
+    # this class), so on first access we issue a focused 1-column query and cache
+    # the result on the instance.
+    def fav_string
+      return @fav_string unless @fav_string.nil?
+      @fav_string = new_record? ? "" : (self.class.unscoped.where(id: id).pick(:fav_string) || "")
+    end
+
+    def reload(*)
+      @fav_string = nil
```
Sindrake, line 1256:
```suggestion
      @fav_string = nil
      @favorited_status_cache = nil
      @vote_by_cache = nil
```
Unless I am missing something, the other caches should be cleared too.

Sindrake (changes requested):
One potential issue.
Otherwise, looks good to me.

Sindrake (approved):

PR #2060, opened 2026-05-19
[Comments] Add vote to comment payloads

clragon:
Adds a `vote` integer to the comment json, which the query already paid for.
To call this field "vote", we also introduce a real blueprint for comments.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -249,8 +249,38 @@ class Comment < ApplicationRecord
     end
   end
 
+  module VoteMethods
+    extend ActiveSupport::Concern
+
+    module ClassMethods
+      # Bulk-populate the user's vote score for a collection of comments so that
+      # subsequent `vote_by(user_id)` calls return without hitting the DB.
+      def preload_vote_by!(comments, user_id)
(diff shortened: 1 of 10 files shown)


app/models/comment.rb
```diff
@@ -305,7 +335,7 @@ def update_last_commented_at_on_destroy
   end
 
   def method_attributes
-    super + %i[creator_name updater_name]
+    super + %i[creator_name updater_name vote_by]
```
Sindrake:
```suggestion
    super + %i[creator_name updater_name]
```

I would remove `vote_by` from here, unless there is a specific reason for it.
All JSON paths go through the CommentBlueprint, which already serializes the field as `vote` (not `vote_by`). Besides the discrepancy with naming, it adds an extra DB hit if `as_json` is ever called directly.

Sindrake (changes requested):
Could you add a test to make sure that the `index_by_comment` includes the `vote` field?
The `show` endpoint is tested, but not the index.

PR looks good overall.

Sindrake (approved):

PR #2070, opened 2026-05-21
[Posts] Fix exception in post editor for anonymous users

DRLaDRLa:
The editor form fields don't exist for anonymous users, which causes an exception.

`$("input[name='post[old_source]']").val()` returns `undefined`, which causes an exception in `splitUrls()`.


Sindrake (changes requested):
Okay, sure.
But why are we even initializing this whole thing if the user is anonymous?
Lines 595 and 596 – could just skip that for logged out users.

DRLaDRLa:
Just a case of me not being entirely familiar with the code base. Thanks for making the change.

Sindrake (approved):

PR #2071, opened 2026-05-21
[PostFlags] Fix bugs regarding resolved/unresolved and deletion flags

DRLaDRLa:
Basically `deletion_flag` and `pending_flag` would return irrelevant flags in some cases, so fix those and add tests.

These fixes broke the deleted post index, so I had a look at that. Turns out the deleted post index repeats posts for every deletion flag, resolved or not, so fix the query so it only shows a post once for the most recent (unresolved) deletion. This also makes it so undeleted posts no longer show up in the deleted index. I assume that's OK?

### DIFF
--- a/app/controllers/deleted_posts_controller.rb
+++ b/app/controllers/deleted_posts_controller.rb
@@ -8,7 +8,7 @@ class DeletedPostsController < ApplicationController
       @user = User.find(params[:user_id])
       @posts = Post.where(is_deleted: true)
       @posts = @posts.where(uploader_id: @user.id)
-      @posts = @posts.includes(:uploader).includes(:flags).where.not(post_flags: { id: nil }).order(Arel.sql("post_flags.created_at DESC")).paginate(params[:page])
+      @posts = @posts.includes(:uploader).includes(:flags).where.not(post_flags: { id: nil }).order(post_flags: { created_at: :desc }).paginate(params[:page])
     else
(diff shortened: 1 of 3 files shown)


app/controllers/deleted_posts_controller.rb
```diff
@@ -8,9 +8,9 @@ def index
       @user = User.find(params[:user_id])
       @posts = Post.where(is_deleted: true)
       @posts = @posts.where(uploader_id: @user.id)
-      @posts = @posts.includes(:uploader).includes(:flags).where("post_flags.id IS NOT NULL").order(Arel.sql("post_flags.created_at DESC")).paginate(params[:page])
+      @posts = @posts.includes(:uploader).includes(:flags).where.not(post_flags: { id: nil }).order(Arel.sql("post_flags.created_at DESC")).paginate(params[:page])
```
Sindrake:
```suggestion
      @posts = @posts.includes(:uploader).includes(:flags).where.not(post_flags: { id: nil }).order(post_flags: { created_at: :desc }).paginate(params[:page])
```

Is there a reason not to do this? ☝️ 
Not using raw SQL strings is generally preferable, even if nothing is interpolated into them.

Sindrake (changes requested):
Minor nitpick, but otherwise looks good.

app/controllers/deleted_posts_controller.rb
DRLaDRLa: I just missed that.

Sindrake (approved):

PR #2072, opened 2026-05-21
[Appeals] Nicer error handling when a flag can't be appealed

DRLaDRLa:
Instead of a "Access Denied" page, give an alert and return the user to the appeal index.

image

This is most likely to happen when clicking an appeal link in a deletion DMail, when the deletion has already been resolved.

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -23,6 +23,11 @@ class AppealsController < ApplicationController
 
   def new
     @appeal = Appeal.new(qtype: params[:qtype], disp_id: params[:disp_id])
+    unless @appeal.can_create_for?(CurrentUser.user)
+      redirect_to appeals_path, alert: "This deletion can't be appealed or has already been resolved."
+      return
+    end
+
     @existing_similar = Appeal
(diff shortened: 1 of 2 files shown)


Sindrake (dismissed):
Looks good.

DRLaDRLa:
Changed the wording a bit for #2081, not sure how that broke the labels check?

Sindrake:
Why would you ever do this?..
This means that the message will be incorrect until the other PR is approved.

Not to mention the fact that the force-push means that I cannot see what actually changed.

Sindrake (changes requested):
Message is incorrect in its current state.

DRLaDRLa:
> Why would you ever do this?.. This means that the message will be incorrect until the other PR is approved.
> 
> Not to mention the fact that the force-push means that I cannot see what actually changed.

So I wouldn't forget to change the message. I'll admit that was a bit dumb, I reverted the change (5e7a369410b7e24acd91ccbe8ed1379facb93f29 is the one you reviewed).

You can check the difference in a force push by using the compare button, Github does keep track of previous branch versions.

image

Sindrake (approved):

PR #2078, opened 2026-05-21
[Posts] Add pool ID array alongside pool_string

DoubleThePsycho:
## Summary

Adds a nullable `posts.pool_ids` array column alongside the existing `posts.pool_string` field.
Phases out all usage of post set IDs stored in the `pool_string`.

This is the first step in a phased migration away from the mixed `pool_string` token format.
This PR keeps the existing `pool_string` as a fallback while the data is being backfilled.

A follow-up PR will remove `pool_string` and its call sites, after backfill and verification.

## Motivation

`pool_string` currently stores both pool membership and post set membership as mixed text tokens, for example:

```text
pool:1 pool:2 set:1 set:5
```

The pool data is necessary.
Users can blacklist entire pools, so we need a way to quickly lookup pool ownership for hundreds of posts at a time.
Set data, on the other hand, is not needed. Sets are not blacklistable and do not require bulk lookups anywhere.

## Changes

- Adds nullable `posts.pool_ids integer[]`
- Adds nullable `posts.set_ids bigint[]`
- Adds GIN indexes for the new array columns
- Keeps `pool_string` intact
- Keeps reads/search/API behavior based on `pool_string`
- Begins dual-writing pool/set membership updates to the new columns where safe
- Keeps raw `pool_ids` / `set_ids` hidden from API output

## Rollout

Deploy and migrate as normal.
After migration, run the `db/fixes/140_backfill_post_pool_ids.rb` script to backfill pool IDs for existing posts.


DoubleThePsycho:
Small follow-up on this.

I have the next phased step implemented locally as a stacked branch, but I’m holding off on opening it until this first schema/dual-write step has been reviewed.

The follow-up step is an online batched backfill from `pool_string` into `pool_ids` / `set_ids`, without switching reads/search or removing `pool_string`.

I’d appreciate a quick sanity check on whether this first PR is split/scoped the way you’d want before I continue building further in this direction.

PR #2081, opened 2026-05-23
[Appeals] A user can only appeal a flag once

DRLaDRLa:
Should merge #2072 first, that'll change the test expected behaviour so this PR will need an update.

### DIFF
--- a/app/models/post_flag.rb
+++ b/app/models/post_flag.rb
@@ -45,6 +45,7 @@ class PostFlag < ApplicationRecord
     def can_appeal?(user = CurrentUser.user)
       return false unless is_deletion?
       return false if is_resolved?
+      return false if has_user_appealed?(user) # A user can only appeal a flag once
       return true if post.linked_users.include?(user.id) # Verified artists can appeal deletions of their own posts
       return false if reason =~ /takedown #\d+/i
       return true if post.uploader_id == user.id # Uploaders can appeal anything except for takedowns
@@ -227,6 +228,15 @@ class PostFlag < ApplicationRecord
     rescue
       nil
     end
+  end
+
+  def has_user_appealed?(user)
+    @has_user_appealed ||= {}
+    @has_user_appealed[user.id] ||= Appeal.where(
+      creator_id: user.id,
+      qtype: "flag",
+      disp_id: id,
+    ).exists?
   end
 
   # Creates an appropriate `PostEvent` unless this is a deletion.
--- a/app/models/post_flag.rb
+++ b/app/models/post_flag.rb
@@ -43,10 +43,8 @@ class PostFlag < ApplicationRecord
 
   module AccessMethods
     def can_appeal?(user = CurrentUser.user)
-      return false unless is_deletion?
-      return false if is_resolved?
-      # Uploaders can appeal except for takedowns, verified artists can appeal deletions of their own posts
-      return false unless (post.uploader_id == user.id && reason !~ /takedown #\d+/i) || post.linked_users.include?(user.id)
+      return false if is_resolved
+      return false unless appealable_by?(user)
       return false if has_user_appealed?(user)
       true
     end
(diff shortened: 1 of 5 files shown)


app/models/post_flag.rb
```diff
@@ -45,6 +45,7 @@ module AccessMethods
     def can_appeal?(user = CurrentUser.user)
       return false unless is_deletion?
       return false if is_resolved?
+      return false if has_user_appealed?(user) # A user can only appeal a flag once
```
Sindrake: This is incorrect.
Placing the check before verifying whether the user even has the ability to create an appeal (aka is uploader / verified artist) means that every single user who visits the deleted post's page also incurs the `exists?` query for no reason.

app/models/post_flag.rb
```diff
@@ -229,6 +230,15 @@ def parent_post
     end
   end
 
+  def has_user_appealed?(user)
+    @has_user_appealed ||= {}
+    @has_user_appealed[user.id] ||= Appeal.where(
```
Sindrake: Memoization does not work like this.
`||=` skips assignment when the right-hand side is falsy. Since `exists?` returns `false` when no appeal exists, the cache miss case is never stored.

At the very least, this should be something like this:
```Ruby
@has_user_appealed[user.id] = Appeal.where(...).exists? unless @has_user_appealed.key?(user.id)
@has_user_appealed[user.id]
```

I'm also not sure what the point of storing this value in a hash is.
Surely, we would only be checking this against the CurrentUser.

Sindrake (changes requested):
Not sure if I like how this is structured.
Besides the other issues I noted, I can foresee users being confused about why they can't appeal specific deletions. Replacing the link to create a new appeal with a link to an existing appeal in the flag notice may be worth exploring.

app/models/post_flag.rb
DRLaDRLa: I don't know what the lifetime of that cache is, but since an appeal can be created after this check, this value can become `true` later, so that's why I purposefully didn't cache `false`. But if it's `true` it will remain `true`. Is that correct or am I just misunderstanding how Rails works?

DRLaDRLa:
> Not sure if I like how this is structured. Besides the other issues I noted, I can foresee users being confused about why they can't appeal specific deletions. Replacing the link to create a new appeal with a link to an existing appeal in the flag notice may be worth exploring.

I wasn't sure either, I just went for the simplest option first, but yeah I do agree this would be better.

app/models/post_flag.rb
Sindrake: This kind of instance variable memoization is tied to the object's lifecycle.
As soon as the user makes a new request, the old object gets garbage-collected, and the memoized value goes with it. On the next page load, the query runs again regardless.

You could make it last longer with actual caching, but there really isn't a need for it.

DRLaDRLa:
"Appeal Deletion" buttons are replaced with "View Appeal":
image

Redirects to the existing appeal for `/appeals/new` (for example when clicking an appeal link in a deletion DMail):
image

app/views/posts/partials/show/content/notices/_flag_reasons.html.erb
```diff
@@ -37,6 +37,10 @@
         <div class="post-flag-appeal">
           <%= link_to "Appeal Deletion", new_appeal_path(qtype: "flag", disp_id: flag.id), class: "st-button" %>
         </div>
+      <% elsif (existing_appeal = flag.user_appeal(CurrentUser.user)).present? %>
```
Sindrake, line 40: This is causing an extra DB query per visible flag on the page for users who could have never appealed the flag in the first place.

At the very least, check if the flag is a deletion – it's pointless to even do an appeal lookup if it's not.

Sindrake (changes requested):

app/views/posts/partials/show/content/notices/_flag_reasons.html.erb
DRLaDRLa: I added the check, but the other option is to keep the "Appeal Deletion" button as-is, since it will redirect to the existing appeal. Maybe a bit weird, but does the same job, without querying anything until the button is clicked.

Side note: the check could be changed into a "can_view_existing_appeal?" that also allows staff to see those "View Appeal" links on the posts. Might be useful?

Sindrake (approved):

PR #2082, opened 2026-05-23
[PostVersions] Improve readability

clragon:
Post versions are very hard to read, and very hard to moderate. 

To make this easier, and surface important moderation information (such as, does this meet our tagging standards), we add category colors.
Since tags are now easier to distinguish, we retain them in their alphabetical order, to help users anchor their view. 
Categories are loaded once per page. The increase in page loading time is minimal. 
We also update the look of URLs to match what is displayed on the post detail page.

image

### DIFF
--- a/app/helpers/post_versions_helper.rb
+++ b/app/helpers/post_versions_helper.rb
@@ -18,7 +18,7 @@ module PostVersionsHelper
       changes << tag.div(tag.del(diff_source_link("-", source)))
     end
 
-    tag.span(safe_join(changes, " "), class: "diff-list")
+    tag.div(safe_join(changes, " "), class: "diff-list")
   end
 
   def diff_source_link(sign, source)
--- a/app/helpers/post_versions_helper.rb
+++ b/app/helpers/post_versions_helper.rb
@@ -3,16 +3,19 @@
 module PostVersionsHelper
   def post_source_diff(post_version)
     diff = post_version.diff_sources(post_version.previous)
-    changes = []
+    new_sources = post_version.source.to_s.split("\n")
+    added = diff[:added_sources].to_set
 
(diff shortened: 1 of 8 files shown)


app/javascript/src/styles/common/diffs.scss
```diff
@@ -25,27 +25,49 @@
 }
 
 .diff-list {
```
Sindrake, line 27: The diff list does not look great on version 1.
Consider adding some spacing between the lines. Example:
```CSS
    display: flex;
    flex-wrap: wrap;
    gap: 0.125rem 0.5rem;
```

Image

app/helpers/post_versions_helper.rb
```diff
@@ -3,64 +3,84 @@
...
+    new_sources = post_version.source.to_s.split("\n")
+    added = diff[:added_sources].to_set
 
-    diff[:added_sources].each do |source|
-      changes << tag.div(tag.ins(wordbreak_source("+#{source}")))
+    changes = new_sources.map do |source|
+      if added.include?(source)
+        tag.div(tag.ins(diff_source_link("+", source)))
+      else
+        tag.div(post_source_tag(source))
```
Sindrake, line 13: Negated links can overflow the page:

Image

Also, ideally the favicon would be on the same line as the decorated link rather than immediately linebreaking if the link is too long. But that's a nitpick.

Sindrake (changes requested):
Technically sound, some minor visual issues.

app/helpers/post_versions_helper.rb
clragon: Turns out its virtually impossible to make the favicon stay on the same line.

Sindrake (approved):

PR #2083, opened 2026-05-23
[Appeals] Enable blacklist on appeals page

mm12:
Add blacklisting to the appeals thumbnails 
image

### DIFF
--- a/app/views/appeals/types/_flag.html.erb
+++ b/app/views/appeals/types/_flag.html.erb
@@ -10,8 +10,9 @@
   </div>
 
   <div class="appeal-display-section appeal-display-flag-post">
-    <%= render(PostThumbnailComponent.new(post: @flag.post, stats: true, show_deleted: true, no_blacklist: true)) %>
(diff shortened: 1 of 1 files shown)


app/views/appeals/types/_flag.html.erb
```diff
@@ -10,8 +10,9 @@
   </div>
 
   <div class="appeal-display-section appeal-display-flag-post">
-    <%= render(PostThumbnailComponent.new(post: @flag.post, stats: true, show_deleted: true, no_blacklist: true)) %>
+    <%= render(PostThumbnailComponent.new(post: @flag.post, stats: true, show_deleted: true, no_blacklist: false)) %>
```
Sindrake:
```suggestion
    <%= render(PostThumbnailComponent.new(post: @flag.post, stats: true, show_deleted: true)) %>
```

I must have had a reason to not want the blacklist to apply here, but I can't remember why now.
Either way, `no_blacklist: false` is the default, and could probably just be dropped.

Otherwise, looks good.

Sindrake (changes requested):

Sindrake (approved):

PR #2084, opened 2026-05-24
[OAuth] Add OpenID Connect provider

clragon:
Adds Sign In With E621.

We use doorkeeper and doorkeeper-openid_connect for most of the logic.
We support both public and confidential clients.
There are no dynamically registered clients.
Client secrets (for confidential clients) are stored in plain text, just like API keys.
Tokens live for 15 minutes. Refresh tokens live forever. Giving refresh tokens a lifetime would have been more complex. There is no token family age (also complex) meaning once a user is signed in, they can stay signed in forever.

Only staff can create applications at the moment.
The amount of oauth applications uses the same config as the amount of API keys. I increased the amount slightly.
Applications can ask for identity only, or full API access scopes.
OAuth cannot access the oauth api or api key api.
User level and permission flags are stored in the tokens.
Applications can define a minimum member level.
If a user is demoted, their token fails immediately.
When a user is banned, all users are logged out of their applications, and can no longer authenticate with their applications. The applications are not deleted.

The design for the consent screen is similar to doppelgangers, which is similar to discord.
Applications have a description and home page field.
Application icons have been left out for now.
Oauth applications and links have been added to the new Apps settings tab which now also contains API keys. The screen was getting crowded.

| | |
|:---:|:---:|
| consent screen<br>**Consent screen** | error screen<br>**Error screen**
| application management<br>**Application management** | oauth applications index<br>**OAuth Applications** |
| settings apps tab<br>**Apps tab in Settings** | connected applications<br>**Connected applications** |


config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,40 @@
...
+  enable_application_owner confirmation: true
+
+  force_ssl_in_redirect_uri Rails.env.production?
+
+  skip_authorization do |resource_owner, client|
+    requested = client.scopes.to_a
+    Doorkeeper::AccessToken
+      .where(application_id: client.application.id, resource_owner_id: resource_owner.id, revoked_at: nil)
+      .any? { |token| (requested - token.scopes.to_a).empty? }
+  end
```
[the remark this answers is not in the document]
clragon: Yes, but the sql query would be really annoying, and the amount of tokens in the table is very limited at any given time.

app/controllers/oauth_authorized_applications_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+        last_used_at: tokens.map(&:last_used_at).compact.max,
+      }
+    end
+  end
+
+  def destroy
+    Doorkeeper.config.application_model.revoke_tokens_and_grants_for(params[:id], CurrentUser.user)
+    flash[:notice] = "Access revoked."
+    redirect_to(oauth_authorized_applications_path)
+  end
```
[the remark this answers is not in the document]
clragon: These methods exist and are provided by doorkeeper.

app/logical/oidc_signing_key.rb
```diff
@@ -0,0 +1,62 @@
...
+    bits = key.n.num_bits
+    if bits < 2048
+      raise "OIDC signing key is too weak: #{bits}-bit RSA (need at least 2048)"
+    end
+
+    if Rails.env.production? && dev_key?
+      raise "OIDC signing key matches the public demo key from docker-compose.yml. " \
+            "Generate a real key (`rake oidc:generate_key`) before booting in production."
+    end
+  end
```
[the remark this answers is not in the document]
clragon: Pem is never changed. This cannot possibly happen.

DonovanDMC:
Could this be expanded to have more scope for more fine grained access? Either now or in the future, if you want to go fully in with it you could do `<controller>.<action>` scopes, or just `<controller>`, `readonly`, etc for more controll on what applications can access

My old api keys overhaul (which you've reference before) GayFurCity/GayFurCity@51cfd950cc9d671846354dc3df14aaf8e8177e34 (I added translated names to some in GayFurCity/GayFurCity@51cfd950cc9d671846354dc3df14aaf8e8177e34, but it's more complicated than it needs to be)
I added granular permissions there, which is kinda similar to what I would envision here

clragon:
We are looking to implement a granular permission system in the future. Because that system is not ready at the moment, this PR will use just identity and full access as the two scopes. I am planning to revisit the topic then.

app/controllers/oauth_applications_controller.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+class OauthApplicationsController < Doorkeeper::ApplicationsController
+  layout "default"
+
+  skip_before_action :authenticate_admin!
+  before_action :janitor_only
+  before_action :reject_api_key_auth
+  before_action :reject_bearer_auth
```
Sindrake, line 9: PR description states that "OAuth cannot access the oauth api or api key api."
The former is true, but the API key controller does not include `before_action :reject_bearer_auth`.
It's currently blocked incidentally by `requires_reauthentication`, but that seems fragile.

I would prefer OAuth access to be explicitly rejected for the API key controller.
Please, add a corresponding test as well.

config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    normal_claim :email do |resource_owner|
+      resource_owner.email
+    end
+
+    normal_claim :email_verified do |resource_owner|
+      # email_verification_key.nil? is also true when email is blank.
+      resource_owner.email.present? && resource_owner.email_verification_key.nil?
+    end
```
Sindrake, line 92:
```suggestion
    normal_claim :email, scope: :email do |resource_owner|
      resource_owner.email
    end

    normal_claim :email_verified, scope: :email do |resource_owner|
      # email_verification_key.nil? is also true when email is blank.
      resource_owner.email.present? && resource_owner.email_verification_key.nil?
    end
```

According to the docs, the scope needs to be specified like this.
Otherwise, these claims will be gated behind `profile`. Please, verify.
Current tests only seem to check the happy path.

app/controllers/oauth_authorized_applications_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+class OauthAuthorizedApplicationsController < Doorkeeper::AuthorizedApplicationsController
+  layout "default"
+
+  skip_before_action :authenticate_resource_owner!
+  before_action :logged_in_only
+  before_action :reject_api_key_auth
+  before_action :reject_bearer_auth
+
+  def index
+    apps = Doorkeeper.config.application_model.authorized_for(CurrentUser.user)
```
Sindrake, line 12: I think I would prefer there to be a way for staff members to see all apps, not just their own.

app/javascript/src/js/pages/oauth_applications/credential_select.ts
Sindrake, line 1: I would strongly prefer both of these files to be in TypeScript.
We are slowly transitioning away from plain JS, and this would add future tech debt.

Sindrake (changes requested):

Sindrake (approved):

PR #2085, opened 2026-05-25
[IpAddrs] Fix errors caused by broken templates and unguarded inputs

### LINKED ISSUE #1944: Missing partial for IpAddrsController with json format
**Describe the bug**
Requesting the moderator IP address search page in JSON format (`/moderator/ip_addrs.json?search[ip_addr]=...`) produces a 500 `ActionView::MissingTemplate` error instead of returning JSON results.

**To Reproduce**
1. Log in as an admin.
2. Make a JSON request to `/moderator/ip_addrs.json` with a valid `search[ip_addr]` param (e.g. via the API or by appending `.json` to the URL while searching by IP address).
3. Observe a 500 error: `Missing partial moderator/ip_addrs/_user_listing.json`.

**Expected behavior**
The endpoint should return a JSON response containing the IP address search results.

**Additional context**
The root cause is an inconsistency in how partials are rendered between `index.html.erb` and `index.json.erb`:

- `index.html.erb` (works): `render "user_listing"` — no explicit format; Rails resolves the correct variant (`_user_listing.erb` or `_user_listing.html.erb`) from the request format.
- `index.json.erb` (broken): `render "user_listing.json"` — the `.json` suffix is treated as part of the literal partial name rather than a format specifier. Rails then looks for `_user_listing.json` with format `:json`, which requires a file named `_user_listing.json.json.erb` (which does not exist). The existing `_user_listing.json.erb` is never matched.

The fix is to change `render "user_listing.json"` (and `render "ip_listing.json"`) in `app/views/moderator/ip_addrs/index.json.erb` to `render "user_listing"` and `render "ip_listing"` respectively, consistent with the HTML template. Rails will then correctly select the `.json.erb` variant based on the request format.

mvanhorn:
Fixes #1944.

`/moderator/ip_addrs.json` raises `ActionView::MissingTemplate` because `index.json.erb` renders its partials by literal name with a `.json` suffix:

```erb
<%= render "ip_listing.json" %>
<%= render "user_listing.json" %>
```

Rails interprets the `.json` as part of the partial path, not a format specifier, so it looks for `_ip_listing.json` / `_user_listing.json` with request format `:json` — i.e. `_user_listing.json.json.erb` (which does not exist). The existing `_user_listing.json.erb` and `_ip_listing.json.erb` partials are never matched, and the response is a 500.

`index.html.erb` already renders these partials by their format-agnostic name (`render "user_listing"` / `render "ip_listing"`) and lets Rails resolve the right format-specific file from the request, so this change brings `index.json.erb` in line with that pattern.

### Repro

1. Log in as a moderator/admin.
2. `GET /moderator/ip_addrs.json?search[ip_addr]=<some ip>` (or `search[user_id]=...`).
3. Before this PR: 500 `Missing partial moderator/ip_addrs/_user_listing.json` (or `_ip_listing.json`).
4. After this PR: JSON body from the appropriate partial.

### DIFF
--- a/app/controllers/moderator/ip_addrs_controller.rb
+++ b/app/controllers/moderator/ip_addrs_controller.rb
@@ -8,13 +8,18 @@ module Moderator
     rescue_from IpAddrSearch::InvalidIpAddr, with: :render_invalid_ip_addr
 
     def index
-      search = IpAddrSearch.new(params[:search])
+      search = IpAddrSearch.new(params[:search] || {})
       @results = search.execute
       respond_with(@results)
     end
 
     def export
-      search = IpAddrSearch.new(params[:search].merge({with_history: true}))
+      unless params[:search].present? && (params[:search][:user_id].present? || params[:search][:user_name].present?)
+        render json: [], status: 200
+        return
+      end
+
+      search = IpAddrSearch.new((params[:search] || {}).merge({with_history: true}))
       @results = search.execute
       respond_with(@results) do |format|
         format.json do
--- a/app/views/moderator/ip_addrs/_ip_listing.json.erb
+++ b/app/views/moderator/ip_addrs/_ip_listing.json.erb
@@ -1,1 +1,1 @@
-<%= raw @results.map {|ip_addr, count| {ip_addr: ip_addr.to_s, count: count}}.to_json %>
+<%= raw @results.to_json %>
--- a/app/views/moderator/ip_addrs/index.json.erb
+++ b/app/views/moderator/ip_addrs/index.json.erb
@@ -1,5 +1,5 @@
 <% if params[:search][:user_id].present? || params[:search][:user_name].present? %>
-  <%= render "ip_listing.json" %>
+  <%= render "ip_listing" %>
 <% else %>
-  <%= render "user_listing.json" %>
+  <%= render "user_listing" %>
 <% end %>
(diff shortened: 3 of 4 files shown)


Sindrake (changes requested):
This technically fixes the issue, but exposes an underyling bug.

When searching by `user_id` or `user_name`, the `_ip_listings.json.erb` is rendered.
Unfortunately, that partial is complete nonsense. The `search_by_user_id` function returns a two-key Hash: `{sums: sums, ip_addrs: ip_addrs}`, which the partial tries to iterate over with `.map` to produce `[key, value]` pairs,. As a result, `ip_addr` becomes `:sums` or `:ip_addrs`, and `count` becomes their values.
At best, this returns garbage JSON. At worst, it just raises an error.

That partial could instead be as simple this:
```ERB
<%= raw @results.to_json %>
```

Note that there is a commented out test for this exact issue in [spec/requests/moderator/ip_addrs_controller_spec.rb:54-61](https://github.com/e621ng/e621ng/blob/master/spec/requests/moderator/ip_addrs_controller_spec.rb#L54-L61).
The bug should be fixed, and the spec in question should be uncommented.

Ideally, there should also be tests to verify that this endpoint returns data in the correct format for both the `user_id` / `user_name` and the `ip_addr` search params.

Sindrake (approved):

PR #2086, opened 2026-05-25
[Misc] Add HTML metatags to hint palette changing tools to not alter site colors

yivarld:
- Adds the [darkreader extension metatag](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#disabling-dark-reader-statically) to inform the site should not have it's colors changed automatically.
- Adds the [`colors-scheme` metatag](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/color-scheme) which should prevent other browsers from automatically changing the colors (for some reason). I set the value to `dark light`, since with the site themes, users can already enjoy darker and lighter palettes.

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -7,7 +7,8 @@
 <% unless disable_mobile_mode? %>
   <meta name="viewport" content="width=device-width,initial-scale=1">
 <% end %>
-
+<meta name="darkreader-lock">
+<meta name="color-scheme" content="light dark">
 <meta name="current-user-name" content="<%= CurrentUser.name %>">
 <meta name="current-user-id" content="<%= CurrentUser.id %>">
 <meta name="current-user-can-approve-posts" content="<%= CurrentUser.can_approve_posts? %>">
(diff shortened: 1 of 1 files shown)


app/views/layouts/_head.html.erb
```diff
@@ -7,7 +7,8 @@
 <% unless disable_mobile_mode? %>
   <meta name="viewport" content="width=device-width,initial-scale=1">
 <% end %>
-
+<meta name="darkreader-lock">
+<meta name="color-scheme" content="light dark">
```
Sindrake:
```suggestion
<meta name="color-scheme" content="dark light">
```

Using `light dark` goes against the PR description and intent.
The first listed value is the preferred scheme. Since e6 defaults to a dark-ish palette, `dark light` would be more accurate.

Please, verify.

Sindrake (changes requested):

app/views/layouts/_head.html.erb
yivarld: Ohh you're totally right!

Sindrake (approved):

PR #2101, opened 2026-05-28
[Search] Allow 'me' shortcut to be used across metatags

mm12:
Notes
* Added test for metatags even though they should all be the same
* Added test to ensure safe behavior from anonymous users
* Added guard to prevent 'downcase' from throwing if a non-string somehow gets passed

### DIFF
--- a/app/logical/tag_query.rb
+++ b/app/logical/tag_query.rb
@@ -1645,6 +1645,7 @@ class TagQuery
 
   def privileged_user_id_or_invalid(val)
     if CurrentUser.user.is_moderator?
+      return CurrentUser.id if val.is_a?(String) && val.casecmp?("me")
       User.name_or_id_to_id(val).presence
     elsif CurrentUser.user.is_authenticated?
       CurrentUser.id.presence
--- a/app/logical/tag_query.rb
+++ b/app/logical/tag_query.rb
@@ -1366,6 +1366,9 @@ class TagQuery
           post_set_id
         end
 
+      # NOTE: The favorite case is the only case where `user_id_or_invalid` isn't used, because it needs to check for the `hide_favorites` user setting.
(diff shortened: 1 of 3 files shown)


spec/logical/tag_query/user_metatags_spec.rb
```diff
@@ -38,6 +38,17 @@
...
       expect(tq[:uploader_ids_should]).to include(target_user.id)
     end
+
+    it "stores the current user's ID for user:me" do
+      tq = TagQuery.new("user:me")
+      expect(tq[:uploader_ids]).to include(CurrentUser.id)
+    end
+
+    it "skips for user:me on an anonymous session" do
+      CurrentUser.user.level = User::Levels::ANONYMOUS
```
Sindrake:
```suggestion
      CurrentUser.user = User.anonymous
```
A bit more resilient than setting the level to `ANONYMOUS`, which could have unpredictable behavior.

app/logical/tag_query.rb
```diff
@@ -1634,6 +1637,9 @@ def parse_boolean(value)
   end
 
   def user_id_or_invalid(val)
+    if val.respond_to?(:downcase) && val.downcase == "me"
```
Sindrake:
```suggestion
    if val.is_a?(String) && val.casecmp?("me")
```

Cleaner. Nitpick more than anything.

Sindrake (changes requested):
`upvote:me`, `downvote:me`, and `voted:me` still return the results for the user named "Me" for moderators.
Since that user never voted, I would suggest extending new `:me` behavior to moderators too.

There are also no tests for `noteupdater:me` and `deleter:me`.

mm12:
> There are also no tests for `noteupdater:me` and `deleter:me`.

When #1718 is merged it will be missing the that too, will need to add it depending on which is merged first.

Sindrake (approved):

PR #2102, opened 2026-05-28
[Build] Switch to Bun

Sindrake:
Alternative to #1895.

Pinned the dependencies in `package.json` to the versions we had in `yarn.lock`.


Sindrake:
Closed in favor of #1895.
May revisit in the future.

PR #2114, opened 2026-05-30
[PostReplacements] Scope original-backup ordering to single-post views

clragon:
This helps us actually paginate the index properly in the API, because otherwise postgres will have undeterministic orders for all entries with 0 as their id.


Sindrake (approved):
Looks good to me.

PR #2117, opened 2026-05-30
[CSS] Minor homepage tweaks

yivarld:
Changes how the navbar looks on the homepage so that all pages share the same similar layout that spans the whole screen width.
In order to make certain elements like the tool buttons container have proper contrast with the background, a dark gradient has been added, spamming the navbar height. (uses the regular background palette color). This is only shown on the homepage, and nowhere else.

This PR also;
- Makes the contents of the homepage centered vertically too, so the search box, mascot, etc, don't just stay stuck at the top of the screen if it is resized vertically.
- Makes the discord link on the navbar stay there at all times instead of disappearing at a certain screen width.
- Decouples styles used for the login profile buttons.

image


app/javascript/src/styles/common/navigation.scss
```diff
@@ -481,38 +468,27 @@ body[data-user-is-anonymous="true"] nav.navigation {
...
-      margin: 0 auto;
-
-      // Remove padding to prevent a scrollbar
-      // at low desktop resolutions
-      padding: 0 0 0.5rem;
-
-      #nav-discord, .nav-secondary, .nav-tools { display: none; }
+      grid-template-areas:
+        "logo primary  help  none  controls"
+        "logo primary  help  tools controls";
```
Sindrake:
```suggestion
        "logo primary  help  .     controls"
        "logo primary  help  tools controls";
```

app/javascript/src/styles/common/navigation.scss
```diff
@@ -454,19 +454,6 @@ nav.navigation, html.nav-toggled nav.navigation {
       #nav-toggle { display: none; }
       a { padding: 0; }
     }
-
-    // Icon collapse
-    // Stage 1: discord button
-    .collapse-1 { display: none; }
```
Sindrake, line 460: `collapse-1` and `collapse-2` styles were removed, but the actual classes are still present in the HTML.
If they are not needed, they should be removed entirely, to avoid future confusion.

app/javascript/src/styles/views/application/_avatar.scss
```diff
@@ -95,6 +95,11 @@ nav.navigation .simple-avatar {
   }
 }
 
+// do not show avatar name if we're on the home page
```
Sindrake: Comment is misleading, the username _is_ displayed on the home page, just without a background for some reason.

Sindrake (changes requested):

Sindrake (changes requested):
Image

Homepage now has a horizontal scrollbar on mobile.
This should not be happening.

yivarld:
I forgot to mention this; the removal of the background of the profile name label on the homepage only is because it is very noticeable with the new gradient. It's still there for the rest of pages, since the navbar items get culled by it at a certain point so that the label is still readable.

app/javascript/src/styles/views/application/_avatar.scss
```diff
@@ -96,9 +110,22 @@ nav.navigation .simple-avatar {
...
+    }
   }
 }
 
+// do not show avatar name background if we're on the home page and on desktop layout
+@include window-larger-than(50rem) {
+   body.a-home nav.navigation .simple-avatar .avatar-name {
+      background: none;
+   }
+}
```
Sindrake, line 127:
```suggestion
@include window-larger-than(50rem) {
  body.a-home nav.navigation .simple-avatar .avatar-name {
    background: none;
  }
}
```
This isn't [Dreamberd](https://github.com/TodePond/GulfOfMexico).

Sindrake (changes requested):

Sindrake (approved):

PR #2126, opened 2026-06-02
[PostReplacements] Fix "type" search

Sindrake:
The "Type" search was using an incorrect query parameter, and was missing the "webp" file type.


Sindrake:
Closed in favor of #2128.

PR #2146, opened 2026-06-04
[Favorites] Improve performance

clragon:
Swaps the favorite index from created_at to just id (same structure as votes table), which allows faster lookups on deep pages. Also fixes ordering on pages above 750. 

Increases the favorite limit to 200,000, moved into its own config var, although the limit could now (or later) be removed. 

Migration builds new indexes after deploy. Old index is retained if the new index fails to build. 

This PR has been tested on a local representative postgres setup with 6.4 mil posts, 1 mil users, 100 of which had 1 mil favorites, 1 mil of which had 1-40 favorites, 600k which had ~500 favorites.

Comparing, with a 1 mil favs user:
In 2023 (limit introduction) page 1 was 2000ms, page 2 was 2000ms, page 750 was 2000ms. 
After PR #1080, page 1 was 0.15ms, page 2 was 0.19ms, page 750 was 106ms.
With this PR, Page 1 is 0.15ms, page 2 is 0.19ms, page 750 is 47ms.

### DIFF
--- a/app/components/paginator_component.rb
+++ b/app/components/paginator_component.rb
@@ -48,7 +48,7 @@ class PaginatorComponent < ViewComponent::Base
 
   def prev_path
     return nav_params_for(current_page - 1) if @mode == :numbered
-    nav_params_for("a#{records.first&.id || 0}")
+    nav_params_for("a#{sequential_first_id}")
   end
 
   def has_next?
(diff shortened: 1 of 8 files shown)


clragon:
With the current index already dominating the chart on gains, I believe, the churn from completely rebuilding the favorite index is not worth the minimal performance improvements at the moment, even if this would help with slightly faster deep pages. 

Raising or disabling the favorite limit would now be quite feasible, even without merging this.

PR #2147, opened 2026-06-04
[Session] DELETE now redirects to referer if possible

yivarld:
The session DELETE method now redirects the user back to where it was located using the referer header. If not present, the root path is used (instead of `/posts`)

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -28,7 +28,7 @@ class SessionsController < ApplicationController
       RateLimiter.hit("login:#{request.remote_ip}", 6.hours)
       DanbooruLogger.add_attributes("user.login" => "fail")
       respond_to do |fmt|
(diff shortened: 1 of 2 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -38,7 +38,7 @@ def destroy
     session.delete(:user_id)
     cookies.delete(:remember)
     session.delete(:last_authenticated_at)
-    redirect_to(posts_path, notice: "You are now logged out")
+    redirect_to(request.referer || root_path, notice: "You are now logged out")
```
Sindrake:
```suggestion
    redirect_back_or_to(root_path, notice: "You are now logged out")
```
More idiomatic, and avoids an error if the referrer is somehow external.

Sindrake (changes requested):

Sindrake (approved):

PR #2150, opened 2026-06-05
[Posts] Reowner expanded

DRLaDRLa:
Add a `Reowner` action on posts for janitors:
- Opens a dialog with a dropdown list of previous post owners
  - A previous post owner is a creator for an "approved" or "original" replacement
  - The list is deduplicated and excludes the current owner
  - If the user is an admin, there's also an "Other..." option
- The janitor picks the new owner from the list and confirms
  - The post owner is changed
  - A post event is created
  - Post version owners remain **unchanged**
- An admin can also pick the "Other..." option
  - A text input appears 
  - The admin enters any user name, or user ID prefixed with `!`
  - The admin confirms and the rest is the same as for a janitor
- A BD admin has additional options to override the post version and event behaviour
  - Create post events (defaults to true)
  - Reowner post versions (defaults to false)

Improve the `/admin/reowner/new` endpoint
- Expose previously hardcoded behaviour as options
  - Create post events (defaults to false)
  - Reowner post versions (defaults to true)
- Show the number of reownered post in the notice

Implementation details:
- The list of previous owners is lazily loaded from a new `GET /moderator/post/posts/:id/previous_owners.json` endpoint

This relates to #1449, while it implements the same feature in a more roundabout way, it's really intended to fix owners on already replaced posts.

Allowing admins to arbitrarily reowner is a new thing, this is currently BD staff only using the `/admin/reowner/new` endpoint, but it's also a less powerful version of it:
- One post at a time
- Always creates post events
- Doesn't change post version owners

In any case I can add config values for the permissions if deemed necessary.

### DIFF
--- a/app/controllers/admin/reowner_controller.rb
+++ b/app/controllers/admin/reowner_controller.rb
@@ -12,6 +12,7 @@ module Admin
       @old_user = User.find_by_name_or_id(@reowner_params[:old_owner])
       @new_user = User.find_by_name_or_id(@reowner_params[:new_owner])
       query = @reowner_params[:search]
+      reowner_versions = @reowner_params[:reowner_versions]&.truthy?
       post_events = @reowner_params[:post_events]&.truthy?
 
       unless @old_user && @new_user
@@ -23,7 +24,7 @@ module Admin
       moved_post_ids = []
       Post.tag_match("user:!#{@old_user.id} #{query}").limit(300).each do |p|
         moved_post_ids << p.id
-        p.reowner!(@new_user, @old_user, reowner_versions: true, post_events: post_events)
+        p.reowner!(@new_user, @old_user, reowner_versions: reowner_versions, post_events: post_events)
       end
 
       StaffAuditLog.log(:post_owner_reassign, CurrentUser.user, { old_user_id: @old_user.id, new_user_id: @new_user.id, query: query, post_ids: moved_post_ids })
@@ -34,7 +35,9 @@ module Admin
     private
 
     def new_params
-      params.require(:reowner).permit(%i[old_owner search new_owner post_events])
+      params.require(:reowner)
+            .permit(%i[old_owner search new_owner reowner_versions post_events])
+            .with_defaults(reowner_versions: true)
     end
   end
 end
--- a/app/controllers/admin/reowner_controller.rb
+++ b/app/controllers/admin/reowner_controller.rb
@@ -12,8 +12,8 @@ module Admin
       @old_user = User.find_by_name_or_id(@reowner_params[:old_owner])
       @new_user = User.find_by_name_or_id(@reowner_params[:new_owner])
       query = @reowner_params[:search]
-      reowner_versions = @reowner_params[:reowner_versions]&.truthy?
-      post_events = @reowner_params[:post_events]&.truthy?
+      reowner_versions = ActiveModel::Type::Boolean.new.cast(@reowner_params[:reowner_versions])
+      post_events = ActiveModel::Type::Boolean.new.cast(@reowner_params[:post_events])
 
       unless @old_user && @new_user
         flash[:notice] = "Old or new user failed to look up. Use !id for name to use an id"
--- a/app/controllers/admin/reowner_controller.rb
+++ b/app/controllers/admin/reowner_controller.rb
@@ -23,15 +23,7 @@ module Admin
       moved_post_ids = []
       Post.tag_match("user:!#{@old_user.id} #{query}").limit(300).each do |p|
         moved_post_ids << p.id
-        p.do_not_version_changes = true
-        p.update({ uploader_id: @new_user.id })
-        p.versions.where(updater_id: @old_user.id).find_each do |pv|
-          pv.update_column(:updater_id, @new_user.id)
-          pv.update_index
-        end
-        if post_events
-          PostEvent.add(p.id, CurrentUser.user, :owner_changed, { old_owner: @old_user.id, new_owner: @new_user.id })
-        end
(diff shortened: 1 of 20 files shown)


clragon:
> Add an option to create post events (defaults to false)

Is it a good idea to give admins an option and default to bypass logging the change?

DRLaDRLa:
> > Add an option to create post events (defaults to false)
> 
> Is it a good idea to give admins an option and default to bypass logging the change?

I maintained the same default behaviour for the existing reowner endpoint. There wasn't a reowner post event before, so I decided to keep it that way, since I'm not sure exactly what the original intent was. There is always a staff log entry added though (it's a BD staff endpoint actually).

The new endpoint on posts accessible to the rest of staff always logs a post event.

app/javascript/src/js/pages/posts/posts.js
```diff
@@ -218,6 +219,83 @@ Post.initialize_links = function () {
...
+      $(document).on("keydown", (event) => {
+        // .isFocused would make more sense if it existed
+        if (event.key === "Enter" && reownerDialog.isOpen) {
+          event.preventDefault();
+          form.trigger("submit");
+        }
+      });
+    }
+
+    form.on("submit", event => {
```
Sindrake: This binds a new event handler on every reowner link click.

app/views/posts/partials/show/sidebar/_options.html.erb
```diff
@@ -59,6 +59,7 @@
       <% end %>
     <% end %>
     <% if CurrentUser.is_janitor? %>
+      <li><%= tag.a "Reowner", href: "#", id: "reowner-post-link", data: { pid: post.id } %></li>
```
Sindrake, line 62: This link should probably be gated behind `CurrentUser.can_approve_posts?`, like the reowner dialog itself.

app/controllers/moderator/post/posts_controller.rb
```diff
@@ -123,6 +123,25 @@ def ai_check
         @ai_result = @post.check_for_ai_content
         redirect_back fallback_location: post_path(@post)
       end
+
+      def previous_owners
+        @post = ::Post.find(params[:id])
+        @previous_owners = @post.previous_version_uploaders
+        respond_with(@previous_owners) do |format|
+          format.json { render json: @previous_owners.to_json }
```
Sindrake: This could probably just return IDs and names.
Minor nitpick.

app/javascript/src/js/pages/posts/posts.js
Sindrake, line 1: I would **STRONGLY** prefer if you extracted this code from `posts.js`, and instead created a separate TypeScript file that is only loaded on `posts#show` pages.

We should be moving away from plain JS, not expanding it.
The `posts.js` file specifically should be getting split apart into multiple more focused files.

app/views/moderator/post/posts/_reowner_dialog.html.erb
```diff
@@ -0,0 +1,15 @@
+<form class="simple_form" id="reowner-dialog" data-title="Reowner Post" data-width="500" data-position="center">
```
Sindrake:
```suggestion
<form class="simple_form" id="reowner-dialog" data-title="Reowner Post" data-position="center">
```
Dialog width is set to 250 in the script.

app/controllers/admin/reowner_controller.rb
```diff
@@ -21,23 +24,20 @@ def create
...
       redirect_back fallback_location: new_admin_reowner_path
     end
 
     private
 
     def new_params
-      params.require(:reowner).permit(%i[old_owner search new_owner])
+      params.require(:reowner)
+            .permit(%i[old_owner search new_owner reowner_versions post_events])
+            .with_defaults(reowner_versions: true)
```
Sindrake, line 40: Using the default boolean value `true` will throw an error, since `truthy?` on line 15 is defined only on String.

spec/models/post/uploader_methods_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+        expect(replacement3.creator).to eq(user1)
+        expect(post.previous_version_uploaders).to eq([user1, user2])
+      end
+
+      it "returns empty if the post was never replaced" do
+        post = create(:post)
+        expect(post.previous_version_uploaders).to eq([])
+      end
+
+      it "returns doesn't return pending or rejected replacement uploaders" do
```
Sindrake:
```suggestion
      it "doesn't return pending or rejected replacement uploaders" do
```

Sindrake (commented):
Looks good, functionality-wise.
I would say that using TypeScript is a requirement at this point, though.

app/views/posts/partials/show/sidebar/_options.html.erb
DRLaDRLa: Actually I was thinking this feature should be janitor only, so it's other one I'll change. If you think it should be gated on the approver flag instead let me know.

app/views/moderator/post/posts/_reowner_dialog.html.erb
DRLaDRLa: I'll remove it from the script and make this 250 instead.

DRLaDRLa:
~~Note to self: I need to test the staff audit log.~~ Done

app/views/posts/partials/show/sidebar/_options.html.erb
Sindrake: Sure, janitor-only is reasonable.

Sindrake:
There are more things that I could complain about, but I do not have the patience to deal with this PR anymore.

PR #2153, opened 2026-06-07
[PostFlags] Move AI flag config values to settings

DRLaDRLa:
- Move the AI flag stuff to settings
  - Omitted from the original DB flag reasons PR to be figured out later
- Minor flag reasons improvements after feedback
- Remove old flag reason config values

### DIFF
--- a/app/controllers/post_flag_reasons_controller.rb
+++ b/app/controllers/post_flag_reasons_controller.rb
@@ -59,9 +59,9 @@ class PostFlagReasonsController < ApplicationController
   end
 
   def set_ai_flag_reason
-    params = ai_flag_reason_params
+    ai_params = ai_flag_reason_params
 
-    reason = PostFlagReason.by_name(params[:reason].to_s)
+    reason = PostFlagReason.by_name(ai_params[:reason].to_s)
     if reason.blank? || reason.needs_parent_id?
       flash[:alert] = "Flag reason doesn't exist or is not usable for AI flagging"
       redirect_to post_flag_reasons_path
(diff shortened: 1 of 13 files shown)


Sindrake (changes requested):
Separate the removal of old flag reasons into their own PR, please.

DRLaDRLa:
> Separate the removal of old flag reasons into their own PR, please.

#2155

app/controllers/post_flag_reasons_controller.rb
```diff
@@ -53,9 +58,35 @@ def clear_cache
     redirect_to post_flag_reasons_path
   end
 
+  def set_ai_flag_reason
+    params = ai_flag_reason_params
```
Sindrake: Shadowing `ActionController::Base#params` with a local variable like this is a bit of a footgun.
Should probably be something like `ai_params` instead.

app/views/post_flag_reasons/index.html.erb
```diff
@@ -21,6 +21,26 @@
       <%= render partial: "reason_row", collection: @reasons, as: :flag_reason %>
       </tbody>
     </table>
+
+    <%= custom_form_for(:ai_flag_reason, url: set_ai_flag_reason_post_flag_reasons_path, method: :post, local: true, class: "simple_form") do |f| %>
```
Sindrake: I would throw some `margin-top` on this.
Looks weird otherwise.

Sindrake (changes requested):
Looks good overall.
Could use a couple minor tweaks.

Sindrake (approved):

PR #2164, opened 2026-06-10
[Posts] Add advanced search panel with rating, sort, and in-pool filters

DoubleThePsycho:
Adds an advanced search panel to the post index so common search options can be changed without manually typing metatags into the search box. Right now it covers rating, sort order, and whether a post is in a pool, while still keeping the actual query text in sync so the normal search behavior is preserved.

All functionalities have been tested on Firefox for Windows 11 and Android.

output

<details>
<summary>View the over-complicated, AI-generated description instead.</summary>

## Summary

This PR adds an **Advanced Search panel** to the post search bar, giving users interactive UI controls for the most common search filters — rating, sort order, and pool membership — without requiring them to write metatags by hand.

---

## Advanced Search Panel

A new panel (`#advanced-search-container`) opens via the new button added to the search controls toolbar.
The panel contains three sections:

### Rating
Three checkboxes (Safe / Questionable / Explicit), color-coded to their rating, that read and write `rating:`/`-rating:` metatags. Uses the most compact token representation to conserve tag budget (site limit: 40) — all three checked produces no token, Safe only produces `rating:s`, Safe + Questionable produces `-rating:e`.

### Sort Order
Four static options (Default, Score, Hot, Favorites) are always shown. When the current query contains any other recognised `order:` value — such as `order:id`, `order:favcount_asc`, `order:random` — a fifth entry is dynamically appended showing the value's label and icon. Unrecognised order values (e.g. `order:foobar`) append a "Custom" entry with a pencil icon instead.

Sort options that support a `_asc` variant (everything except `random`, `hot`, `landscape`, `portrait`) show a small square toggle button flush with the right edge of their label. Clicking it toggles the query between descending and ascending order without disturbing the rest of the query.

### In Pool
A three-state toggle (Any / Yes / No) that writes `inpool:true` or `inpool:false`, or removes the token entirely when set back to Any.

---

## Query Parsing

A full token-level query parser was added to support the above:

- **`scan_top_level_tokens`** — tokenises a query string respecting quoted strings and parenthesised groups, returning only top-level tokens with their character positions.
- **`parse_order_token`** — handles `order:value`, `order:value_asc`, `order:value_desc`, and negated forms (`-order:…`), resolving each to a `{ value, direction }` pair.
- **`parse_inpool_token`** / **`parse_rating_token`** — simple parsers for their respective metatags.
- **`replace_top_level_metatags`** — removes all existing tokens matching a predicate and optionally inserts a replacement, operating right-to-left on character positions to keep offsets stable.

The panel syncs bidirectionally with the textarea: typing directly updates the controls, and interacting with the controls rewrites the relevant metatags in the query while leaving everything else untouched.

---

## Controls Toolbar Restructure

The old single `#search-settings` button and `.search-settings-container` panel have been split into two independent buttons and panels:

| Old | New |
|---|---|
| `#search-fullscreen` | `#toggle-fullscreen` |
| `#search-settings` | `#advanced-search-open` + `#layout-settings-open` |
| `.search-settings-container` | `#advanced-search-container` + `#layout-settings-container` |

SCSS updated across `_index.scss`, `_index.desktop.scss`, `_show.scss`, and `_show.desktop.scss` to match.

---

## Fullscreen Button Refactor

Previously the fullscreen toggle button embedded two server-rendered SVGs in the DOM purely so JavaScript could extract their `outerHTML` at runtime — an awkward pattern. The button is now empty in ERB and renders its `sidebar_hide`/`sidebar_show` icons directly via `SVGIcon.render()`, removing the server-side dependency entirely.

---

## Icons

**`icon_helper.rb`** (server-side, used in ERB):
`rating_s`, `rating_q`, `rating_e`, `layout`, `sliders`

**`SVGIcon.ts`** (client-side, used by dynamic JS):
`hash`, `shuffle`, `pencil`, `sidebar_hide`, `sidebar_show`, `clock_fading`, `file`, `fullscreen`, `images`, `message_square`, `notepad`, `reset`, `tags`, `star`, `flame`, `trending_up`

The `star`, `flame`, and `trending_up` icons are registered in `SVGIcon.ts` as a precaution — they correspond to the statically rendered Score, Hot, and Favorites panel buttons and are not reachable via the dynamic sort entry path under current conditions, but would be needed if those static buttons were ever removed.

---

## Notes

- The offclick handler for the advanced search panel (closing on outside click) has an intentionally empty callback pending a decision from Cinder.
</details>


app/javascript/src/js/pages/posts/post_search.js
Sindrake, line 1: I understand the desire to reuse existing files, but you are adding 500+ lines of plain JS code here.
We _should_ be transitioning to TypeScript.

This file also uses the outdated approach of creating an object and adding functions to it, rather than just using a class. That is something we should be eliminating, not perpetuating.

app/javascript/src/js/pages/posts/post_search.js
```diff
@@ -29,6 +80,357 @@ PostSearch.initialize_input = function ($form) {
   }
 };
 
+PostSearch.initialize_advanced_search_details = function ($section) {
+  const $details = $section.find("details.post-advanced-search").first();
+  if (!$details.length) return;
+
+  $details.prop("open", LStorage.Posts.AdvancedSearchOpen);
+  $details.on("toggle", (event) => {
+    LStorage.Posts.AdvancedSearchOpen = event.currentTarget.open;
```
Sindrake: `LStorage.posts.AdvancedSearchOpen` does not exist.

At its core, this _is_ correct – it does not make a ton of sense to persist the state of the advanced search window between page loads. But at the same time, LStorage is explicitly _pesistent_ storage. Creating an ephemeral JS property like this is confusing.

I would suggest just using a PostSearch object property instead. Or just a variable.

app/javascript/src/js/pages/posts/post_search.js
```diff
@@ -29,6 +80,357 @@ PostSearch.initialize_input = function ($form) {
   }
 };
 
+PostSearch.initialize_advanced_search_details = function ($section) {
```
Sindrake: I don't think that this function is called anywhere.

app/javascript/src/js/utility/SVGIcon.ts
```diff
@@ -31,9 +31,26 @@ export default class SVGIcon {
...
+    "images": '<path d="M18 22H4a2 2 0 0 1-2-2V6"/><path d="m22 13-1.296-1.296a2.41 2.41 0 0 0-3.408 0L11 18"/><circle cx="12" cy="8" r="2"/><rect width="16" height="16" x="6" y="2" rx="2"/>',
+    "message_square": '<path d="M21 15a2 2 0 0 1-2 2H7l-4 4V5a2 2 0 0 1 2-2h14a2 2 0 0 1 2 2z"/>',
+    "notepad": '<path d="M8 2v4"/><path d="M12 2v4"/><path d="M16 2v4"/><path d="M16 4h2a2 2 0 0 1 2 2v2"/><path d="M20 12v2"/><path d="M20 18v2a2 2 0 0 1-2 2h-1"/><path d="M13 22h-2"/><path d="M7 22H6a2 2 0 0 1-2-2v-2"/><path d="M4 14v-2"/><path d="M4 8V6a2 2 0 0 1 2-2h2"/><path d="M8 10h6"/><path d="M8 14h8"/><path d="M8 18h5"/>',
+    "reset": '<path d="M3 12a9 9 0 1 0 9-9 9.75 9.75 0 0 0-6.74 2.74L3 8"/><path d="M3 3v5h5"/>',
+    "tags": '<path d="m15 5 6.3 6.3a2.4 2.4 0 0 1 0 3.4L17 19"/><path d="M9.586 5.586A2 2 0 0 0 8.172 5H3a1 1 0 0 0-1 1v5.172a2 2 0 0 0 .586 1.414L8.29 18.29a2.426 2.426 0 0 0 3.42 0l3.58-3.58a2.426 2.426 0 0 0 0-3.42z"/><circle cx="6.5" cy="9.5" r=".5" fill="currentColor"/>',
   };
 
-  static render (name: string): SVGElement | null {
+  static render (name: string, strokeWidth = 2): SVGElement | null {
```
Sindrake: I personally would leave the `strokeWidth` hard-coded to 2.
The custom order icons look fine like that, in my opinion.

app/javascript/src/js/pages/posts/post_search.js
```diff
@@ -62,29 +464,69 @@ PostSearch.initialize_wiki_preview = function ($preview) {
...
 
 PostSearch.initialize_controls = function () {
+  // Advanced search panel
+  const advOffclickHandler = Offclick.register("#advanced-search-open", "#advanced-search-container", () => {
+    // CLOSE ADVANCED OPTIONS
+    // advSearch.removeClass("active");
+    // advSearchBtn.removeClass("active");
+    // LStorage.Posts.AdvancedSearchOpen = false;
+    // CLOSE ADVANCED OPTIONS
+  });
```
Sindrake:
```suggestion
  const advOffclickHandler = Offclick.register("#advanced-search-open", "#advanced-search-container, .search", () => {
    advSearch.removeClass("active");
    advSearchBtn.removeClass("active");
    LStorage.Posts.AdvancedSearchOpen = false;
  });
```

I think closing the advanced search menu when the user clicks outside of it is fine.
By adding `.search` to the menu string, we can also prevent it from closing when the user interacts with the search UI.

app/javascript/src/js/pages/posts/post_search.js
```diff
@@ -29,6 +80,357 @@ PostSearch.initialize_input = function ($form) {
...
+        $controls.find(`#${sort_custom}`).prop("checked", true);
+        if (entry?.asc) addAscBtn(sort_custom, direction);
+      }
+    }
+  };
+
+  const inpool_states = ["unset", "yes", "no"];
+  const inpool_values = { unset: "", yes: "true", no: "false" };
+  const inpool_state_map = { "": "unset", "true": "yes", "false": "no" };
+  const getInpool = () => $("[data-advanced-search=inpool]").first();
```
Sindrake: Isn't `getInpool()` returning the same thing as `$inpool`?

app/javascript/src/js/pages/posts/post_search.js
```diff
@@ -29,6 +80,357 @@ PostSearch.initialize_input = function ($form) {
...
+  const $details = $section.find("details.post-advanced-search").first();
+  if (!$details.length) return;
+
+  $details.prop("open", LStorage.Posts.AdvancedSearchOpen);
+  $details.on("toggle", (event) => {
+    LStorage.Posts.AdvancedSearchOpen = event.currentTarget.open;
+  });
+};
+
+PostSearch.initialize_advanced_search = function ($section) {
```
Sindrake: This one massive function with multiple functions inside of it is driving me up the wall.
We don't exactly follow Clean Code principles  for function length in this repo, but this is still a little ridiculous.

app/javascript/src/js/pages/posts/post_search.js
```diff
@@ -29,6 +80,357 @@ PostSearch.initialize_input = function ($form) {
...
+  if ($sortInputs.length) $textarea.closest("form").on("submit", () => $sortInputs.prop("disabled", true));
+
+  $textarea.on("input", syncControls);
+  $controls.on("change", "[name='advanced-search-sort']", updateOrder);
+  $controls.on("click", ".sort-asc-btn", (event) => {
+    event.preventDefault();
+    event.stopPropagation();
+    toggleAsc();
+  });
+  $(document).on("click", "[data-advanced-search=inpool]", updateInpool);
```
Sindrake: Why bind this to `$(document)` rather than `$controls`?

app/views/posts/partials/index/_controls.html.erb
```diff
@@ -1,16 +1,107 @@
...
+  </div>
+
+  <div class="ssc-entry">
+
+    <div class="ssc-label">
+      <label for="ssc-in-pool">In pool</label>
+    </div>
+
+    <div class="ssc-body">
+      <%= tag.label(class: "sto-toggle sto-toggle--tri", id: "scc-in-pool", data: { advanced_search: "inpool", state: "unset" }) do %>
```
Sindrake:
```suggestion
      <%= tag.label(class: "sto-toggle sto-toggle--tri", id: "ssc-in-pool", data: { advanced_search: "inpool", state: "unset" }) do %>
```

Typo?

app/javascript/src/js/utility/SVGIcon.ts
```diff
@@ -31,9 +31,26 @@ export default class SVGIcon {
     "face_frown": '<circle cx="12" cy="12" r="10"/><path d="M16 16s-1.5-2-4-2-4 2-4 2"/><line x1="9" x2="9.01" y1="9" y2="9"/><line x1="15" x2="15.01" y1="9" y2="9"/>',
 
     "times": '<path d="M18 6 6 18"/><path d="m6 6 12 12"/>',
+    "star": '<path d="M11.525 2.295a.53.53 0 0 1 .95 0l2.31 4.679a2.123 2.123 0 0 0 1.595 1.16l5.166.756a.53.53 0 0 1 .294.904l-3.736 3.638a2.123 2.123 0 0 0-.611 1.878l.882 5.14a.53.53 0 0 1-.771.56l-4.618-2.428a2.122 2.122 0 0 0-1.973 0L6.396 21.01a.53.53 0 0 1-.77-.56l.881-5.139a2.122 2.122 0 0 0-.611-1.879L2.16 9.795a.53.53 0 0 1 .294-.906l5.165-.755a2.122 2.122 0 0 0 1.597-1.16z"/>',
```
Sindrake: We should be keeping the icons included in `SVGIcon.ts` to an absolute bare minimum, since every single icon is being served to the user on every single page.

Anything that could be rendered server-side should be rendered server-side.
Anything new icon added here should be well-justified.

app/javascript/src/styles/common/post_search.scss
```diff
@@ -8,12 +8,13 @@
...
     line-height: 1.5rem;
   }
+
   .post-search-help {
     font-size: 90%;
     line-height: 1rem;
 
     // Make way for the settings button
-    margin-right: 2.5rem;
+    margin-right: 5rem;
```
Sindrake, line 17: Image

Buttons cover up the text.

Sindrake (changes requested):
Functionality looks good.
I would say that refactoring this to use Typescript and a class structure is necessary.

DoubleThePsycho:
Addressed most of the feedback. The two remaining comments (function length and TypeScript migration) are noted — I can tackle them in this PR if you'd prefer, or defer to a follow-up. Let me know.

app/javascript/src/js/pages/posts/SearchFilters.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (!inputId) return;
+    const $label = this.$controls.find(`label[for='${inputId}']`);
+    $("<button>")
+      .attr({
+        "type": "button",
+        "aria-label": "Toggle ascending",
+      })
+      .addClass("sort-asc-btn")
+      .toggleClass("active", direction === ORDER_ASC)
+      .text("-")
```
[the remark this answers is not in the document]
DoubleThePsycho: Just throwing my two cents in here, I think the arrows would look a lot better than a static `-`.

But considering you rewrote this entire soup, up to you @Sindrake

Sindrake (approved):

Issue #1827, opened 2026-04-15
Login-blocked posts: file URLs reconstructable via API and files directly accessible without authentication


Sindrake:
Posts that are login-blocked (`loginblocked?`) correctly show a "You must be logged in to view this image" message in the frontend, and the API correctly returns `null` for `file.url`, `preview.url`, and `sample.url` fields for unauthenticated requests. However, the protection is incomplete in two ways:

## 1. `md5` and `ext` are unconditionally exposed in API responses

`PostBlueprint` always includes `file.md5` and `file.ext` in its response, regardless of `post.visible?`. Since file URLs follow a deterministic pattern based on these two values (`/data/{md5[0..1]}/{md5[2..3]}/{md5}.{ext}`), any anonymous user can reconstruct the full file URL from the API response alone.

The legacy serializer (`hidden_attributes` in `Post`) correctly suppresses `md5` and `file_ext` when a post is not visible, but `PostBlueprint` is a separate code path that does not replicate this behaviour.

Similarly, the `sample.alternates` field (`video_sample_list`) is included unconditionally and contains direct video URLs.

## 2. Login-blocked post files are served from the unauthenticated public path

`protect_file?` on the `Post` model only returns `true` for deleted posts:

```ruby
def protect_file?
  is_deleted?
end
```

Deleted posts are moved to `/data/deleted/…` and their URLs require a time-limited HMAC token. Login-blocked posts are not deleted, so their files remain at `/data/…` with no access control whatsoever — a direct HTTP request to the file URL succeeds for any unauthenticated client.

## Expected behaviour

Accessing the file URL of a login-blocked post without a valid session should be denied (or at minimum, `md5` and `ext` should be suppressed from the API response to prevent trivial URL reconstruction).

## Affected code

- `app/blueprints/post_blueprint.rb` — `md5`/`ext` exposed unconditionally; `sample.alternates` exposed unconditionally
- `app/models/post.rb` — `protect_file?` does not account for `loginblocked?`

DonovanDMC:
I'm begging you to not fully login protect the media of any non-deleted posts, that fully breaks the assumption of being able to parse the posts export and fetch any non-deleted post without ever touching the api

If you do go through with it, could some scheme be used where we wouldn't need to hit the api for each post? Something like a per-user secret that you could fetch once and use to generate some token, or even something with api keys (though I know that's even further of a long shot)

[...]

PR #2166, opened 2026-06-10
[Blueprints] Hide md5/ext and video samples for non-visible posts

mvanhorn:
## Summary
Login-blocked posts null out their URL fields in the v2 JSON API, but `PostBlueprint` still exposes `files.meta.md5` and `files.meta.ext`, and video posts expose direct sample URLs in `files.video`. With the deterministic `/data/{md5[0..1]}/{md5[2..3]}/{md5}.{ext}` path, that is enough for an anonymous user to reconstruct the full file URL. This change nulls `md5`/`ext` and omits the video sample list when `post.visible?` is false, so the Blueprinter path matches what the legacy serializer already does.

## Why this matters
Sindrake's report in #1827 describes the leak: the legacy serializer path suppresses `md5` and `file_ext` for non-visible posts via `Post::ApiMethods#hidden_attributes` (app/models/post.rb), but the newer Blueprinter path never replicated that behavior, so the v2 API quietly undoes the access control. The fix extends the `post.visible?` guard that already exists in the same `files` block for the URL fields.

Scope note: this covers the API response leak only. The second part of the issue, protecting the static file paths at the serving layer, drew pushback from API consumers in the thread and needs a maintainer decision on an access scheme, so it is intentionally untouched here.

## Testing
New request specs in `spec/requests/posts_controller_spec.rb` cover both directions: a visible v2 post still returns `md5`, `ext`, and the full `files.video` structure, and a login-blocked post returns null `md5`/`ext`, null URLs, intact size/width/height metadata, and no `video` key. The specs follow the existing v2 patterns in that file and run in the dockerized test environment.

Refs #1827

AI was used for assistance.

### DIFF
--- a/app/blueprints/post_blueprint.rb
+++ b/app/blueprints/post_blueprint.rb
@@ -14,11 +14,12 @@ class PostBlueprint < Blueprinter::Base
   ### File Information ###
 
   field :files do |post|
+    visible = post.visible?
     output = {}
 
     output[:meta] = {
-      md5: post.md5,
-      ext: post.file_ext,
+      md5: visible ? post.md5 : nil,
+      ext: visible ? post.file_ext : nil,
       size: post.file_size,
       duration: post.duration&.to_f,
 
@@ -45,7 +46,7 @@ class PostBlueprint < Blueprinter::Base
       webp: nil,
     }
 
-    if post.visible?
+    if visible
       output[:original][:url] = post.file_url
 
       preview = post.preview_file_url_pair
@@ -57,7 +58,7 @@ class PostBlueprint < Blueprinter::Base
       output[:sample][:jpg] = sample[1]
     end
 
-    if post.is_video?
+    if visible && post.is_video?
       output[:video] = post.video_sample_list
     end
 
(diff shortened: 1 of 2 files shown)


DonovanDMC:
https://github.com/e621ng/e621ng/blob/ac21366119739d51beced13fd2cf290eb5625f1d/app/models/post.rb#L1648-L1654
https://github.com/e621ng/e621ng/blob/ac21366119739d51beced13fd2cf290eb5625f1d/app/serializers/post_serializer.rb#L13-L20

The md5 and ext has never been suppressed, the code was indeed there but the attributes were nested under file so it did nothing

In the same vein as https://github.com/e621ng/e621ng/issues/1827#issuecomment-4445120885 **please** do not hide this info, it's useful to know the md5 of a post regardless of what the content of the post is (in the case of deleted posts it can be used for matching to find the post for other sources)

Reconstructing urls be damned, it has worked fine for decades and the requirement for a login would block a decent number of chibi artists (`young adult_(lore)`) from ever showing up in things like scrollers and slideshows where no auth is present, but urls can still be reconstructed

mvanhorn:
That's fair, and thanks for the context. The deleted-post matching use case and the unauthenticated scroller/slideshow scenarios are good reasons to keep md5/ext exposed, and given the old suppression never actually applied, hiding them now would be a behavior regression rather than a fix. Closing this. If you'd ever want the video-samples part handled separately, happy to spin that out, but I'll take the direction on #1827 from you.

PR #2167, opened 2026-06-10
[Uploads] Add pairing tags based on selected genders

DonovanDMC:
This removes the intersex tags from the uploader, and replaces other pairing tags with a filtered list based on the selected gender tags
This works with tags being specified in the query (`http://localhost:3000/uploads/new?tags=male`), as well as deselecting all affected pairs when the gender tag is deselected
image

### DIFF
--- a/app/javascript/src/js/pages/uploads/new/uploader.vue
+++ b/app/javascript/src/js/pages/uploads/new/uploader.vue
@@ -653,7 +653,10 @@
         const selected = this.checkboxes.selected;
         return all_pairing_pairs
           .filter(p => selected[p.tagA] && selected[p.tagB])
-          .map(p => ({ name: `${sex_names[p.tagA]}/${sex_names[p.tagB]}`, tag: `${pairing_tag_name(p.tagA)}/${pairing_tag_name(p.tagB)}` }));
+          .map(p => ({
+            name: sex_names[p.tagA] + "/" + sex_names[p.tagB],
+            tag: pairing_tag_name(p.tagA) + "/" + pairing_tag_name(p.tagB),
+          }));
       },
       tags() {
         const self = this;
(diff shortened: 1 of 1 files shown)


DonovanDMC:
Not sure what the failure is about, but I don't think it's on me

app/javascript/src/js/pages/uploads/new/uploader.vue
```diff
@@ -640,12 +649,21 @@
       },
     },
     computed: {
+      filteredPairings() {
+        const selected = this.checkboxes.selected;
+        return all_pairing_pairs
+          .filter(p => selected[p.tagA] && selected[p.tagB])
+          .map(p => ({ name: `${sex_names[p.tagA]}/${sex_names[p.tagB]}`, tag: `${pairing_tag_name(p.tagA)}/${pairing_tag_name(p.tagB)}` }));
```
Sindrake:
```suggestion
          .map(p => ({
            name: sex_names[p.tagA] + "/" + sex_names[p.tagB],
            tag: pairing_tag_name(p.tagA) + "/" + pairing_tag_name(p.tagB),
          }));
```
Less incomprehensible.

Sindrake (dismissed):
Minor style nitpick.
The rest looks good.

DonovanDMC:
Should I apply that same change to the other templated strings like it?
https://github.com/DonovanDMC/e621ng/blob/d52662c88d1d5d12a23a7e6670818b6988a6930e/app/javascript/src/js/pages/uploads/new/uploader.vue#L450 for instance
I do agree it makes it much more readable

Sindrake:
> Should I apply that same change to the other templated strings like it? https://github.com/DonovanDMC/e621ng/blob/d52662c88d1d5d12a23a7e6670818b6988a6930e/app/javascript/src/js/pages/uploads/new/uploader.vue#L450 for instance I do agree it makes it much more readable

Would be a good idea, yeah.

Sindrake (approved):

PR #2172, opened 2026-06-10
[DbExports] Add a legacy export path

jmortiger:
> [!WARNING]
> Made w/ web editor; not tested locally

Old [`/db_export`](https://e621.net/db_export) redirects to new [`/db_exports`](https://e621.net/db_exports) path.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -456,15 +456,12 @@ Rails.application.routes.draw do
     end
   end
 
-  # #region Aliases
-  # #region Resource Shorthand
+  # Resource Shorthand
   resources :wpages, controller: "wiki_pages"
   resources :ftopics, controller: "forum_topics"
   resources :fposts, controller: "forum_posts"
-  # #endregion Resource Shorthand
 
-  # #region Legacy aliases
-  # #region Pre-Danbooru Paths
+  # Legacy aliases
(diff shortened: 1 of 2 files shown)


DonovanDMC:
Dumping the old exports immediately is kinda shitty, this has broken at least 4 separate projects and 2 modules I maintain with little to no warning

I don't think the redirect really adds any benefit, I think it would be infinitely better to instead show an html notice page for anything `/^\/db_export(?:\/.*)?$/` (like the old posts api pages have: https://e621.net/post/index.json)
So consumers can actively see that something has changed rather than hitting redirects/not found pages

jmortiger:
I mean, I was thinking of people linking to it from the forums. I could _probably_ make it a more graceful transition, but this was a quick fix for the problem I had in mind at the time.

DonovanDMC:
> I mean, I was thinking of people linking to it from the forums. I could _probably_ make it a more graceful transition, but this was a quick fix for the problem I had in mind at the time.

[It isn't really mentioned that often](https://e621.net/forum_posts?search%5Bbody_matches%5D=db_export), [and it being a not found has been brought up](https://e621.net/forum_topics/64650)

I think a redirect (which most libraries would probably follow) resulting in a page of garbled data to pull from would be worse than an explanation page and status code to match, plus this only covers the root directory for viewing
Something should also be done about the actual subpaths so they aren't just returning plain 404 errors

Sindrake (approved):

PR #2180, opened 2026-06-12
[Replacements] Add config value for penalize-by-default

DRLaDRLa:


Sindrake:
No, we do not need this.

PR #2181, opened 2026-06-13
[DB Export] Additions & Cleanups

DonovanDMC:
While going over the exports I notice a lot of things that I think should be present, in addition to some nice to have things.
I'm unsure if any indexes should be needed for the joins, but I believe it should be performant enough.

- Add populate scripts for AIBURs & forum post votes
- Rename `artists.urls` to `active_urls`
- coalesce `artist.active_urls` to an empty string to prevent outputting nulls
- Format SQL scripts for easy reading
- Move SQL scripts out of code and into individual files

| Export | Added |
| :----: | :---: |
| `artists` | `inactive_urls` |
| `bulk_update_requests` | `down_votes`, `meh_votes`, `up_votes` |
| `pools` | none |
| `post_replacements` | none |
| `post_versions` | none |
| `posts` | `bg_color`. `last_noted_at`, and `last_commented_at` |
| `tag_aliases` | `forum_post_id`, `forum_topic_id`, `reason`, `updated_at`, `approver_id`, `post_count`, `down_votes`, `meh_votes`, `up_votes` |
| `tag_implications` | `forum_post_id`, `forum_topic_id`, `reason`, `updated_at`, `approver_id`, `descendant_names`, `down_votes`, `meh_votes`, `up_votes` |
| `tags` | `created_at`, `updated_at`, `is_locked` |
| `wiki_pages` | `parent` |

In addition, what are the considerations for extending this to more export types in the future?
If we were to go full in on including everything that might be used, I think that would be these:
- `artist_versions`
- `mod_actions`
- `notes`
- `note_versions`
- `pool_versions`
- `post_events`
- `tag_type_versions`
- `wiki_page_versions`

If we're going bare usability, I think `mod_actions` and `post_events` would be the most useful (despite having to neuter mod_actions heavily)

([within my own codebase I also include users, but I'm not sure if we'd really want to do that](https://github.com/GayFurCity/GayFurCity/blob/b9cc5d5d64f83c8a6e1d32526ca6ed7604a25bf2/db/export/users.sql))


DonovanDMC:
@Sindrake is the 😕 because of the vote counts? I only tested it against 100 of each AIBUR at creation time so I haven't put it into a full scale test, I'll do that later today to see what the performance is like (from the small test it didn't seem like it would be too bad), though I'll definitely have to tweak the populate scripts I made since they can only make a few thousand unique word tags, I'll just make it fully random characters 

And if it's about moving the sql scripts into files, they should be read only once when the class is first loaded, and I left the previous structure mostly in tact in case some future export ends up needing actual ruby code and not just sql

db/populate.rb
```diff
@@ -12,13 +12,18 @@
...
   postvotes: ENV.fetch("POSTVOTES", 0).to_i,
   commentvotes: ENV.fetch("COMVOTES", 0).to_i,
+  forumvotes: ENV.fetch("FORUMVOTES", 0).to_i,
   pools: ENV.fetch("POOLS", 0).to_i,
   furids: ENV.fetch("FURIDS", 0).to_i,
   dmails: ENV.fetch("DM", 0).to_i,
   trends: ENV.fetch("TRENDS", 0).to_i,
   trends_hours: ENV.fetch("TRENDS_HOURS", 0).to_i,
+  aliases: ENV.fetch("ALIASES", 0).to_i,
+  implications: ENV.fetch("ALIASES", 0).to_i,
```
Sindrake:
```suggestion
  implications: ENV.fetch("IMPLICATIONS", 0).to_i,
```

db/populate.rb
```diff
@@ -27,29 +32,40 @@
...
-FURIDS    = presets[:furids]
-DMAILS    = presets[:dmails]
-TRENDS    = presets[:trends]
+USERS        = presets[:users]
+POSTS        = presets[:posts]
+COMMENTS     = presets[:comments]
+FAVORITES    = presets[:favorites]
+FORUMS       = presets[:forums]
+TOPICS       = presets[:topics]
+TOPICS       = 1 if TOPICS == 0 && FORUMS >= 1
```
Sindrake:
```
/app/db/populate.rb:57: warning: already initialized constant TOPICS
/app/db/populate.rb:56: warning: previous definition of TOPICS was here
```

This whole thing is pointless anyways.
`TOPICS` is assigned but never used.

db/exports/tag_aliases.sql
```diff
@@ -0,0 +1,25 @@
...
+  COALESCE(votes.up, 0)   AS up_votes
+FROM tag_aliases
+LEFT JOIN LATERAL (
+  SELECT
+      COUNT(*) FILTER (WHERE forum_post_votes.score = -1) AS down,
+      COUNT(*) FILTER (WHERE forum_post_votes.score = 0)  AS meh,
+      COUNT(*) FILTER (WHERE forum_post_votes.score = 1)  AS up
+  FROM forum_post_votes
+  WHERE forum_post_votes.forum_post_id = tag_aliases.forum_post_id
+) votes ON true
```
Sindrake: Extremely questionable.
This is adding ~83k extra database queries to the export job for aliases alone, then another ~74k for implications and ~14k for BURs.

I am extremely concerned about the `DbExportJob`'s performance.
It already takes over an hour to complete.

I would suggest looking into whether a pre-aggregated form would be better.

```suggestion
FROM tag_aliases
LEFT JOIN (
  SELECT
    forum_post_id,
    COUNT(*) FILTER (WHERE score = -1) AS down,
    COUNT(*) FILTER (WHERE score = 0)  AS meh,
    COUNT(*) FILTER (WHERE score = 1)  AS up
  FROM forum_post_votes
  GROUP BY forum_post_id
) votes ON votes.forum_post_id = tag_aliases.forum_post_id
```
Not tested.

Sindrake (changes requested):
The DB export portion of this PR looks pretty solid.
Extracting SQL queries into files is a good idea, and a clean improvement.
Including AIBUR votes is somewhat questionable in its current implementation.

However, this PR seems to mix two distinct concerns.
The `populate.rb` is already massive, messy, and buggy. Realistically, it should be split into multiple smaller module files. It definitely should not be in this PR, though.

db/populate.rb
DonovanDMC: That started out used but I removed it eventually, it isn't even present in my local codebase so I'm not sure why it's here

db/exports/tag_aliases.sql
DonovanDMC: We could also potentially add count columns to each of the AIBURs so we don't need to have any extra queries in the export at all
I don't think we could use a counter_cache since it would have to be on ForumPost for that, and votes have no reference to what AIBUR they're for

DonovanDMC:
> The `populate.rb` is already massive, messy, and buggy. Realistically, it should be split into multiple smaller module files. It definitely should not be in this PR, though.

I had actually thought about splitting it out but figured it was well out of scope here, it's painful to find the code I'm looking for with how large the file is

DonovanDMC:
I'm running a script overnight to generate a bunch of data for benchmarking
- 500,000 users
- 100,000 aliases
- 100,000 implications
- 25,000 BURs
- 2,000,000 votes

I also might look into creating a vote count column on each, but I'm not going to put too much effort into that without it being a desired change

If you could give me the actual numbers of votes we have (and their distribution among AIBURs), I could make this much more accurate

DonovanDMC:
With the benchmark script added, the initial output was this:
```
Export                   Planning    Execution   Nodes  Subq Runs   Total Qs
artists                     1.41ms        0.26ms       6          1          2
bulk_update_requests        1.18ms      543.34ms       5      25102      25103
pools                       0.98ms        0.04ms       2          0          1
post_replacements           0.89ms        0.04ms       2          0          1
post_versions               0.79ms        0.08ms       2          0          1
posts                       1.99ms        0.08ms       2          0          1
tag_aliases                 1.05ms     1542.07ms       5     100002     100003
tag_implications            0.66ms     1399.06ms       5     100100     100101
tags                        0.54ms      132.54ms       1          0          1
wiki_pages                  0.72ms        0.04ms       2          0          1
```

And after the sql change, this:
```
Export                   Planning    Execution   Nodes  Subq Runs   Total Qs
artists                     1.35ms        0.25ms       6          1          2
bulk_update_requests        1.52ms      533.26ms       5          1          2
pools                       0.76ms        0.03ms       2          0          1
post_replacements           0.74ms        0.04ms       2          0          1
post_versions               0.60ms        0.04ms       2          0          1
posts                       1.77ms        0.08ms       2          0          1
tag_aliases                 0.89ms     1582.73ms       6          3          4
tag_implications            0.78ms     1605.82ms       6          2          3
tags                        0.64ms      131.79ms       1          0          1
wiki_pages                  0.81ms        0.05ms       2          0          1
```

Little to no difference in timing, but immense difference in subqueries
These were both ran with 100,002 aliases (888,107 votes), 100,100 implications (889,098 votes), and 25,102 BURs (222,795 votes)

Sindrake (changes requested):
Very good, thank you.

Please, remove the changes to `populate.rb` out of this PR.
They do not belong in "DBExport Additions and Cleanups".
A separate PR may be more appropriate.

The benchmark script should also not be committed to the repo.

Sindrake (approved):

PR #2182, opened 2026-06-14
[Appeals] Misc cleanup

DRLaDRLa:
- Move flag specific stuff out of the controller and into Flag appeal type
- Simplify appeal UI to focus on flags since that's the only type that can be appealed. This can easily be reverted later.
- Status search UI bugfix

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -30,10 +30,7 @@ class AppealsController < ApplicationController
     end
 
     unless @appeal.can_create_for?(CurrentUser.user)
-      redirect_path = @appeal.content_path
-      if redirect_path.blank?
-        redirect_path = appeals_path
-      end
+      redirect_path = @appeal.content_path || appeals_path
       redirect_to redirect_path, alert: @appeal.messages[:cannot_create]
       return
     end
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -24,22 +24,17 @@ class AppealsController < ApplicationController
   def new
     @appeal = Appeal.new(qtype: params[:qtype], disp_id: params[:disp_id])
 
-    # These messages will have to be moved to AppealTypes if anything else than deletion becomes appealable
     if (existing_appeal = @appeal.find_duplicate_for(CurrentUser.user)).present?
-      redirect_to appeal_path(existing_appeal), alert: "This deletion has already been appealed."
+      redirect_to appeal_path(existing_appeal), alert: @appeal.messages[:duplicate]
       return
     end
 
     unless @appeal.can_create_for?(CurrentUser.user)
-      redirect_path = appeals_path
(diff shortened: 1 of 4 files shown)


app/views/appeals/index.html.erb
```diff
@@ -8,11 +8,10 @@
         <tr>
           <th style="width: 0;"></th>
           <% if CurrentUser.is_staff? %>
-            <th style="width:10%">Reporter</th>
-            <th style="width:10%">Accused</th>
+            <th style="width:10%">Appealer</th>
+            <th style="width:10%">Deleter</th>
```
Sindrake: This will make 0 sense when/if we expand this system to include stuff other than deletions.

DRLaDRLa: I can look for a more generic term, but I was thinking it could just be changed if necessary.

DRLaDRLa: Some potential terms I found
- Origin / Originator
- Source
- Initiator
- Doer
- Agent

I'm partial to "Initiator".

app/views/appeals/_search.html.erb
```diff
@@ -1,19 +1,16 @@
...
-    <%= f.user :creator, label: "Reporter" %>
-    <%= f.user :accused %>
+    <%= f.user :creator, label: "Appealer" %>
+    <%= f.user :accused, label: "Initiator" %>
     <%= f.user :claimant %>
     <%= f.input :reason %>
   <% end %>
-  <%= f.input :qtype, label: "Type", collection: [
-      ["Post Flag appeal", "flag"],
-  ], include_blank: true %>
```
Sindrake, line 10: Put it back, comment it out if you absolutely must.

```Ruby
  <%#= f.input :qtype, label: "Type", collection: [
      ["Post Flag appeal", "flag"],
  ], include_blank: true %>
```

app/controllers/appeals_controller.rb
```diff
@@ -24,22 +24,17 @@ def show
...
-      # For a post flag, send the user back to the post: either it's undeleted, it's been re-deleted
-      # (so they need to use the new flag's appeal button), or it's something they just can't appeal.
-      # Anyway it's more helpful for the user to see the current post status than the appeals index.
-      content = @appeal.content
-      if content.is_a?(::PostFlag) && (post = content.post.presence)
-        redirect_path = post_path(post)
+      redirect_path = @appeal.content_path
+      if redirect_path.blank?
+        redirect_path = appeals_path
       end
```
Sindrake:
```suggestion
      redirect_path = @appeal.content_path || appeals_path
```
❓

Sindrake (changes requested):

app/views/appeals/index.html.erb
Sindrake: I would go with something like "Accused".
But it does not matter a whole lot.

Sindrake:
Looks good, aside from the nitpicks.

Sindrake (approved):

PR #2202, opened 2026-06-18
[Posts] Unified spacing above search field

DoubleThePsycho:
The margin and padding above the search bar was inconsistent between Default and Fullscreen view. This has been unified.

The alternatives were keeping or removing the spacing for both layouts, I found it prettier to keep the spacing.
output

### DIFF
--- a/app/javascript/src/styles/views/posts/index/_index.desktop.scss
+++ b/app/javascript/src/styles/views/posts/index/_index.desktop.scss
@@ -9,8 +9,7 @@
 
   & > .search {
     box-shadow: inset -0.25rem 0px 0.25rem -0.25rem themed("color-background");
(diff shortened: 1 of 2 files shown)


Sindrake (changes requested):
I would recommend removing the padding instead.
This way, the index page will be consistent with the show page.

With your current approach, the position of the searchbar changes between these two pages.

DoubleThePsycho:
> I would recommend removing the padding instead. This way, the index page will be consistent with the show page.
> 
> With your current approach, the position of the searchbar changes between these two pages.

I forgot to take into consideration how deeply the pages are separated.

This has been resolved.

Sindrake (approved):
lgtm

PR #2205, opened 2026-06-18
[Posts] Add relationship selectors to the advanced search menu

DoubleThePsycho:
Added "Post relationships" with tri-state toggles for "Has parent" and "Has child".

Semi-related, the arrow characters were replaced with actual SVGs.
> (I know you said to avoid `SVGIcon.ts`, but `icon_helper.tb` didn't wanna cooperate on this..)

TS/JS is generated like previously, but is now written to the appropriate file, and is not a 500 line blob.


app/javascript/src/js/pages/posts/SearchQuery.ts
```diff
@@ -230,6 +250,18 @@ export default class SearchQuery {
     return { value, negated: match[1] === "-" };
   }
 
+  private static parseIschildToken (text: string): string | null {
+    const match = text.match(/^ischild:(true|false)$/i);
```
Sindrake: Does not recognize the `hasparent` metatag.

app/javascript/src/js/pages/posts/SearchQuery.ts
```diff
@@ -230,6 +250,18 @@ export default class SearchQuery {
...
+    if (!match) return null;
+    return match[1].toLowerCase();
+  }
+
+  private static parseIsparentToken (text: string): string | null {
+    const match = text.match(/^isparent:(true|false)$/i);
```
Sindrake: Does not recognize the `haschild` and `haschildren` metatags.

app/javascript/src/js/utility/SVGIcon.ts
```diff
@@ -27,6 +27,8 @@ export default class SVGIcon {
 
     // ThumbnailEngine
     "score": '<circle cx="12" cy="12" r="3"></circle><line x1="3" x2="9" y1="12" y2="12"></line><line x1="15" x2="21" y1="12" y2="12"></line>',
+    "arrow_up": '<path fill="currentColor" stroke="none" transform="translate(5.6 3.5) scale(0.0333)" d="M169.4 502.6c12.5 12.5 32.8 12.5 45.3 0l160-160c12.5-12.5 12.5-32.8 0-45.3s-32.8-12.5-45.3 0L224 402.7 224 32c0-17.7-14.3-32-32-32s-32 14.3-32 32l0 370.7-105.4-105.4c-12.5-12.5-32.8-12.5-45.3 0s-12.5 32.8 0 45.3l160 160z"/>',
+    "arrow_down": '<path fill="currentColor" stroke="none" transform="translate(5.6 3.5) scale(0.0333)" d="M214.6 9.4c-12.5-12.5-32.8-12.5-45.3 0l-160 160c-12.5 12.5-12.5 32.8 0 45.3s32.8 12.5 45.3 0L160 109.3 160 480c0 17.7 14.3 32 32 32s32-14.3 32-32l0-370.7 105.4 105.4c12.5 12.5 32.8 12.5 45.3 0s12.5-32.8 0-45.3l-160-160z"/>',
```
Sindrake: Not only are these icons unnecessary (existing icons were fine).
Not only does this go against the principle of SVGIcon.ts (minimize the number of icons used).
Not only are these FontAwesome icons and not Lucide (looks inconsistent).

... but worst of all, they are in the wrong section (ThumbnailEngine instead of SearchQuery).

Please, either remove them or provide justification for their usage.

app/javascript/src/js/pages/posts/SearchFilters.ts
```diff
@@ -7,6 +7,14 @@ const INPOOL_STATES = ["unset", "yes", "no"];
 const INPOOL_VALUES: Record<string, string> = { unset: "", yes: "true", no: "false" };
 const INPOOL_STATE_MAP: Record<string, string> = { "": "unset", "true": "yes", "false": "no" };
 
+const ISCHILD_STATES = ["unset", "yes", "no"];
+const ISCHILD_VALUES: Record<string, string> = { unset: "", yes: "true", no: "false" };
+const ISCHILD_STATE_MAP: Record<string, string> = { "": "unset", "true": "yes", "false": "no" };
+
+const ISPARENT_STATES = ["unset", "yes", "no"];
+const ISPARENT_VALUES: Record<string, string> = { unset: "", yes: "true", no: "false" };
+const ISPARENT_STATE_MAP: Record<string, string> = { "": "unset", "true": "yes", "false": "no" };
```
Sindrake: Do we need separate constants for `ischild` and `isparent` toggles?
It's just used for validation. Could rename the `INPOOL` constants to something generic, and reuse them for all three cases.

Sindrake (changes requested):

app/javascript/src/js/pages/posts/SearchQuery.ts
DoubleThePsycho: This along with the `hasparent` issue was an oversight, on account of them - seemingly - not being listed on the cheatsheet.

This has been fixed. All 3 metatags have been tested and appear to work just fine.

app/javascript/src/js/pages/posts/SearchFilters.ts
DoubleThePsycho: These have been merged into `TOGGLE_STATES`, `TOGGLE_VALUES` and `TOGGLE_STATE_MAP`.

app/javascript/src/js/utility/SVGIcon.ts
DoubleThePsycho: > Not only are these icons unnecessary (existing icons were fine).

They are, I just thought they were a bit thin and not vertically centered within the button.

> Not only does this go against the principle of SVGIcon.ts (minimize the number of icons used).

For whatever reason I had some difficulties loading the icons using `IconHelper`.
Possibly related, but this was not long before I had to rebuild the container *again*.

> Not only are these FontAwesome icons and not Lucide (looks inconsistent).

Guess which idiot saw the `lucide.dev` comment and decided it was not worth checking out?
In my defense, I have not heard of Lucide before which is why I used FontAwesome.

> ... but worst of all, they are in the wrong section (ThumbnailEngine instead of SearchQuery).

Yup, this is my OCD saying "arrows should go with other arrows", rather than placing them in/near SearchQuery.

---

My justification:
The pure-text icons looked a bit thin and were awkward to vertically align within the button.

I can replace them with Lucide icons and try again with IconHelper, or I can undo this change and we can forget it ever happened. Your call.

Sindrake: One potential justification for using icons over ASCII characters is that the arrows in question are not a part of the standard character set. The way they are rendered (and whether they are rendered at all) may depend on the device.

Including Lucide icons within SVGIcon.ts is more justifiable, since they are relatively tiny.

DoubleThePsycho: No kidding, the Lucide icons were **must** tinier.

While I am at it, I've noticed the buttons in `.search-controls` are shorter on `posts#show` than `posts#index`.
May I tackle that real quick in this PR, or do you want a separate one for that?

Sindrake: The button sizes have already been fixed in #2201. By you.

DoubleThePsycho: Huh... I remember fixing the overlap, not the button height...

Updating my branch shows the fix though, so all is good.

> Except my memory, I should probably see a doctor at this point.

If switching to Lucide icons is a sufficient fix for you, I don't think I have anything to add then.

Sindrake (approved):

PR #2208, opened 2026-06-19
[Posts] Fix some sample image colour bugs

DRLaDRLa:
Main issue is that colour spaces weren't being converted correctly, so HDR images would end up washed out in samples. Explicitly converting using the `perceptual` intent fixes this.

Also move the background colour used to replace alpha to the config, for forks that use a different default theme.

### DIFF
--- a/app/logical/image_sampler.rb
+++ b/app/logical/image_sampler.rb
@@ -228,7 +228,11 @@ module ImageSampler
 
     # Convert embedded colour spaces to sRGB for compatibility.
     if result.get_typeof("icc-profile-data") != 0
-      result = result.icc_transform("srgb", intent: :perceptual)
+      begin
+        result = result.icc_transform("srgb", intent: :perceptual)
+      rescue Vips::Error => e
+        Rails.logger.warn("[ImageSampler] icc_transform failed, stripping profile: #{e.message}")
+      end
     end
 
     # save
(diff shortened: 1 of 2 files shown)


app/logical/image_sampler.rb
```diff
@@ -226,6 +226,11 @@ def gen_target_image(image, original_dims, target_size, crop: false, background:
       result = result.smartcrop(crop_area[0], crop_area[1], interesting: :entropy)
     end
 
+    # Convert embedded colour spaces to sRGB for compatibility.
+    if result.get_typeof("icc-profile-data") != 0
+      result = result.icc_transform("srgb", intent: :perceptual)
```
Sindrake: `icc_transform` raises `Vips::Error` on corrupt or mismatched embedded profile.
(e.g. broken profile data, or a profile whose colorspace doesn't match the image's bands)

Before this PR, images like that were accepted since profile was simply stripped - now they will raise.
I would rescue `Vips::Error` here, and fall back to the old behavior.


```suggestion
      begin
        result = result.icc_transform("srgb", intent: :perceptual)
      rescue Vips::Error => e
        Rails.logger.warn("[ImageSampler] icc_transform failed, stripping profile: #{e.message}")
      nd
```

Sindrake (changes requested):

Sindrake (approved):

PR #2209, opened 2026-06-20
[PasswordReset] Check that user ID is a valid int for PostgreSQL

### LINKED ISSUE #1729: Password reset controller throws ActiveRecord::RangeError if UID is too large
**Describe the bug**
The `PasswordResetsController` throws an unhandled exception when receiving an excessively long user ID (e.g., a value far exceeding a 32-bit integer). The method `find_nonce_from_params` does contain some sanitization, but it does not prevent IDs that are out of range for the integer type used by the database.

Example log:
```
ActiveRecord::RangeError
PG::NumericValueOutOfRange: ERROR:  value "212488730429208853042920885" is out of range for type integer
CONTEXT:  unnamed portal parameter $1 = '...'

app/controllers/maintenance/user/password_resets_controller.rb:52:in `find_nonce_from_params'
app/controllers/maintenance/user/password_resets_controller.rb:11:in `edit'
lib/middleware/parameter_sanitizer.rb:27:in `call'
```

**To Reproduce**
1. Submit a password reset with an extremely large or non-numeric UID value.
2. The controller processes the input and attempts to use it as an integer.
3. Error occurs when the value is too large for PostgreSQL's integer type.

**Expected behavior**
The controller should properly validate and reject user IDs that are out of the acceptable range (e.g., not a valid integer ID for `user_id`) before making a database query. It should return a reasonable error notice rather than raising an unhandled exception.

**Additional context**
- The method `find_nonce_from_params` should have explicit checks for integer range before query.
- Consider returning a generic error if the UID is plainly invalid or excessively large.
- This may have security implications if untrusted parameters are not sufficiently checked.

mint-dog:
Check that user ID is a valid int for PostgreSQL

Closes #1729

### DIFF
--- a/app/controllers/maintenance/user/password_resets_controller.rb
+++ b/app/controllers/maintenance/user/password_resets_controller.rb
@@ -39,16 +39,14 @@ module Maintenance
 
       private
 
-      USER_ID_MIN = 0
-      USER_ID_MAX = 2_147_483_647 # PostgreSQL max int
-
       def find_nonce_from_params
         # Some translator software keeps appending extra text to the end of the UID.
         # Just strip the non-numeric characters, ex: "1739850关闭网页" -> "1739850"
         sanitized_uid = params[:uid].to_s.gsub(/[^\d]/, "")
-        return nil if sanitized_uid.blank? || !sanitized_uid.to_i.between?(USER_ID_MIN, USER_ID_MAX)
+        sanitized_uid = ParseValue.safe_id(sanitized_uid)
+        return nil if sanitized_uid < 1
 
-        if params[:uid].to_s != sanitized_uid
+        if params[:uid].to_s != sanitized_uid.to_s
           Rails.logger.warn("Password reset UID sanitized: original='#{params[:uid]}', sanitized='#{sanitized_uid}', key='#{params[:key]}', user_agent='#{request.user_agent}'")
         end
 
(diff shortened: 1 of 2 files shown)


app/controllers/maintenance/user/password_resets_controller.rb
```diff
@@ -39,11 +39,14 @@ def update
...
 
+      USER_ID_MIN = 0
+      USER_ID_MAX = 2_147_483_647 # PostgreSQL max int
+
       def find_nonce_from_params
         # Some translator software keeps appending extra text to the end of the UID.
         # Just strip the non-numeric characters, ex: "1739850关闭网页" -> "1739850"
         sanitized_uid = params[:uid].to_s.gsub(/[^\d]/, "")
-        return nil if sanitized_uid.blank?
+        return nil if sanitized_uid.blank? || !sanitized_uid.to_i.between?(USER_ID_MIN, USER_ID_MAX)
```
Sindrake: This is functionally correct, but I would recommend using `ParseValue.safe_id` instead of reinventing the wheel.

Additionally, `USER_ID_MIN` is already dead code in your implementation.
The sanitization on line 48 strips the `-` from any negative numbers already.

Sindrake (changes requested):

Sindrake (approved):

PR #2230, opened 2026-06-27
[Posts] Rework the post info bar on /posts and /favorites pages

DoubleThePsycho:
On the `/favorites` page, the `.search-controls` panel is removed to make room for the “Searching favorites does not preserve the order in which they were added.” warning.

That works on desktop, but on mobile the header still keeps the `margin-right: 5rem;` spacing used when the search controls are present. Since `/favorites` does not show those controls, the `[search help]` link ends up sitting too far left.

This quick fix removes that extra spacing only for the favorites page.

image

Personally, I think the cleaner long-term layout might be to keep `.search-controls` on the favorites page and move the warning message to where search stats normally appear. For now, I went with the smallest page-specific fix. Let me know if this approach works, or if you’d rather I try the layout adjustment instead.


Sindrake:
I think I would prefer adding the same search controls to the favorites page.

DoubleThePsycho:
> I think I would prefer adding the same search controls to the favorites page.

Honestly? Me too.

But this would have to be done strategically, because this doesn't look too great...
image

I suggest we move the warning label to the area above the post grid like this. Your thoughts?
image

I'll workshop something a bit better while I await your response.

---

### Edit:

I made some changes, adding the buttons on the search page and moved the warning text.

In hindsight, I'm unsure about my decision of altering the position and alignment of the warning text depending on presence of the "Nobody here but us chickens!" text. If this design is preferred, I'll unify it in the morning.

#### Desktop:
image

#### Mobile:
image

Sindrake (changes requested):
This leaves a lot to be desired, even at a glance.

- `.search-controls` buttons on the favorites page do not work.
- The colored `.posts-info-bar` section is intrusive and breaks the visual flow of the page - the eye "stumbles" upon it. Would be better if it was transparent.
- "Searching favorites doesn't preserve added order" is also intrusive and will annoy people. It's also more related to the search functionality than the results, so its location does not make sense.
- In fullscreen mode, `.search-controls` buttons are layered on top of the result counter.

On the upside, being able to display a "warning" message on the posts page would be useful in the future.

Sindrake (approved):

PR #2239, opened 2026-06-30
[Favorites] Increase performance of FlushFavorites

Tarrgon:
Switches flush favorites to use opensearch bulk index updating to greatly increase the performance of flushing user favorites.


Tarrgon:
I've removed the testing spec in the flush favorites job spec that utilize `FavoritesManager` as it is no longer used at all in flushing favorites. Not sure if this spec should somehow be replaced, but the current spec seems to be accurate and tests the functionality of removing favorites.

Tarrgon:
I'm going to hold off on merging until this is ready to prevent unnecessary merge commits

app/jobs/flush_favorites_job.rb
```diff
@@ -8,16 +8,21 @@ class FlushFavoritesJob < ApplicationJob
...
-      Favorite.for_user(user.id).includes(:post).find_each do |fav|
-        tries = 5
-        begin
-          FavoriteManager.remove!(user: user, post: fav.post)
-        rescue ActiveRecord::SerializationFailure
-          tries -= 1
-          retry if tries > 0
+      Favorite.for_user(user.id).select(:id, :post_id).find_in_batches(batch_size: 10_000) do |batch|
+        ids = batch.pluck(:post_id)
+        Favorite.for_user(user.id).where(post_id: ids).delete_all
```
Sindrake:
```suggestion
        Favorite.where(id: batch.map(&:id)).delete_all
```
This seems more direct, and does not require re-scanning by `post_id` with a fresh `for_user` relation.
Relatively minor.

spec/jobs/flush_favorites_job_spec.rb
Sindrake, line 1: I feel like the new behavior warrants new tests.
At least, check whether the BulkIndexUpdateJob has been enqueued.

```Ruby
expect { perform }.to have_enqueued_job(BulkIndexUpdateJob)
```

Sindrake (changes requested):
Looks good overall, with a few minor changes.

app/jobs/flush_favorites_job.rb
Tarrgon: ~~This is required because if it's not present, it will delete all favorites on that post.~~
image


Oh, wait. I used the wrong ID. Nevermind.

spec/jobs/flush_favorites_job_spec.rb
Tarrgon: Added on all test cases, as well as not expecting it when the user has no favorites.

Sindrake (approved):

PR #2241, opened 2026-06-30
[TagQuery] Alias playtime duration tags to equivalent `duration` values

jmortiger:
Added METATAG_ALIASES for long and short playtime tags.

- [x] Base implementation
   - [ ] Add unit tests
- [ ] Account for in [`Post.add_automatic_tags`](https://github.com/e621ng/e621ng/blob/feat/playtime-tags-to-duration/app/models/post.rb#L837).
   - [ ] Add unit tests


Sindrake:
This is a reasonable approach.
However, I feel like this would remove people's ability to blacklist posts by duration, since the current blacklisting system does not support that metatag at all. Implementing that should probably come before retiring these two tags.

PR #2253, opened 2026-07-01
[OAuth] Add enabled flag

clragon:
Add enable oauth flag, off by default to prevent dev issues.

### DIFF
--- a/app/logical/session_loader.rb
+++ b/app/logical/session_loader.rb
@@ -23,7 +23,7 @@ class SessionLoader
     CurrentUser.user = User.anonymous
     CurrentUser.ip_addr = request.remote_ip
 
-    if has_bearer_token?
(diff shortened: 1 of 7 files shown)


Sindrake (changes requested):
Looks good.
Two things to note.

First, bearer-token API auth is not gated by this flag.
While there isn't a way to mint tokens, pre-existing tokens will still authenticate.
Please, confirm if this is the intended behavior.

Second, I would prefer there to be at least cursory tests - make sure that `/oauth/applications/` returns 404 when disabled, for example.

clragon:
Testing whether this 404s is a bit difficult because the env is written at the beginning of a test. I elected not to test that.

Sindrake (approved):

PR #2261, opened 2026-07-06
[OAuth] Fix consent CSP form-action

clragon:
The form post redcirect works fine on Firefox. Not on chrome, which implements a very annoying policy. We precisely allow CSP for the hosts we are about to redirect to, to make this work.

### DIFF
--- a/config/initializers/doorkeeper_authorization_csp.rb
+++ b/config/initializers/doorkeeper_authorization_csp.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+# The consent form POSTs to /oauth/authorize then 302s to the app's redirect_uri,
+# and browsers enforce form-action across that redirect chain, so the consent page
+# must allow that redirect origin the global CSP would otherwise block. Only added once the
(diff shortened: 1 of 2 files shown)


config/initializers/doorkeeper_authorization_csp.rb
```diff
@@ -0,0 +1,27 @@
...
+  rescue URI::InvalidURIError
+    nil
+  end
+end
+
+Rails.application.config.to_prepare do
+  Doorkeeper::AuthorizationsController.class_eval do
+    content_security_policy do |policy|
+      origin = DoorkeeperAuthorizationCsp.redirect_origin(pre_auth.redirect_uri)
+      policy.form_action(:self, origin) if origin
```
Sindrake: If I am reading this correctly, wouldn't the computed origin be appended to the CSP directives even on non-authorizable/error path?
No form is rendered in that case, of course, but I'm still not sure if that's right.

Please, verify whether this is the correct approach.

Sindrake (changes requested):

Sindrake (approved):

PR #2276, opened 2026-07-08
[Appeals] Don't use a default status for pending appeals

DRLaDRLa:
Too easy to forget to pick the right status before submitting. Defaults to blank in the select, and the submit button is disabled until you pick a valid status.

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -134,9 +134,7 @@ class AppealsController < ApplicationController
   end
 
   def update_appeal_params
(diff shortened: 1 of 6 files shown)


Sindrake (changes requested):
Looks good.

One minor nitpick – the validation here is purely client-side, button being disabled is purely cosmetic. If an empty status reaches the controller, it would raise 500 `ActiveRecord::NotNullViolation` rather than return a friendly validation error.

Unlikely to happen in a real-life scenario, but I feel like it's still worth guarding against.

Sindrake (approved):
Looks good to me, thank you!

PR #2277, opened 2026-07-09
[PostVotes] Fix vote trends stylesheet error

mm12:
The HTML was updated to change `admin` to staff, but the pathing of (and the SCSS itself) were not updated. As a result, the styles did not render correctly. 

Current: 
image

Corrected:
image


Sindrake (changes requested):
`/views/staff/vote_trends/index.html.erb` still uses the `c-admin-vote-trends` ID.
Does not affect anything, since the styles hang off the body class. But it's still worth fixing.

Sindrake (approved):

Sindrake (changes requested):
Merge conflicts after PR #2278 was merged.
Sorry.

Sindrake (approved):

PR #2278, opened 2026-07-09
[PostVotes] Vote trends access for all staff

mm12:
Changes `admin` access to `staff` access. Note that while the `/post_votes` subnav buttons are perm'ed for all staff, the page itself is currently mod+, so the buttons do not render for janitors or base staff. Also note that as of current, the page is manually excluded from the sitemap. The only way for staff under `moderator` to access it is with the direct url, `/staff/vote_trends`.


app/views/post_votes/_secondary_links.html.erb
```diff
@@ -5,6 +5,8 @@
   <% if CurrentUser.is_admin? %>
     <% user_filter = params.dig(:search, :user_name) %>
     <% user_filter ||= "!#{params.dig(:search, :user_id)}" if params.dig(:search, :user_id).present? %>
+  <% end %>
+  <% if CurrentUser.is_staff? %>
     <%= subnav_link_to "User Vote Trends", staff_vote_trends_path(user: user_filter) %>
   <% end %>
```
Sindrake: `user_filter` is defined within the `is_admin?` conditional, but used within `is_staff?`.
I'm not sure what this was meant to accomplish, but it looks wrong.

spec/requests/staff/vote_trends_controller_spec.rb
Sindrake, line 1: This is a pre-existing bug, but worth fixing.
The spec file is still located at `/spec/requests/admin/`, while it should be at `/staff/`.

Sindrake (changes requested):

Sindrake:
Wow, that was quick.

Sindrake (approved):

PR #2294, opened 2026-07-15
[Misc] Remove empty files

mm12:
Inadvertently introduced in e621ng/e621ng#931. This is the only empty file outside of the png text fixture and dependency/lock/vcs files.


Sindrake:
That file is critically important.

Sindrake (approved):

PR #2295, opened 2026-07-15
[Users] Add PostReplacement count to stats

mm12:
* Add PostReplacement count to stats
    * Labeled as image
    * Counts `submitted` replacements (non-original)
    * Stored in UserStatus
    * Links to PostReplacement search for user's `submitted` replacements
  * Add `submitted` search to PostReplacements
    * Intercept the `status` attribute check, use `where.not(status: original)` if submitted, otherwise use the existing search method.
    * Add `submitted` option to search partial
  * Add helpers for `UserStatus` updates
    * Add a db migration to add the `post_replacement_submitted_count` column to `UserStatus`
    * Add a db fix to populate the column (mostly copied from the ticket count fix)
    * Add incrementer to the `replacer`

Notes:
  * UserStatus has no tests as of current, so none are included.
  * Behavior on destory of replacements is unaccounted for
  * There may be unrealized optimizations that can be made to the DB calls.

Feedback on UI appreciated. 

image

image

### DIFF
--- a/app/logical/upload_service/replacer.rb
+++ b/app/logical/upload_service/replacer.rb
@@ -76,6 +76,7 @@ class UploadService
           penalize_uploader_on_approve: penalize_current_uploader.to_s.truthy?,
         })
 
+        UserStatus.for_user(CurrentUser.id).update_all("post_replacement_submitted_count = post_replacement_submitted_count + 1")
         UserStatus.for_user(previous_uploader).update_all("own_post_replaced_count = own_post_replaced_count + 1")
         if penalize_current_uploader.to_s.truthy?
           UserStatus.for_user(previous_uploader).update_all("own_post_replaced_penalize_count = own_post_replaced_penalize_count + 1")
--- a/app/logical/upload_service/replacer.rb
+++ b/app/logical/upload_service/replacer.rb
@@ -76,7 +76,6 @@ class UploadService
           penalize_uploader_on_approve: penalize_current_uploader.to_s.truthy?,
         })
 
-        UserStatus.for_user(CurrentUser.id).update_all("post_replacement_submitted_count = post_replacement_submitted_count + 1")
         UserStatus.for_user(previous_uploader).update_all("own_post_replaced_count = own_post_replaced_count + 1")
         if penalize_current_uploader.to_s.truthy?
           UserStatus.for_user(previous_uploader).update_all("own_post_replaced_penalize_count = own_post_replaced_penalize_count + 1")
--- a/app/models/post_replacement.rb
+++ b/app/models/post_replacement.rb
@@ -336,7 +336,10 @@ class PostReplacement < ApplicationRecord
 
         q = q.attribute_exact_matches(:file_ext, params[:file_ext])
         q = q.attribute_exact_matches(:md5, params[:md5])
-        q = q.attribute_exact_matches(:status, params[:status])
+
+        if params[:status].present?
+          q = params[:status].to_s == ("submitted") ? q.where.not(status: "original") : q.attribute_exact_matches(:status, params[:status])
+        end
 
         q = q.where_user(:creator_id, :creator, params)
         q = q.where_user(:approver_id, :approver, params)
@@ -399,6 +402,10 @@ class PostReplacement < ApplicationRecord
 
       def for_user(id)
         where(creator_id: id.to_i)
+      end
+
+      def submitted
+        where.not(status: "original")
       end
 
       def for_uploader_on_approve(id)
--- a/app/models/post_replacement.rb
+++ b/app/models/post_replacement.rb
@@ -340,7 +340,7 @@ class PostReplacement < ApplicationRecord
         q = q.attribute_exact_matches(:md5, params[:md5])
 
         if params[:status].present?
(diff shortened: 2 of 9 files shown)


app/logical/upload_service/replacer.rb
```diff
@@ -76,6 +76,7 @@ def process!(penalize_current_uploader:)
           penalize_uploader_on_approve: penalize_current_uploader.to_s.truthy?,
         })
 
+        UserStatus.for_user(CurrentUser.id).update_all("post_replacement_submitted_count = post_replacement_submitted_count + 1")
```
Sindrake: This seems to be incrementing the current user's counter.
In this context, that would be the staff member / approver.

db/migrate/20260715181809_add_submitted_replacements_count_to_user_statuses.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddSubmittedReplacementsCountToUserStatuses < ActiveRecord::Migration[8.1]
+  def change
+    add_column :user_statuses, :post_replacement_submitted_count, :integer
```
Sindrake: This column should have a `DEFAULT 0 NOT NULL`, like the other ones.

Sindrake (changes requested):
This is a really strange counter.
It's labeled as "submitted", which reads as "all replacements this user created".
The search seems to support that.

However, the counter is only incremented on approval - so, rejections or promotions do not count. Destroying a replacement is also unaccounted for, as you pointed out.

I would increment this counter in a `after_create` callback, and decrement it in an `after_destroy`.

mm12:
> This is a really strange counter. It's labeled as "submitted", which reads as "all replacements this user created". The search seems to support that.
> 
> However, the counter is only incremented on approval - so, rejections or promotions do not count. Destroying a replacement is also unaccounted for, as you pointed out.
> 
> I would increment this counter in a `after_create` callback, and decrement it in an `after_destroy`.

I'm not totally sure why tests pass then, but yes i def has some oversights here. Will try to get to it this weekend.

db/structure.sql
```diff
@@ -6103,6 +6104,8 @@ ALTER TABLE ONLY public.oauth_access_tokens
 SET search_path TO "$user", public;
 
 INSERT INTO "schema_migrations" (version) VALUES
+('20260715181809'),
+('20260714184832'),
```
Sindrake: What is that second migration doing here?

app/models/post_replacement.rb
```diff
@@ -430,6 +439,16 @@ def set_previous_uploader
...
   end
 
+  def increment_user_replacement_count
+    return if status == "original"
+    UserStatus.for_user(CurrentUser.id).update_all("post_replacement_submitted_count = post_replacement_submitted_count + 1")
+  end
+
+  def decrement_user_replacement_count
+    return if status == "original"
+    UserStatus.for_user(CurrentUser.id).update_all("post_replacement_submitted_count = post_replacement_submitted_count - 1")
```
Sindrake, line 449: Code relocated, bug unfixed.
Still updates the counter for the **CurrentUser**, which will be the staff member handling a replacement.

app/models/post_replacement.rb
```diff
@@ -336,7 +338,10 @@ def search(params)
 
         q = q.attribute_exact_matches(:file_ext, params[:file_ext])
         q = q.attribute_exact_matches(:md5, params[:md5])
-        q = q.attribute_exact_matches(:status, params[:status])
+
+        if params[:status].present?
+          q = params[:status].to_s == ("submitted") ? q.where.not(status: "original") : q.attribute_exact_matches(:status, params[:status])
```
Sindrake:
```suggestion
          q = params[:status].to_s == ("submitted") ? q.submitted : q.attribute_exact_matches(:status, params[:status])
```

Sindrake (changes requested):
Problem unfixed.

It seems like the tests are passing because both the creator and the deleter are the same user. Test this properly: one user creates a replacement, another user approves/rejects it.

app/models/post_replacement.rb
mm12: i forgot that destroy is done by staff.

db/structure.sql
mm12: i forgor to remove it when i removed all the other junk

mm12:
> Problem unfixed.
> 
> It seems like the tests are passing because both the creator and the deleter are the same user. Test this properly: one user creates a replacement, another user approves/rejects it.

yes , i will try to see if i can update tests sometime in the next 8 days . difficult to test since no other userstatus has any test actually checking it is updated, and also because docker is being a nuisance.

C6Yfd-dfY6C:
I like the idea, although I do have a minor concern with the UI. It's labelled as Image in the profile stats there, but videos and animated GIF/APNG/WebP posts can also be replaced, making that wording possibly ever so slightly awkward. Not the end of the world obviously, but probably just a nitpick for some. You can probably get away with saying "Image" for the animated ones, but videos would probably need different wording for this for sure. I'm not entirely sure what the best route for labeling this would be then. Possibly just "Replaced" or "Replacements"? Separate it into different stat numbers for "Image" and "Video"? Those are just quick ideas, not entirely confident in them unless you're fine with those. I just rather wanted to bring that small "issue" up if anything.

PR #2304, opened 2026-07-19
[PostSets] Fix post_list json

DonovanDMC:
Currently attempting to visit `/post_sets/{id}/post_list.json` results in an error due to the template's path missing `.html`, resulting in attempting to render the `secondary_links` partial as json (none exists, thus error).

This PR just fixes that error, the route is functionally the same to `show` for JSON, should we add special functionality for JSON or just leave it as is?


DonovanDMC:
The rubocop failure was a pre-existing (previously ignored) issue, I went ahead and fixed it since it's simple

Sindrake (approved):

Issue #1553, opened 2025-11-22
safe mode shows "deleted" for the profile picture in case the post is NSFW


nilFinx:
**Describe the bug**
While browsing the comment section in safe mode (through the safe mode toggle in settings or by using e926), I see `deleted` for people with a NSFW profile picture, instead of something that actually informs users that the profile picture with NSFW.

**To Reproduce**
Either enable safe mode, or go to the SFW mirror on instances that has it set up. (e926 for the official one), and see the comment from somebody with a NSFW profile picture, and see that the profile picture is replaced with `deleted` which is not accurate

**Expected behavior**
Something that indicates that the post is NSFW, like just `NSFW`, or any other labels that may fit this

**Screenshots**

<img width="824" height="225" alt="Image" src="https://github.com/user-attachments/assets/62199b6d-ba9e-47f7-b29f-e803607154f3" />

PR #2319, opened 2026-07-25
Solution (#1553): safe mode shows "deleted" for the profile picture in case the pos

TFGSUMIT:
This pull request introduces a new `AvatarComponent` class to render profile pictures based on user status and profile picture tags. It fixes the issue with the "deleted" label being displayed for profile pictures in safe mode when the post is NSFW.

Changes:

* Introduced `AvatarComponent` class in `app/components/avatar_component.rb`
* Added `is_nsfw_profile_picture` helper function in `app/helpers/application_helper.rb`
* Modified `User` model in `app/models/user.rb` to include `profile_picture_tags` attribute
* Added tests for `AvatarComponent` in `app/components/avatar_component_spec.rb`

Testing instructions:

* Run `bundle exec rspec` to execute the tests
* Verify that the `AvatarComponent` class renders the correct profile picture based on user status and profile picture tags
* Verify that the "deleted" label is not displayed for profile pictures in safe mode when the post is NSFW

### DIFF
--- a/n
+++ b/n
@@ -1,0 +1,22 @@
+require 'spec_helper'
+
+describe AvatarComponent do
+    let(:user) { User.new(profile_picture_tags: ["nsfw"]) }
+
+    it "renders NSFW label for NSFW profile picture" do
+        component = AvatarComponent.new(user)
+        expect(component.render).to eq("NSFW")
+    end
+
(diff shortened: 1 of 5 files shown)


Sindrake:
image

What?

PR #2349, opened 2026-08-02
[Stats] Add `updated_at` field & coerce stringified floats

DonovanDMC:
* Add an `updated_at` property to the stats, returned in both the JSON and shown in the UI
* Fix some fields returning strings rather than parsed floats (the averages)
* Replace the `|| 0` instances with an explicit `.to_i` call
* Modify the `:started` key to function with 0 posts present


Sindrake (approved):
lgtm

PR #2370, opened 2026-08-06
[Wiki] Add table of contents

clragon:
Table of contents implementation for Wikis.

Tries its best to work in the way users already write wikis:
1. Headers and sections are jump points.
2. Links lose trailing parenthesis.
3. An automatic `#top`  always exists, and ▲ and friends are excluded from the links and labels.
4. Header followed by section named the same does not gain a duplicate link.
5. Weirdly nested headers (e.g. starting with h4) are normalized relative to each other.
6. Existing dtext anchors inside headers override the auto-generated one.

Uses a StringScanner instead of a html parser because Nokogiri would reformat the HTML (e.g. transform `&#38;` to `&`) and I didnt like that.

Preview:
image

Some articles might look strange after merging this, e.g. e621:cheatsheet uses h6 plus bold to emphasize the tag searches. I believe these should be edited to use just bold later.

### DIFF
--- a/app/helpers/wiki_pages_helper.rb
+++ b/app/helpers/wiki_pages_helper.rb
@@ -12,4 +12,8 @@ module WikiPagesHelper
   def wiki_page_alias_and_implication_list(wiki_page)
     render "tags/alias_and_implication_list", tag: wiki_page.tag || Tag.new(name: wiki_page.title)
   end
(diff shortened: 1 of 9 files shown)


Sindrake (approved):
Looks good to me.

PR #2414, opened 2026-08-12
[Posts] Add tumblr source sanitizer

vo-lte:
(The description for #1753 should also be edited to address the changes in #2004.)

The new tumblr source sanitizer implementation eagerly removes the `source` query parameter from tumblr.com URLs. I can't find other richer examples of tumblr URLs, but this addresses the complaint in #1753 for Tumblr URLs.


Sindrake (changes requested):
Looks good, thank you.

I would recommend either stripping the following, or justifying why they should be kept:
1. Bare trailing `?`
2. `redirect_to`
3. `is_related_post`

Here is a dataset from production: https://etc.cnd.pet/misc/tumblr-sources.log

vo-lte:
I'm fine with the state of this PR as is, though given how the twitter sanitizer works with nitter, it could be worthwhile to:

- Transform themed URLs to just tumblr URLs (`https://deriazironfist.tumblr.com/post/72456853182/sketchdaily-ducks` -> `https://www.tumblr.com/deriazironfist/72456853182/sketchdaily-ducks`)
- Drop the human-readable (and optional!) human-readable slug (`https://www.tumblr.com/deriazironfist/72456853182/sketchdaily-ducks` -> `https://www.tumblr.com/deriazironfist/72456853182`)

(I'm hesitant on doing this immediately since I'm unsure if this is too much for a sanitizer, though it seems in-scope.)

Sindrake:
Looks good to me, thank you.
I think it's fine to skip further normalization for now.

Sindrake (approved):

PR #2415, opened 2026-08-12
[TagCategories] Use a separate mapping for the internal (symbol) names

DRLaDRLa:
That way you can rename a category without needing to change all uses of the `Tag.categories` member. Will be most useful for *that fork*.

### DIFF
--- a/app/logical/tag_category.rb
+++ b/app/logical/tag_category.rb
@@ -52,6 +52,21 @@ class TagCategory
     8 => "lore",
   }.freeze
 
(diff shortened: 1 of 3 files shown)


DRLaDRLa:
I don't think the `coverage-report` failure has anything to do with my changes.

Sindrake (approved):
Added a test to ensure that REVERSE_MAPPING does not drift away from ENUM_MAPPING – for example, if a new category is added.

The rest looks good.

PR #2423, opened 2026-08-19
[DbExports] Record list of columns

DonovanDMC:
This records the list of columns used in queries along with their postgres type. It's surfaced in both the json and html, with the html having the row as toggle a table showing a list of the columns and their types.
I also expanded the checksum input so the full value is visible when possible, since there is plenty of leftover space.

### DIFF
--- a/app/blueprints/db_export_blueprint.rb
+++ b/app/blueprints/db_export_blueprint.rb
@@ -5,6 +5,7 @@ class DbExportBlueprint < Blueprinter::Base
   field :file_name
   field :file_size
   field :checksum
+  field :columns
   field :updated_at
   field :url
 end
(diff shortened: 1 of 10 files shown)


Sindrake:
What's the use case?

DonovanDMC:
Validating the known columns without needing to download the export (e.g. a background job could fail early without needing to download anything), as well as having the types explicitly known, no need to either look at the sql schema in the repo or guess based on the data

Sindrake:
I guess.

Truth be told, I don't want to keep indulging a feature that adds very little of value to the site.
It costs us bandwidth, it costs me time to keep reviewing these PRs - all for the benefit of like 200 people. And this is me being generous.

A line has to be drawn somewhere.
I'll think about it.

DonovanDMC:
It keeps those 200 people from making thousands of individual requests to keep their local copies up to date, seems well worth it to me

Sindrake:
So does banning them.

DonovanDMC:
Yes that seems very productive

Sindrake:
No.

PR #2441, opened 2026-08-23
[Misc] Hide links that use `post_tags_match` from anonymous users

DRLaDRLa:
- `Comments` on the `Artwork` section of a user profile (when linked)
- Search links on the flags page

Those fail with `found unpermitted parameter: :post_tags_match` and redirect to the login page, which is confusing to users, and they're far from essential, so just hide them. Reported by a user: https://discord.com/channels/1352777803036364892/1421568898951807068/1540644444376666194


Sindrake (approved):
lgtm!

PR #2457, opened 2026-08-26
[PostEvents] Add extra data to json & search

DonovanDMC:
Same idea as what was added to modactions, expanded to post events. Included a few more related additions.
* Added `replacement_id` to `replacement_promoted`
* * All of the other replacement post events had the replacement id in their extra data
* Added more descriptions/info to the html
* * Every event that has extra data had an html description added
* Converts the constants of ModAction into SCREAMING_SNAKE_CASE like rubocop wants


DRLaDRLa:
Since the replacement penalty is part of the
karma math, I think it makes sense to keep it public.

DonovanDMC:
> Since the replacement penalty is part of the karma math, I think it makes sense to keep it public.

It isn't shown on the actual replacement, so if this is kept it's just incredibly obscure, nobody but power users even knows post events exist

DonovanDMC:
With the addition of <https://e621.net/upload_karma_events?commit=Search&search%5Breason%5D=replacement_penalty>, I'd call the replacements ui the outlier
I was going to leave the scaffolding in place, but there's really no point

DRLaDRLa:
> > Since the replacement penalty is part of the karma math, I think it makes sense to keep it public.
> 
> It isn't shown on the actual replacement, so if this is kept it's just incredibly obscure, nobody but power users even knows post events exist

I'd argue the replacement penalty state should instead be made visible to users since it affects the upload limit and karma stats. I don't really see how it could be harmful, except maybe for annoying users begging for it to be removed. But anyway that's out of scope for this PR.

EDIT: You can just search replacements by penalty... https://e621.net/post_replacements?commit=Search&search%5Bpenalized%5D=true

Sindrake:
I haven't reviewed the PR yet, but I'm generally in agreement with DRLa.
https://e621.net/upload_karma_events?search[reason]=replacement_penalty
The replacement penalty is surfaced in a couple of places. It's not really privileged information.

DonovanDMC:
> > > Since the replacement penalty is part of the karma math, I think it makes sense to keep it public.
> > 
> > 
> > It isn't shown on the actual replacement, so if this is kept it's just incredibly obscure, nobody but power users even knows post events exist
> 
> I'd argue the replacement penalty state should instead be made visible to users since it affects the upload limit and karma stats. I don't really see how it could be harmful, except maybe for annoying users begging for it to be removed. But anyway that's out of scope for this PR.
> 
> EDIT: You can just search replacements by penalty... https://e621.net/post_replacements?commit=Search&search%5Bpenalized%5D=true



> I haven't reviewed the PR yet, but I'm generally in agreement with DRLa. [https://e621.net/upload_karma_events?search[reason]=replacement_penalty](https://e621.net/upload_karma_events?search%5Breason%5D=replacement_penalty) The replacement penalty is surfaced in a couple of places. It's not really privileged information.

I'll point both of you to the commit and comment right before, I already removed it for the same reasoning both of you mentioned

app/models/post_event.rb
```diff
@@ -51,33 +78,103 @@ def is_creator_visible?(user)
...
+      else
+        raise ArgumentError, "Value must be truthy or falsy"
+      end
     end
 
-    if params[:action].present?
-      if !CurrentUser.user.is_moderator? && MOD_ONLY_SEARCH_ACTIONS.include?(actions[params[:action]])
-        raise(User::PrivilegeError)
+    def jsonb_numeric_attribute_matches(attribute, range)
+      qualified_column = Arel.sql("extra_data ->> '#{attribute}'")
```
Sindrake, line 111: I am pretty sure that the range queries don't work.
Does not seem like there are tests that cover those.

The `mod_actions.rb` version of this function casts to integer.

```suggestion
      qualified_column = Arel.sql(
        "CASE WHEN (extra_data ->> '#{attribute}') ~ '^-{0,1}[0-9]+$' " \
        "THEN (values ->> '#{attribute}')::INTEGER END",
      )
```

app/models/post_event.rb
```diff
@@ -51,33 +78,103 @@ def is_creator_visible?(user)
...
+    return original_data if CurrentUser.is_admin?
 
-    if params[:post_id].present?
-      q = q.where(post_id: params[:post_id])
+    valid_keys = KNOWN_ACTIONS[action.to_sym]&.keys&.map(&:to_s) || []
+    sanitized_values = original_data.slice(*valid_keys)
+
+    if %w[replacement_deleted].include?(action)
+      sanitized_values = sanitized_values.except("storage_id")
     end
```
Sindrake, line 91: `storage_id` gets stripped - correct.

However, `storage_id: :string` stays in `KNOWN_ACTIONS[:replacement_deleted]`, and `replacement_deleted` is not in `MOD_ONLY_SEARCH_ACTIONS`.
A regular member can issue `?search[action]=replacement_deleted&search[storage_id]=...` and the `to_tsvector` filter runs against the real column.

That's a confirm/deny oracle on a value you deliberately hid from output.
Consider excluding mod/admin-only fields (like storage_id) from the searchable field set for non-privileged users, rather than only from the rendered output.

spec/models/post_event/search_spec.rb
```diff
@@ -139,6 +135,84 @@ def make_event(overrides = {})
...
+    end
+  end
+
+  # -------------------------------------------------------------------------
+  # Unknown action — no JSONB sub-filters applied
+  # -------------------------------------------------------------------------
+  describe "unknown action param" do
+    it "does not raise an error for an unknown action" do
+      make_event(action: :deleted)
+      expect { PostEvent.search(action: "nonexistent_action").to_a }.not_to raise_error
```
Sindrake, line 212: This test is correct.
However, it hides a behavior change: an unknown action now returns **everything**, rather than nothing.

Please, confirm that this is an intentional change.

Sindrake (changes requested):

PR #2461, opened 2026-08-29
[DbExports] Add eris export

DonovanDMC:
As per the request of https://e621.net/forum_topics/61561, adds an export for the eris service. I Chose to name it eris since that's the name of the new service, but can definitely name it iqdb if we're sticking with that name when public facing.

Added the eris database to database.yml, with `ERIS_DB_HOST`, `ERIS_DB_PORT`, and `ERIS_DB_PASSWORD` env variables
Only includes the images table since that seems the most relevant.

Not sure what to really do about testing since the database is managed externally and we have no way to make a copy or really even insert rows beyond manual queries.

### DIFF
--- a/app/jobs/db_export_job.rb
+++ b/app/jobs/db_export_job.rb
@@ -13,33 +13,47 @@ class DbExportJob < ApplicationJob
   EXPORTS = {
     "artists" => {
       query: read_export_sql("artists"),
+      connection: nil,
     },
     "bulk_update_requests" => {
       query: read_export_sql("bulk_update_requests"),
+      connection: nil,
+    },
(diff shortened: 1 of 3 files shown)


DonovanDMC:
Those tests pass locally, I'm not sure why they're failing here

DonovanDMC:
It's failing due to the eris tables not existing, we can either try to setup eris for tests or just make its db export tolerate failure more safely

Sindrake:
Let's come back to this at a later point.

The testing issue could be overcome, but it's kinda a hassle, and I don't want to do it.
But this also has other issues - for example, no username/database overrides for ERIS.

PR #2513, opened 2026-09-07
[DB] Update change_seq tracking & add tests

DonovanDMC:
This pr updates the `posts_trigger_change_seq` database trigger which updates the `posts.change_seq` column in line with the newly added columns, and adds tests to ensure it does not get out of sync again. This also formats the function for better reading, and switches from `!=` to `IS DISTINCT FROM` for null safety.
Added:
- uploader_id
- locked_tags
- bg_color
- video_samples
- pool_ids

Ignored:
- id, created_at, updated_at, change_seq, uploader_ip_addr, has_children (not relevant)
- up_score, down_score, score, hotness, fav_count, comment_count, tag_count, tag_count_* (transient counters)
- file_ext, file_size, image_width, image_height, duration (handled by md5)
- last_comment_bumped_at, last_commented_at, is_comment_disabled, is_comment_locked (comments)

Documentation was added with claude, but the actual code and concept are from GayFurCity/GayFurCity@3e6a8abd5c867b111f34ff7467b2576715f7e64b

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -14,6 +14,16 @@ class Post < ApplicationRecord
   # Smaller favours fresher posts; larger lets posts stay hot longer.
   HOTNESS_TIME_DIVISOR = 432_000.0
 
+  # Columns the posts_trigger_change_seq() DB trigger does not bump change_seq for.
+  # Every posts column must appear either in the trigger (db/migrate/*_update_posts_trigger_change_seq.rb) or here
+  # ChangeSeqSpec fails otherwise, so a newly added column can't silently go uncategorized.
+  CHANGE_SEQ_IGNORED = [
+    :id, :created_at, :updated_at, :change_seq, :uploader_ip_addr, :has_children, # not relevant
+    :up_score, :down_score, :score, :hotness, :fav_count, :comment_count, :tag_count, *TagCategory::CATEGORIES.map { |n| :"tag_count_#{n}" }, # transient counters
+    :file_ext, :file_size, :image_width, :image_height, :duration, # handled by md5
+    :last_comment_bumped_at, :last_commented_at, :is_comment_disabled, :is_comment_locked, # comments
+  ].freeze
+
   # Replaced by pool_ids for pools; sets keep no post-side state (queried via
   # post_sets.post_ids). Dropped in a follow-up migration after the rollback window.
   self.ignored_columns += %w[pool_string]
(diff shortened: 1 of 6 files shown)


app/models/post.rb
```diff
@@ -2091,6 +2101,24 @@ def create_post_events
...
+  module ChangeSeqMethods
+    # Reads the columns the live posts_trigger_change_seq() trigger checks, straight from
+    # Postgres, so this can't drift out of sync with the migration that defines it.
+    # @return [Array<Symbol>] the posts columns the trigger currently compares
+    def change_seq_tracked_columns
+      source = connection.select_one("SELECT prosrc FROM pg_proc WHERE proname = $1", nil, ["posts_trigger_change_seq"])&.[]("prosrc")
+
+      return [] if source.nil?
+      source.scan(/NEW\.(\w+) IS DISTINCT FROM OLD\.\1\b/).flatten.map(&:to_sym)
+    end
```
Sindrake, line 2110: This is an almost verbatim copy of the `MigrationHelpers#existing_change_seq_columns`.
Since `migration_helpers.rb` is already required on boot, I would just delegate to that, and then map to symbols.

spec/models/post/change_seq_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        case col.type
+        when :boolean  then !current
+        when :integer  then current.to_i + 1
+        when :datetime then 1.minute.from_now.change(usec: 0)
+        when :jsonb    then { "changed" => SecureRandom.hex(4) }
+        else "#{current}_#{SecureRandom.hex(4)}"
+        end
+      end
+
+      Post.change_seq_tracked_columns.sort.each do |column|
```
Sindrake, line 51: I would also recommend testing whether `change_seq_tracked_columns` actually returns anything.

```suggestion
      it "found the trigger function" do
        expect(Post.change_seq_tracked_columns).not_to be_empty
      end

      Post.change_seq_tracked_columns.sort.each do |column|
```

lib/migration_helpers.rb
```diff
@@ -0,0 +1,64 @@
...
+  # live function already tracks) plus `add` and minus `remove`.
+  #
+  # With `add`/`remove`, this is reversible on its own - the down direction just swaps them.
+  # Passing an explicit `columns` list instead is a one-off rewrite, and not reversible.
+  # @param columns [Array<String, Symbol>, nil] the exact set of columns to track; defaults
+  #   to whatever posts_trigger_change_seq() currently tracks (see #existing_change_seq_columns)
+  # @param add [Array<String, Symbol>] columns to add to `columns`
+  # @param remove [Array<String, Symbol>] columns to drop from `columns`
+  # @return [void]
+  def update_change_seq(columns = nil, add: [], remove: [])
```
Sindrake, line 52: I would consider raising if both `columns` and `add`/`remove` are provided.

Passing an explicit `columns` list together with `add:`/`remove:` makes the `dir.down` branch produce `columns | remove - add`, which isn't the pre-migration state.

It's an edge case, take it or leave it.

Sindrake (changes requested):
Looks good to me, for the most part.
Could use a tiny bit of cleanup.

Sindrake (approved):
lgtm

PR #2515, opened 2026-09-13
[misc] Added .env-backed ability for custom hosts

DoubleThePsycho:
Comment in `development.rb` says it all, allows contributors to add their custom domain to `.env` without the risk of dirtying a commit.

Personally a huge life-saver, I don't add the relevant files, I just commit all affected files and hope for the best.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -58,6 +58,9 @@
 # When running behind an SSL-terminating reverse proxy, you may want to set this to true
 # RAILS_ASSUME_SSL=true
 
+# Optional hostname to allow in development when accessing through a reverse proxy.
+# RAILS_HOST=e621.example.com
+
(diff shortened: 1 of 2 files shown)


Sindrake (changes requested):
Rails already does this natively.
https://github.com/rails/rails/blob/main/railties/lib/rails/application/configuration.rb#L41-L42

Just setting `RAILS_DEVELOPMENT_HOSTS` should be sufficient.

DoubleThePsycho:
> Rails already does this natively. https://github.com/rails/rails/blob/main/railties/lib/rails/application/configuration.rb#L41-L42
> 
> Just setting `RAILS_DEVELOPMENT_HOSTS` should be sufficient.

Somehow I had overlooked this, and now I feel stupid lol.

PR #2538, opened 2026-09-14
[README] Add link for the populate command file in README

yivarld:
Simply adds a tiny link to point the user to the available environment variables to be used for the populate command.


Sindrake:
I guess?

PR #2555, opened 2026-09-20
[Posts] Save original tag string

DonovanDMC:
Saves the original tag string used when uploading/editing, useful for knowing what invalid tag was added as well as how many general tags were used when uploading.

image


db/exports/post_versions.sql
```diff
@@ -19,7 +19,8 @@ SELECT
   post_versions.description_changed,
   post_versions.updater_id,
   post_versions.updated_at,
-  post_versions.reason
+  post_versions.reason,
+  post_versions.original_tags
```
Sindrake, line 23: The exports now include the `original_tags`, but the API output does not – that has an explicit whitelist.

app/models/post.rb
```diff
@@ -679,6 +679,7 @@ def merge_old_changes
 
     def apply_tag_diff
       return unless tag_string_diff.present?
+      @tags_before_parse = remove_metatags(tag_string_diff.split)
```
Sindrake, line 682: There is an edge case here with quoted metatags.
For example, `source:"a b"` gets split into `source:"a` and `b"`.
I would just save the output of `TagQuery.scan(tag_string_diff)` from a few lines down.

Relatively minor. Take it or leave it.

app/models/post.rb
```diff
@@ -887,6 +889,13 @@ def apply_casesensitive_metatags(tags)
       return tags
     end
 
+    def remove_metatags(tags)
+      tags = tags.grep_v(/\A(?:-pool|pool|newpool|-set|set|fav|-fav|child|-child|upvote|downvote):/i)
```
Sindrake, line 893: This does not strip `@pre_metatags` (`rating:`, `parent:`, `-parent:`, `-?locked:`) or the case-sensitive `source:`/`+source:`.
Intentional?

app/models/post.rb
```diff
@@ -887,6 +889,13 @@ def apply_casesensitive_metatags(tags)
...
+      prefixed, unprefixed = tags.partition { |x| x =~ Tag.categories.regexp }
```
Sindrake, line 894: This feels a little sloppy. `general_store` would match against `general`, for example.
The subsequent `tag.sub` anchors it properly, so it's relatively harmless.
Still, I would use a `/\A(#{categories.regexp}):/` here too.

Sindrake (changes requested):
Pretty solid.
Found some minor issues.

app/models/post.rb
DonovanDMC: Theoretically Intentional, but probably not even thought about due to not being sensitive

The implementation is from GayFurCity/GayFurCity@7de76ec4b75b869f770df582d5c2faca1b23f92f which was over 3 years ago, so any actual logic has been lost to time

